PR #2, opened 2016-05-24
Moar Plugins

FenrirUnbound:
This PR allows the API to load an extra set of plugins in the HapiJS server, as well as any associated `options`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,2 +1,32 @@
 'use strict';
-module.exports = require('./lib/server')();
+const server = require('./lib/server');
+
+/**
+ * User Plugin. A plugin that the user wishes for the Screwdriver API to register and use
+ * @typedef {Object} UserPlugin
+ * @property {Object} register  A HapiJS plugin `register` function. It honors the interface
+ *                              that all HapiJS plugins implement.
+ * @property {Object} options  Additional options to load with the plugin.
+ */
+
+/**
+ * Start the Screwdriver API.
+ * Optionally, it can be started up with an additional set of plugins
+ * @method
+ * @param  {Array.<UserPlugin>}   [userPlugins]  Optional. An array of additional plugins to start
+ *                                               up the server with.
+ * @param  {Function}             callback       Function to invoke when server is started
+ * @return {http.Server}                         A listener: NodeJS http.Server object
+ */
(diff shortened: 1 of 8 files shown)


index.js
```diff
@@ -1,2 +1,32 @@
...
+/**
+ * Start the Screwdriver API.
+ * Optionally, it can be started up with an additional set of plugins
+ * @method
+ * @param  {Array.<UserPlugin>}   [userPlugins]  Optional. An array of additional plugins to start
+ *                                               up the server with.
+ * @param  {Function}             callback       Function to invoke when server is started
+ * @return {http.Server}                         A listener: NodeJS http.Server object
+ */
+module.exports = (userPlugins, callback) => {
```
stjohnjohnson: Why not make plugins required?  Can this be part of a `config` object?

index.js
```diff
@@ -1,2 +1,32 @@
 'use strict';
-module.exports = require('./lib/server')();
+const server = require('./lib/server');
+
+/**
+ * User Plugin. A plugin that the user wishes for the Screwdriver API to register and use
+ * @typedef {Object} UserPlugin
```
stjohnjohnson: How do you do this once and use it everywhere?

lib/registerPlugins.js
```diff
@@ -18,4 +24,45 @@ module.exports = (server, callback) => {
...
+ */
+
+/**
+ * Register any user-requested plugins
+ * @method registerAdditionalPlugins
+ * @param  {Object}                 server    HapiJS server object
+ * @param  {Array.<UserPlugin>}     plugins   An array of plugins to register
+ * @param  {Function}               callback  Function to invoke when completed
+ */
+function registerAdditionalPlugins(server, plugins, callback) {
```
stjohnjohnson: Why are we adding plugins?  Isn't this supposed to just be adding config like what datastore system to use?

stjohnjohnson:
Can you also document in the README how to use this server?

d2lam:
@stjohnjohnson  yes there is a separate branch for README

FenrirUnbound:
Closing this PR. The intent is to allow users to configure the individual plugins, but not configure which plugins get loaded.

We'll open up a follow-up PR with that workflow change

PR #10, opened 2016-06-21
Register jobs plugin

d2lam:
Register the jobs plugin. 
Bump version so that we don't need to pin it in api-aws

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "screwdriver-api",
-  "version": "0.1.0",
+  "version": "1.0.0",
   "description": "API module for the Screwdriver CD service",
   "main": "index.js",
   "scripts": {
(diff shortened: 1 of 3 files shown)


FenrirUnbound:
👍 

Looks good to me

package.json
```diff
@@ -42,6 +42,7 @@
     "hapi-swagger": "^5.1.0",
     "inert": "^3.2.1",
     "screwdriver-plugin-builds": "^1.0.1",
+    "screwdriver-plugin-jobs": "^1.0.0",
     "screwdriver-plugin-pipelines": "0.0.2",
```
tkyi: Can you update `screwdriver-plugin-pipelines` to `0.0.4` while you're here

d2lam: Let's bump plugin-pipelines version so that we don't need to keep pinning it:
https://github.com/screwdriver-cd/plugin-pipelines/pull/4

nkatzman:
👍

PR #11, opened 2016-06-21
Adding Dockerfile and using in-memory-datastore

stjohnjohnson:
This allows us to run this in Docker as well as auto-use the in-memory datastore


jer:
👍

PR #13, opened 2016-06-22
Register platforms plugin

tkyi:
Register the platforms plugin so we can use it in aws-api


d2lam:
👍

FenrirUnbound:
👍 

Looks good to me

giphy 35

PR #15, opened 2016-06-24
Mounting the IMDB EBS volume into the container for persistence

jer:
- Also switched to :LATEST tag for the container

### DIFF
--- a/kubernetes/api_deployment.yaml
+++ b/kubernetes/api_deployment.yaml
@@ -12,14 +12,23 @@ spec:
     spec:
       containers:
       - name: screwdriver-api
-        image: screwdrivercd/api:mockdata
+        image: screwdrivercd/api
         ports:
         - containerPort: 8080
         volumeMounts:
         - name: apikey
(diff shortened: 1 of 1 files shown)


kubernetes/api_deployment.yaml
```diff
@@ -23,3 +23,7 @@ spec:
       - name: apikey
         secret:
           secretName: default-token-sw0ar
+      - name: screwdriver-imdb
+        awsElasticBlockStore:
+          volumeID: vol-3ea656b7
```
nkatzman: Where does the volumeID come from?

jer: I just manually created it in the AWS console. It's the EBS volume id.

nkatzman: Okay, it'll just require a code change if we ever delete it/recreate it

FenrirUnbound:
👍 

dappermenacingbudgie-size_restricted

nkatzman:
👍

PR #17, opened 2016-06-27
Add screwdriver-plugin-login

nkatzman:
This PR adds in the screwdriver-plugin-login plugin for authorization

The login requires it's own config value so in order to pass that to the plugin this PR also makes it possible to override the config values passed to plugins being registered by specifying it directly in the interface

This PR also:
Changes the AWS volume mounted to a new one created (the old one had invalid data in the imdb)
Mounts secrets in a directory called /etc/secrets for use when starting up the server

### DIFF
--- a/bin/server
+++ b/bin/server
@@ -4,12 +4,26 @@
 const winston = require('winston');
 const path = require('path');
 const inMemoryDatastore = require('screwdriver-datastore-imdb');
+const fs = require('fs');
+
+const jwtPrivateKey = process.env.SECRET_JWT_PRIVATE_KEY;
+const oauthClientId = process.env.SECRET_OAUTH_CLIENT_ID;
+const oauthClientSecret = process.env.SECRET_OAUTH_CLIENT_SECRET;
+const password = process.env.SECRET_PASSWORD;
+const https = process.env.IS_HTTPS || false;
 
 require('../')({
     port: process.env.PORT || 8080,
     datastore: new inMemoryDatastore({
         filename: path.join(process.env.DATA_DIR || process.cwd(), '/database.json')
-    })
+    }),
+    'screwdriver-plugin-login': {
+        https,
+        jwtPrivateKey,
+        oauthClientId,
+        oauthClientSecret,
+        password
+    }
 }, (err, server) => {
     if (err) {
         winston.error(err);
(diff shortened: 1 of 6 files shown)


kubernetes/api_deployment.yaml
```diff
@@ -19,6 +19,9 @@ spec:
         - name: apikey
           mountPath: /etc/kubernetes/apikey
           readOnly: true
+        - name: plugin-login-secrets
+          mountPath: /etc/secrets
```
stjohnjohnson: When we add more secrets, this dir will not be the same.

package.json
```diff
@@ -44,6 +44,7 @@
     "screwdriver-datastore-imdb": "^1.0.0",
     "screwdriver-plugin-builds": "^1.0.1",
     "screwdriver-plugin-jobs": "^1.0.0",
+    "screwdriver-plugin-login": "0.0.2",
```
stjohnjohnson: :( 0.0.2?

kubernetes/api_deployment.yaml
nkatzman: okay. Adding in an environment variable for the 'SECRETS_DIR' that will be able to be change

package.json
nkatzman: https://github.com/screwdriver-cd/plugin-login/pull/2

d2lam:
LGTM 👍

tkyi:
👍

bin/server
```diff
@@ -4,12 +4,27 @@
...
+const password = fs.readFileSync(`${secretsDir}/password`, 'utf8').trim();
+const https = process.env.IS_HTTPS || false;
 
 require('../')({
     port: process.env.PORT || 8080,
     datastore: new inMemoryDatastore({
         filename: path.join(process.env.DATA_DIR || process.cwd(), '/database.json')
-    })
+    }),
+    'screwdriver-plugin-login': {
```
d2lam: update README too? :)

nkatzman: updated

d2lam: 👍  
Will let stjohn do the final review :)

FenrirUnbound:
Do we want `bin/server` to bail if the secrets aren't available in the environment? Or is that the responsibility of k8s and we shouldn't worry about it?

nkatzman:
the plugin-login will scream if the secrets aren't there

FenrirUnbound:
👍 

vjt73iu - imgur

Issue #25, opened 2016-07-18
Upgrade to Node 6


stjohnjohnson:
We're currently tied to Node 4 and Node 6 is the latest stable release.  Should be a pretty easy upgrade:

 - https://github.com/nodejs/node/wiki/Breaking-changes-between-v4-and-v5
 - https://github.com/nodejs/node/wiki/Breaking-changes-between-v5-and-v6

stjohnjohnson:
Resolved thanks to @catto

PR #26, opened 2016-07-18
Upgrade to Node.js v6

catto:
This PR upgrades the build and the runtime container to Node.js v6. (#25)
It seems that there is no incompatible api use. I confirmed that all the tests were successful with Node.js v6.3.0.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:4
+FROM node:6
 
 # Create our application direcory
 RUN mkdir -p /usr/src/app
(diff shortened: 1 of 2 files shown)


stjohnjohnson:
👍  Woot! 
do_you_even_lift__bro

nkatzman:
The wercker build didn't run. We should validate that the tests work before we merge

catto:
Hmm.. I pushed another commit (just `commit --amend`) but it seems that the wercker build doesn't run.

stjohnjohnson:
Looks like wercker doesn't run for PRs, it runs for branches.  What?!

stjohnjohnson:
@nkatzman I verified that the test ran locally fine in v6 👍 continues

nkatzman:
👍

PR #31, opened 2016-07-21
Use new volume

d2lam:
I think this is all we need?


FenrirUnbound:
👍 

Looks reasonable to me.

olvycih

kubernetes/api_deployment.yaml
```diff
@@ -49,7 +49,7 @@ spec:
       - name: apikey
         secret:
           secretName: default-token-sw0ar
-      - name: screwdriver-imdb
+      - name: SD-imdb
```
nkatzman: shouldn't change the volume name, it's referenced above

nkatzman:
👍

PR #40, opened 2016-07-24
Validation route to parse/translate screwdriver.yamls

stjohnjohnson:
This adds a new route `/v3/validate` that passes a given yaml file through the [config-parser](https://github.com/screwdriver-cd/config-parser).  This way we don't need to install the config-parser locally in the launcher AND we can start a validate page.

### DIFF
--- a/plugins/validator.js
+++ b/plugins/validator.js
@@ -1,0 +1,46 @@
+'use strict';
+
+const parser = require('screwdriver-config-parser');
+const schema = require('screwdriver-data-schema');
+const validatorSchema = schema.api.validator;
+const boom = require('boom');
+
+/**
+ * Hapi Validator Plugin
+ *  - Validates screwdriver.yaml and returns the expected execution steps
+ * @method register
+ * @param  {Hapi.Server}    server
+ * @param  {Object}         options
+ * @param  {Function} next
+ */
+exports.register = (server, options, next) => {
+    server.route({
+        method: 'POST',
+        path: '/validator',
+        config: {
+            description: 'Validate a given screwdriver.yaml',
+            notes: 'Returns the parsed config or validation errors',
+            tags: ['api', 'validation', 'yaml'],
+            handler: (request, reply) => {
+                parser(request.payload.yaml, (err, pipeline) => {
+                    if (err) {
+                        return reply(boom.badRequest(err.message));
+                    }
+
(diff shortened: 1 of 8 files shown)


test/plugins/validator.test.js
```diff
@@ -0,0 +1,81 @@
...
+
+        it('returns 406 for bad yaml', (done) => {
+            server.inject({
+                method: 'PUT',
+                url: '/validator',
+                payload: {
+                    yaml: 'jobs: {}'
+                }
+            }, (reply) => {
+                assert.equal(reply.statusCode, 406);
```
FenrirUnbound: Are the accept header invalid?

stjohnjohnson: What?

FenrirUnbound: > **406 Not Acceptable**
The requested resource is capable of generating only content not acceptable according to the Accept headers sent in the request.

Source: https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_Client_Error

Is a 400 not sufficient enough to tell the user that their input is invalid?

stjohnjohnson: Good catch.  That'll teach me for reading boom error codes and not the wikipedia page.

FenrirUnbound:
:+1: 

Looks good to me

plugins/validator.js
```diff
@@ -0,0 +1,46 @@
...
+ * Hapi Validator Plugin
+ *  - Validates screwdriver.yaml and returns the expected execution steps
+ * @method register
+ * @param  {Hapi.Server}    server
+ * @param  {Object}         options
+ * @param  {Function} next
+ */
+exports.register = (server, options, next) => {
+    server.route({
+        method: 'PUT',
```
d2lam: Interesting. Why would this be a PUT?

stjohnjohnson: I'm torn.  @jer and I went back and forth.  The reason we went with a PUT is that it's cacheable.  Although a POST makes more sense from the "Providing a block of data, such as the result of submitting a form, to a data-handling process;" perspective.

I'm not picky on this.

Filbird:
*thumbs up* 👍

jer:
👍

d2lam:
👍

PR #52, opened 2016-07-27
Removing Platforms until we come up with a Template design

stjohnjohnson:
Blocked on https://github.com/screwdriver-cd/data-schema/pull/14 and https://github.com/screwdriver-cd/models/pull/22

Right now we have no plans to add Platforms (that affect the entire pipeline), but instead design a concept of Templates (that affect the jobs), let's remove all the Platform code as it's kinda misleading and wasteful.


d2lam:
👍

nkatzman:
👍

PR #56, opened 2016-07-28
Calling list with correct parameters

nkatzman:
Because i broke stuff, here's a fix that calls the model.list functions with the correct values


tkyi:
👍

d2lam:
Makes sense 👍

PR #60, opened 2016-07-29
use sha to start a build

FenrirUnbound:
When Github sends a webhook, it includes the SHA1 in the payload. This change is to include the specific SHA1 that triggered the hook, so that we can build on that particular commit.


jer:
👍

stjohnjohnson:
OOohh, very nice. 👍 
mrw_i_fight_off_a_zerg_rush_and_counter_to_clear_the_base

PR #63, opened 2016-08-01
User's username to start the build

FenrirUnbound:
There is a fundamental gap where we do not have a headless user that acts on behalf of the system. In lieu of that, we need the a user to update the PR status.

This PR is to utilize the user who opened the PR. This will allow us to make status changes to the given PR.

### DIFF
--- a/plugins/github.js
+++ b/plugins/github.js
@@ -17,6 +17,7 @@ let Build;
  * @param  {String}       options.jobId      Identifier for the Job
  * @param  {String}       options.name       Name of the new job (PR-1)
  * @param  {String}       [options.sha]      specific SHA1 commit to start the build with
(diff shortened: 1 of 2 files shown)


plugins/github.js
```diff
@@ -17,6 +17,7 @@ let Build;
  * @param  {String}       options.jobId      Identifier for the Job
  * @param  {String}       options.name       Name of the new job (PR-1)
  * @param  {String}       [options.sha]      specific SHA1 commit to start the build with
+ * @param  {String}       [options.username] Username that triggered the event
```
d2lam: why optional?

stjohnjohnson:
We don't want to use the user that started the build. 👎

FenrirUnbound:
In light of new information, closing this for something better

PR #67, opened 2016-08-02
Refactoring Credentials and adding Scope

stjohnjohnson:
Also added pretty-printing of errors (no more 500 with no context).

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -4,6 +4,29 @@ const Hapi = require('hapi');
 const registrationMan = require('./registerPlugins');
 
 /**
+ * If we're throwing errors, let's have them say a little more than just 500
+ * @method prettyPrintErrors
+ * @param  {Hapi.Request}    request Hapi Request object
+ * @param  {Hapi.Reply}      reply   Hapi Reply object
+ */
+function prettyPrintErrors(request, reply) {
+    if (request.response.isBoom) {
+        const err = request.response;
+        const errName = err.output.payload.error;
+        const errMessage = err.message;
+        const statusCode = err.output.payload.statusCode;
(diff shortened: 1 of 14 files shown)


lib/server.js
```diff
@@ -4,6 +4,29 @@ const Hapi = require('hapi');
...
+ * If we're throwing errors, let's have them say a little more than just 500
+ * @method prettyPrintErrors
+ * @param  {Hapi.Request}    request Hapi Request object
+ * @param  {Hapi.Reply}      reply   Hapi Reply object
+ */
+function prettyPrintErrors(request, reply) {
+    if (request.response.isBoom) {
+        const err = request.response;
+        const errName = err.output.payload.error;
+        const errMessage = err.message;
```
petey: What is the difference between `err.message` and `err.output.payload.message`? Can this method just `return reply(err.output.payload).code(statusCode)`?

stjohnjohnson: Boom will automatically hide all 500 errors from the end user.  https://github.com/hapijs/boom/blob/master/lib/index.js#L136-L138

This reverts that feature.

petey: Got it. 😦  Lame.

petey:
👍 

cat_gif

jer:
👍

lib/server.js
FenrirUnbound: It's also [in their docs](https://github.com/hapijs/boom#http-5xx-errors)

> All 500 errors hide your message from the end user. Your message is recorded in the server log.

FenrirUnbound:
👍 


ckhe4w1 - imgur

PR #69, opened 2016-08-02
Datastore/Executor refactor, HTTPs, and TokenGenerator

stjohnjohnson:
- Refactoring Datastore and Executor into global server.app settings
- Support HTTP or HTTPs (and alternative cnames)
- Pass apiUri and token generator to build.create


petey:
👍  I really like how much simpler the method signatures are.

strike_

bin/server
```diff
@@ -17,30 +17,21 @@ const executor = new ExecutorPlugin(executorConfig[executorConfig.plugin]);
 // Setup Login
 const loginConfig = config.get('login');
 
-loginConfig.datastore = datastore;
-
 // Setup httpd
 const httpdConfig = config.get('httpd');
```
tkyi: You can just set this to `const httpd = config.get('httpd');` and replace line 27 with `httpd,`

tkyi:
👍 
giphy-2

PR #72, opened 2016-08-03
Remove overuse of bind/apply in waterfalls.

petey:
Both bind and async.apply produce anonymous functions. Function.prototype.apply normally takes a context, but async.apply uses `null` for this context, thus requiring a class method to be bound. We don't need to do any of this since we can create a simple ES6 style anonymous function that executes the original function properly, without needing to insert a context at all, while being just as readable.

I've also fixed the caps on the "instanceOf" variables so they don't look like their respective classes.

### DIFF
--- a/plugins/github.js
+++ b/plugins/github.js
@@ -4,9 +4,9 @@ const hoek = require('hoek');
 const async = require('async');
 const boom = require('boom');
 const Models = require('screwdriver-models');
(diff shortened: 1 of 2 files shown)


FenrirUnbound:
👍 

Looks good to me

c0bu5fi7d8dx

d2lam:
👍

PR #75, opened 2016-08-04
Build create passes username

shruthi-venkateswaran:
Using the new models, Build create now takes in another param called config.username: https://github.com/screwdriver-cd/models/blob/master/lib/build.js#L95

Build model constructor now takes in password.

### DIFF
--- a/plugins/builds/create.js
+++ b/plugins/builds/create.js
@@ -1,5 +1,6 @@
 'use strict';
 const boom = require('boom');
+const hoek = require('hoek');
 const urlLib = require('url');
 const validationSchema = require('screwdriver-data-schema');
 const Model = require('screwdriver-models');
@@ -26,8 +27,11 @@ module.exports = (server) => ({
                 tokenGen: (buildId) =>
                     request.server.plugins.login.generateToken(buildId, ['build'])
             };
+            const config = hoek.applyToDefaults(payload, {
+                username: request.auth.credentials.username
+            });
 
-            Build.create(payload, (err, data) => {
+            Build.create(config, (err, data) => {
                 if (err) {
                     return reply(boom.wrap(err));
                 }
--- a/plugins/builds/create.js
+++ b/plugins/builds/create.js
@@ -1,11 +1,12 @@
 'use strict';
+const async = require('async');
 const boom = require('boom');
 const hoek = require('hoek');
 const urlLib = require('url');
 const validationSchema = require('screwdriver-data-schema');
 const Model = require('screwdriver-models');
 
-module.exports = (server) => ({
+module.exports = (server, options) => ({
     method: 'POST',
     path: '/builds',
     config: {
@@ -17,9 +18,20 @@ module.exports = (server) => ({
             scope: ['user']
         },
         handler: (request, reply) => {
-            const Build = new Model.Build(
+            const build = new Model.Build(
                 server.settings.app.datastore,
-                server.settings.app.executor
+                server.settings.app.executor,
+                options.password
+            );
+            const job = new Model.Job(
+                server.settings.app.datastore
+            );
+            const pipeline = new Model.Pipeline(
+                server.settings.app.datastore
+            );
+            const user = new Model.User(
+                server.settings.app.datastore,
+                options.password
             );
             const payload = {
                 jobId: request.payload.jobId,
@@ -27,11 +39,38 @@ module.exports = (server) => ({
                 tokenGen: (buildId) =>
                     request.server.plugins.login.generateToken(buildId, ['build'])
             };
-            const config = hoek.applyToDefaults(payload, {
-                username: request.auth.credentials.username
-            });
+            const username = request.auth.credentials.username;
 
-            Build.create(config, (err, data) => {
+            // Check if user can create a build
+            async.waterfall([
+                // Get job
+                (next) => job.get(request.payload.jobId, next),
+
+                // Get pipeline
+                (jobObj, next) => pipeline.get(jobObj.pipelineId, next),
+
+                // Get permissions
+                (pipelineObj, next) => user.getPermissions({
+                    username,
+                    scmUrl: pipelineObj.scmUrl
+                }, next),
+
+                (permissions, next) => {
+                    if (!permissions.push) {
+                        return next(boom.unauthorized(`User ${username} `
+                          + 'does not have push permission for this repo'));
+                    }
+
+                    return next();
+                },
+                (next) => {
+                    const config = hoek.applyToDefaults(payload, {
+                        username
+                    });
+
+                    build.create(config, next);
+                }
+            ], (err, result) => {
                 if (err) {
                     return reply(boom.wrap(err));
                 }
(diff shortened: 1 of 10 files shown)


plugins/builds/create.js
```diff
@@ -26,8 +27,11 @@ module.exports = (server) => ({
                 tokenGen: (buildId) =>
                     request.server.plugins.login.generateToken(buildId, ['build'])
             };
+            const config = hoek.applyToDefaults(payload, {
+                username: request.auth.credentials.username
```
stjohnjohnson: This is great an all, but shouldn't we be checking if they can create the build first?  And if so, shouldn't we add them to the list?

d2lam: Hmmm why do we want to add them to the list? If they don't have permission then they shouldn't be creating the build... Right?

plugins/builds/create.js
```diff
@@ -16,18 +18,49 @@ module.exports = (server) => ({
...
                 tokenGen: (buildId) =>
                     request.server.plugins.login.generateToken(buildId, ['build'])
             };
+            const username = request.auth.credentials.username;
 
-            Build.create(payload, (err, data) => {
+            // Check if user can create a build
+            async.waterfall([
+                // Get job
+                (next) => build.job.get(request.payload.jobId, next),
```
stjohnjohnson: What is `build.job`?

plugins/builds/create.js
```diff
@@ -16,18 +18,49 @@ module.exports = (server) => ({
...
+
+                // Get pipeline
+                (jobObj, next) => build.pipeline.get(jobObj.pipelineId, next),
```
stjohnjohnson: What is `build.pipeline`?

plugins/builds/create.js
```diff
@@ -16,18 +18,49 @@ module.exports = (server) => ({
...
+
+                // Get permissions
+                (pipelineObj, next) => build.user.getPermissions({
+                    username,
+                    scmUrl: pipelineObj.scmUrl
+                }, next),
+
+                // eslint-disable-next-line consistent-return
```
stjohnjohnson: Anytime you have to disable an eslint warning, it's a code-smell

plugins/builds/create.js
```diff
@@ -16,18 +18,49 @@ module.exports = (server) => ({
...
+                (permissions, next) => {
+                    if (!permissions.push) {
+                        return reply(boom.unauthorized(`User ${username} `
```
stjohnjohnson: If you do this, then the waterfall will wait forever.

plugins/builds/create.js
```diff
@@ -16,18 +18,49 @@ module.exports = (server) => ({
...
+                          + 'does not have push permission for this repo'));
+                    }
+                    next();
```
stjohnjohnson: Remove the lint error by adding `return` in front of this next (or wrap in else)

plugins/webhooks/github.js
```diff
@@ -35,26 +37,24 @@ function pullRequestOpened(options, request, reply) {
...
-        (pipelineData, next) => {
-            const username = Object.keys(pipelineData.admins)[0];
-
+        (next) => {
             request.log([
                 'webhook-github',
                 eventId,
                 jobId,
                 pipelineId
             ], `${username} admin selected`);
```
stjohnjohnson: Update this text

plugins/builds/create.js
d2lam: It's the job model. Build has these models tied to it:
https://github.com/screwdriver-cd/models/blob/master/lib/build.js#L73-L75

plugins/builds/create.js
d2lam: Same as above:
https://github.com/screwdriver-cd/models/blob/master/lib/build.js#L73-L75

plugins/pipelines/create.js
```diff
@@ -31,15 +31,15 @@ module.exports = (server, options) => ({
...
                 (permissions, next) => {
                     if (!permissions.admin) {
-                        return reply(boom.unauthorized(`User ${username} `
+                        return next(boom.unauthorized(`User ${username} `
                             + 'is not an admin of this repo'));
                     }
 
                     return Pipeline.get(pipelineId, next);
                 },
                 (data, next) => {
```
FenrirUnbound: We're not leveraging the data for anything other than whether it's truthy or falsy. Could you update the variable to better describe what you'll use it for? 

In this specific case, you're using it to check whether a pipeline exists for a given SCM url. Renaming it to something along the lines of `doesPipelineExist` would make it much more clearer as to what significance it has for the function.

plugins/builds/create.js
```diff
@@ -1,10 +1,12 @@
 'use strict';
+const async = require('async');
 const boom = require('boom');
+const hoek = require('hoek');
 const urlLib = require('url');
 const validationSchema = require('screwdriver-data-schema');
 const Model = require('screwdriver-models');
 
-module.exports = (server) => ({
+module.exports = (server, options) => ({
```
FenrirUnbound: Question for anyone (not just the submitters): do we want to bother with documenting the method signature here?

petey:
👍

FenrirUnbound:
I'm confused. Are we saying that the user who created the PR must also be in the admins list?

plugins/builds/create.js
```diff
@@ -16,18 +18,59 @@ module.exports = (server) => ({
             scope: ['user']
         },
         handler: (request, reply) => {
-            const Build = new Model.Build(
+            const build = new Model.Build(
```
stjohnjohnson: These should be defined on line 10

plugins/pipelines/create.js
d2lam: Hmm... this is not the scope of this PR. I'm just changing the `reply` to `next`. But since this is a simple change, I can do that.

d2lam:
@FenrirUnbound  No. The person who created the PR doesn't need to be an admin. We are just passing in whoever creates the PR so we can put this in the `cause` field: `Started by user Darren`. 

Build create takes care of getting the admin to create the github status.

plugins/builds/create.js
d2lam: huh? How do you do that? This whole thing is an object
```
{
method: 'POST,
path: '/builds', 
....
}
```

FenrirUnbound:
> Build create takes care of getting the admin to create the github status. 

👍 Makes sense

plugins/builds/create.js
stjohnjohnson: We should, so we at least know the options required.

plugins/builds/create.js
stjohnjohnson: Yeah, that's because it returns an object.  You can do this, but since we're refactoring anyway, it's okay to skip.

stjohnjohnson:
👍  LGTM

tkyi:
👍 LGTM
giphy-3

plugins/builds/create.js
FenrirUnbound: I suppose it was decided against that notion

PR #80, opened 2016-08-05
Register webhooks correctly

d2lam:
`npm start` is not working because the plugin is called `webhooks`, not `github`
@stjohnjohnson


stjohnjohnson:
👍  Sorry

PR #86, opened 2016-08-09
Route for returning statistics

nkatzman:
Creating a new plugin that can be used for exposing statistics about the entire system. 

Right now there are only executor statistics but eventually there could be a plethora of statistics we want to expose about the system


nkatzman:
This PR is blocked on pulling in the new executor version from https://github.com/screwdriver-cd/executor-k8s/pull/26

jer:
👍

petey:
👍

PR #88, opened 2016-08-11
Use new models in api

petey:
Refactored api plugins to utilize new models/factories.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -30,20 +30,24 @@ function prettyPrintErrors(request, reply) {
  * @method
  * @param  {Object}      config
  * @param  {Object}      config.httpd
- * @param  {Integer}     config.httpd.port Port number to listen to
- * @param  {String}      config.httpd.host Public hostname
- * @param  {Object}      config.httpd.tls  TLS Configuration
- * @param  {Datastore}   config.datastore  Datastore to use for resources
- * @param  {Executor}    config.executor   Executor to use for resources
- * @param  {Function}    callback          Callback to invoke when server has started.
- * @return {http.Server}                   A listener: NodeJS http.Server object
+ * @param  {Integer}     config.httpd.port          Port number to listen to
+ * @param  {String}      config.httpd.host          Public hostname
+ * @param  {Object}      config.httpd.tls           TLS Configuration
+ * @param  {Factory}     config.pipelineFactory     Pipeline Factory instance
+ * @param  {Factory}     config.jobFactory          Job Factory instance
+ * @param  {Factory}     config.userFactory         User Factory instance
+ * @param  {Factory}     config.buildFactory        Build Factory instance
+ * @param  {Function}    callback                   Callback to invoke when server has started.
+ * @return {http.Server}                            A listener: NodeJS http.Server object
  */
 module.exports = (config, callback) => {
     // Create a server with a host and port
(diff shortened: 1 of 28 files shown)


jer:
Tough to review thoroughly, but I trust it 👍 

ygo9bh2

lib/server.js
```diff
@@ -42,8 +42,10 @@ module.exports = (config, callback) => {
     // Create a server with a host and port
     const server = new Hapi.Server({
         app: {
-            datastore: config.datastore,
-            executor: config.executor
+            pipelineFactory: config.pipelineFactory,
```
nkatzman: will need to update documentation of the function with this config change

Also, I'm wondering if there is a way we can make this generic

As in we pass the values in the config as:
```
config.factories = {
   pipelineFactory: instanceOfPipelineFactory
   ...
}
``` 

And then we can loop each factory and place them in the app.

Idk if that would be a good solution, but right now it seems like this `lib/server` is completely dependent on the `bin/server` to pass in all of the factories

File: plugins/builds/README.md [hunk 2]
```diff
@@ -8,16 +8,21 @@ const server = new Hapi.Server();
...
 server.connection({ port: 3000 });
 
 server.register({
     register: buildsPlugin,
     options: {
-        datastore: new Datastore(),
-        executor: new Executor({
-            executorOption1: 'hostname'
-        })
+        pipelineFactory: Models.PipelineFactory.getInstance({ datastore }),
```
nkatzman: does this actually work?

lib/server.js
petey: I'll update the docs.

`lib/server` is already dependent on `bin/server` to pass datastore and executor. Since datastore and executor are needed only by the factory methods, it makes sense to pass the factories in the same way.

plugins/builds/README.md
petey: Hmm. Maybe? Did `options.datastore` actually work before? Alternatively, we could just remove the options part entirely.

plugins/webhooks/github.js
```diff
@@ -22,52 +18,43 @@ let build;
  * @param  {Hapi.reply}   reply   Reply to user
  */
 function pullRequestOpened(options, request, reply) {
+    const jobFactory = request.server.settings.app.jobFactory;
+    const buildFactory = request.server.settings.app.buildFactory;
     const eventId = options.eventId;
     const pipelineId = options.pipelineId;
-    const jobId = options.jobId;
```
d2lam: We no longer use options.jobId then? Should remove it from comment section.

plugins/pipelines/create.js
```diff
@@ -18,60 +39,60 @@ module.exports = (server, options) => ({
...
-                if (err) {
-                    return reply(boom.wrap(err));
-                }
-                const location = urlLib.format({
-                    host: request.headers.host,
-                    port: request.headers.port,
-                    protocol: request.server.info.protocol,
-                    pathname: `${request.path}/${result.id}`
-                });
+                // hooray, a pipeline is born!
```
tkyi: 🎉 🎈

tkyi:
👍  LGTM
giphy-6

FenrirUnbound:
👍 

Looks good to me

ioxaegu

PR #89, opened 2016-08-11
Add stack trace on 500

petey:
Hard to debug errors when 500 as there is no stack trace in the logs. This PR aims to fix that.


jer:
👍

nkatzman:
👍

PR #105, opened 2016-08-17
Add scmPlugin and fix Endtime

d2lam:
- Pull in new datastore dynamo and models, which uses the new schema to validate. 
- Pull in scmPlugin to pass into model factories


nicolaifsf:
endtime fixes look good to me

nkatzman:
👍

jer:
👍

PR #107, opened 2016-08-18
Pass Factories via server.app

nkatzman:
This PR changes how we pass the factories to the plugins, this is because the old way, via the instantiation of the server, actually sets the values via a shallow copy (as seen https://github.com/hapijs/hapi/blob/master/lib/server.js#L34)

Instead, we are going to pass them via server.app settings, which are available wherever there is a server object. 
https://github.com/hapijs/hapi/blob/master/API.md#serverapp

This fixes a bug where the factories do not contain the instance variables that are set

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -46,14 +46,16 @@ function prettyPrintErrors(request, reply) {
  */
 module.exports = (config, callback) => {
     // Create a server with a host and port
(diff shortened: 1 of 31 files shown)


d2lam:
👍  LGTM

petey:
👍

jer:
👍

PR #112, opened 2016-08-18
Fixing the externally routable URI so it can be used easily

stjohnjohnson:
Now `server.info.uri` will be correct.

### DIFF
--- a/kubernetes/api_deployment.yaml
+++ b/kubernetes/api_deployment.yaml
@@ -19,12 +19,8 @@ spec:
         env:
           - name: DATASTORE_PLUGIN
             value: dynamodb
-          - name: HOST
-            value: 0.0.0.0
-          - name: API_URI
-            value: http://a4677c9873c9611e6aa7102b92f75d5c-1135862614.us-west-2.elb.amazonaws.com
-          - name: DATA_DIR
-            value: /opt/screwdriver/data
+          - name: URI
+            value: http://api.screwdriver.cd
           - name: SECRET_JWT_PRIVATE_KEY
             valueFrom:
               secretKeyRef:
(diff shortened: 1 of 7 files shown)


plugins/builds/list.js
```diff
@@ -12,6 +12,7 @@ module.exports = () => ({
         notes: 'Returns all build records',
         tags: ['api', 'builds'],
         handler: (request, reply) => {
+            console.log(request.server.info.uri);
```
d2lam: remove console.log

kubernetes/api_deployment.yaml
```diff
@@ -19,12 +19,8 @@ spec:
...
           - name: DATASTORE_PLUGIN
             value: dynamodb
-          - name: HOST
-            value: 0.0.0.0
-          - name: API_URI
-            value: http://a4677c9873c9611e6aa7102b92f75d5c-1135862614.us-west-2.elb.amazonaws.com
-          - name: DATA_DIR
-            value: /opt/screwdriver/data
+          - name: URI
+            value: http://api.screwdriver.cd
```
d2lam: Is this set up yet?

petey:
Is there any reason why apiUrl, which is for all purposes a constant for as long as the server is running, is not part of the executor config to begin with? There doesn't seem to be a good reason why models and plugins need to shuttle this value around.

I have similar issues with the tokenGen function.

plugins/builds/list.js
stjohnjohnson: Oh snap.

kubernetes/api_deployment.yaml
stjohnjohnson: Yes, but the LB is gone.

stjohnjohnson:
You can configure anything on start, but it would be best coming from `server.info.uri`

kubernetes/api_deployment.yaml
tkyi: Can you add a comment here (since we are pointing users from the guide to this file) that they should put their `LoadBalancer Ingress` here?

petey:
👍

PR #121, opened 2016-08-20
Setting Start Time on build start

stjohnjohnson:


plugins/webhooks/build.js
```diff
@@ -42,10 +42,13 @@ module.exports = () => ({
                 if (['SUCCESS', 'FAILURE', 'ABORTED'].indexOf(status) > -1) {
                     build.meta = request.payload.meta || {};
                     build.endTime = (new Date()).toISOString();
+                } else {
+                    build.startTime = (new Date()).toISOString();
```
d2lam: Not sure how the launcher interacts with this. Does it only hit this endpoint when the on start and notify? Are we sure it won't do it for status = `RUNNING`?

stjohnjohnson: That's the status it sends on start.

d2lam: oh I mean would it hit this end points multiple times or just once?

Filbird:
👍

plugins/webhooks/build.js
stjohnjohnson: Right now it's only twice, once on start and once on end.

d2lam:
👍  Looks good

PR #123, opened 2016-08-21
Send the new SHA to the build create for a PR sync event

stjohnjohnson:
Right now, everytime we send a sync event (updating a PR), the current API will use the SHA from the master branch instead of the PR.  This corrects that.


d2lam:
Makes sense 👍

Filbird:
👍 LGTM 👍

PR #126, opened 2016-08-22
Case-insensitive SCM URLs from Github Webhooks

### LINKED ISSUE #82: scmUrl case sensitive?
We lower case all the ScmUrl a while ago: https://github.com/screwdriver-cd/screwdriver/pull/61

It's giving us problem. The github API actually makes a distinction between lower and upper case:
https://api.github.com/repos/screwdriver-cd/models/branches/newformat gives nothing
https://api.github.com/repos/screwdriver-cd/models/branches/NewFormat gives the branch

FenrirUnbound:
When creating a pipeline, we [save the repository as lowercase](https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/pipelines/create.js#L25). This causes issues for the webhook notifications, since the repository name is case-sensitive in the payload.

This PR is to take a similar approach to the pipelines endpoint and convert the SCM URL to lowercase before processing the information further. This should be the last bit to resolve #82


catto:
:+1:

plugins/webhooks/github.js
```diff
@@ -171,7 +171,7 @@ function pullRequestEvent(request, reply) {
     const payload = request.payload;
     const action = hoek.reach(payload, 'action');
     const prNumber = hoek.reach(payload, 'pull_request.number');
-    const repository = hoek.reach(payload, 'pull_request.base.repo.ssh_url');
+    const repository = hoek.reach(payload, 'pull_request.base.repo.ssh_url').toLowerCase();
```
nkatzman: Part of me thinks you should be running scmUrl, and prRef through the `scmPlugin.formatScmUrl` https://github.com/screwdriver-cd/scm-github/blob/master/index.js#L71-L79

FenrirUnbound: `prRef` wont' be formatted properly by `scmPlugin.formatScmUrl`

plugins/webhooks/github.js
```diff
@@ -173,8 +175,8 @@ function pullRequestEvent(request, reply) {
     const prNumber = hoek.reach(payload, 'pull_request.number');
     const repository = hoek.reach(payload, 'pull_request.base.repo.ssh_url');
     const branch = hoek.reach(payload, 'pull_request.base.ref');
-    const scmUrl = `${repository}#${branch}`;
-    const prRef = `${repository}#pull/${prNumber}/merge`;
+    const scmUrl = scmGithub.formatScmUrl(`${repository}#${branch}`);
```
nkatzman: should we make this a static method? I don't like that we need to instantiate a new plugin.

The alternative would be to expose the scmPlugin in the server like we do with factories

FenrirUnbound: It's annoying how I had to instantiate an object specifically for this function, which is another reason why I was adverse to changing to this way. Not that it isn't correct, but using it is awkward to implement for a case like this.

nkatzman: I agree, instantiating an object just for this case is awkward, thats why a better solutions is to make a static function or use the scmPlugin that's exposed in the server

I feel like it's definitely the right solution to be using the `formatScmUrl` over just lower-casing the `repository` though

petey: Definitely expose the one passed to the server. We don't have multiple scm plugins now, but we don't want to start hardcoding things here, either.

stjohnjohnson:
Ideally we should stop storing things as the SCM url and instead use some sort of unique identifier like the GitHub repository ID (number)

petey:
For purposes of comparisons and search, using github repo id is a great idea. We still need to be storing things like scmUrl so the UI doesn't need to have to figure out how to generate repo links.

stjohnjohnson:
Closing in favor of https://github.com/screwdriver-cd/screwdriver/issues/160

PR #127, opened 2016-08-22
Fixing Screwdriver.yaml (we don't support Node 5 or 4)

stjohnjohnson:


nkatzman:
👍

tkyi:
do you mean we don't support Node 4 or 5?

PR #134, opened 2016-08-24
Set common data in build factory

petey:
Depends on models-10.0.0

Sets global apiUri and tokenGen functions on server's buildFactory instance, so these values do not need to be retrieved and passed around to all build models later in multiple places.


d2lam:
👍

PR #136, opened 2016-08-24
Adding support for admins to impersonate users or builds

### LINKED ISSUE #98: 500 when refreshing `/v3/login` route after token is returned
When you go to `/v3/login` from the browser and then refresh the URL, a 500 is returned with:

```
{
"statusCode": 500,
"error": "Internal Server Error",
"message": "Missing github request token cookie"
}
```

stjohnjohnson:
This allows users to impersonate other users by passing additional args to login:

- Login as Batman: `GET /login/batman`
- Login as Build: `GET /login/0beec7b5ea3f0fdbc95d0dd47f3c5bc275da8a33/true`

Fixes #98


plugins/login/login.js
```diff
@@ -13,37 +13,64 @@ const creds = require('./credentials');
...
+                if (!scope.includes('admin')) {
+                    return reply(boom.forbidden(
+                        `User ${username} is not an admin and cannot impersonate`
+                    ));
+                }
+                let isBuild = false;
 
-            const githubToken = request.auth.credentials.token;
+                [username, isBuild] = request.params.impersonate.split('/');
+                scope = [
```
d2lam: We reassigned `scope` here, so that means it won't include `admin` in the `scope` anymore. Is that what we want? If that's the case, we can just use a flag to indicate it's an admin to replace line 46.

d2lam: Hmmm... I guess you want line 46 instead of a flag for the case where it's an `admin` and not `impersonated`?

d2lam:
Just some questions, the rest LGTM 👍

plugins/login/login.js
stjohnjohnson: Yup!

config/default.yaml
```diff
@@ -30,6 +30,20 @@ login:
...
+    admins:
+        - FenrirUnbound
+        - Filbird
+        - cynthiax
+        - d2lam
+        - dvdizon
+        - jer
+        - minz1027
+        - nicolaifsf
+        - nkatzman
```
jer: hmmm....

jer:
👍

PR #137, opened 2016-08-25
Add cors support for all routes

petey:
Add a CORS configuration to the server globally that applies to all routes.


stjohnjohnson:
👍  
cors

Issue #135, opened 2016-08-24
Streaming/Grouped Logs


stjohnjohnson:
> As a user, I want to read the logs of my build, both in-progress and after it is done.

For this feature, we need to take the live streaming logs and present them to the user on a web UI grouped by the step.  Steps should contain the exit code as well as the time it took to complete.

stjohnjohnson:
We're pretty done with the first release of this right now.  Logs are "streamed" to the UI.  

There is one major feature I want to add before we consider this done:
 - Local alternative to S3

And then the bugs we found I want to move into our HIP sprint:
 - Build page doesn't update to next step when previous one finishes
 - Build page takes 30+ seconds in browser to load 3k logs
 - Build page doesn't scroll to follow streaming build

stjohnjohnson:
Done done!

PR #138, opened 2016-08-25
Adding steps and a shell of logs

stjohnjohnson:
Depends on https://github.com/screwdriver-cd/data-schema/pull/34

Feature: https://github.com/screwdriver-cd/screwdriver/issues/135

### DIFF
--- a/plugins/builds/steps/get.js
+++ b/plugins/builds/steps/get.js
@@ -1,0 +1,41 @@
+'use strict';
+const boom = require('boom');
+const schema = require('screwdriver-data-schema');
+const getSchema = schema.models.build.getStep;
+
+module.exports = () => ({
+    method: 'GET',
+    path: '/builds/{id}/steps/{name}',
+    config: {
+        description: 'Get a step for a build',
+        notes: 'Returns a step record',
+        tags: ['api', 'builds', 'steps'],
+        handler: (request, reply) => {
+            const factory = request.server.app.buildFactory;
+
+            factory.get(request.params.id)
+                .then(model => {
+                    if (!model) {
+                        throw boom.notFound('Build does not exist');
+                    }
+
+                    const stepModel = model.steps.filter((step) => (
+                        step.name === request.params.name
+                    )).pop();
+
+                    if (!stepModel) {
+                        throw boom.notFound('Step does not exist');
+                    }
+
+                    return reply(stepModel);
+                })
+                .catch(err => reply(boom.wrap(err)));
+        },
+        response: {
+            schema: getSchema
+        },
+        validate: {
+            params: schema.api.loglines.params
+        }
+    }
+});
--- a/plugins/builds/steps/update.js
+++ b/plugins/builds/steps/update.js
@@ -1,0 +1,62 @@
+'use strict';
+const boom = require('boom');
+const schema = require('screwdriver-data-schema');
+
+module.exports = () => ({
+    method: 'PUT',
+    path: '/builds/{id}/steps/{name}',
+    config: {
+        description: 'Update a step for a build',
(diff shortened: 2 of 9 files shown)


plugins/builds/steps/update.js
```diff
@@ -0,0 +1,67 @@
...
+            }
+
+            return factory.get(buildId)
+                .then(build => {
+                    if (!build) {
+                        throw boom.notFound(`Build ${buildId} does not exist`);
+                    }
+                    const steps = build.steps;
+
+                    stepIndex = steps.reduce((prevIndex, step, index) => {
```
petey: Same thing in one line:
`steps.findIndex(s => s.name === request.params.name);`

stjohnjohnson: I couldn't get findIndex to work.  I'll try again.

plugins/builds/steps/update.js
```diff
@@ -0,0 +1,61 @@
...
+                .then(build => {
+                    if (!build) {
+                        throw boom.notFound(`Build ${buildId} does not exist`);
+                    }
+                    const steps = build.steps;
+
+                    stepIndex = steps.findIndex((step) => step.name === request.params.name);
+
+                    if (stepIndex === -1) {
+                        throw boom.notFound('Step does not exist');
```
d2lam: Would be helpful to add the `Step ${name}` here

petey:
👍 

giphy-1

plugins/builds/steps/update.js
stjohnjohnson: Meh.

plugins/builds/steps/get.js
```diff
@@ -0,0 +1,41 @@
...
+
+                    return reply(stepModel);
+                })
+                .catch(err => reply(boom.wrap(err)));
+        },
+        response: {
+            schema: getSchema
+        },
+        validate: {
+            params: schema.api.loglines.params
```
d2lam: This is more related to the schema: we never set `{id}` and `{name}` to be required keys, so this validation is not really strict enough:
https://github.com/screwdriver-cd/data-schema/blob/master/api/loglines.js#L6-L9

d2lam: I guess it's fine since it's going to return 404 anyway

plugins/builds/steps/update.js
d2lam: Lol what happened?

plugins/builds/steps/update.js
```diff
@@ -0,0 +1,61 @@
...
+                    }
+
+                    if (request.payload.code) {
+                        steps[stepIndex].code = request.payload.code;
+                        steps[stepIndex].endTime = request.payload.endTime;
+                    } else {
+                        steps[stepIndex].startTime = request.payload.startTime;
```
minz1027: What if there is no `startTime` in the payload? It is optional in `schema.models.build.updateStep`, the user can send the request without it.

plugins/builds/steps/update.js
```diff
@@ -0,0 +1,62 @@
...
+
+                    if (stepIndex === -1) {
+                        throw boom.notFound('Step does not exist');
+                    }
+
+                    if (request.payload.code) {
+                        steps[stepIndex].code = request.payload.code;
+                        steps[stepIndex].endTime = request.payload.endTime || now;
+                    } else {
+                        steps[stepIndex].startTime = request.payload.startTime || now;
```
d2lam: It still kinda concerns me that our validation is not very strict. Like they can put endTime only in the payload. And then this would update the startTime instead.

jer:
👍

petey:
👍 

giphy-2

PR #139, opened 2016-08-26
Fixes Job creation during PR open

stjohnjohnson:
Apparently we create Jobs in the GitHub PR event as well as in pipeline.sync, we should centralize this sooner rather then later


petey:
👍 

giphy-3

Filbird:
👍

Issue #144, opened 2016-08-27
Vulnerable to CSRF


stjohnjohnson:
To protect against CSRF, we should serve crumbs and require them for writes.

Probably use https://github.com/hapijs/crumb

stjohnjohnson:
Resolved by #147

PR #147, opened 2016-08-27
Add crumb issuer

catto:
I added a crumb issuer for write accesses. #144 
* Implemented as a plugin for v3 route (`/v3/crumb`)
* All accesses to POST, PUT and DELETE endpoints are verified with the crumb.
* Crumb needs to be passed by `X-CSRF-Token` header.

### DIFF
--- a/lib/registerPlugins.js
+++ b/lib/registerPlugins.js
@@ -42,6 +42,7 @@ function registerResourcePlugins(server, config, callback) {
         'jobs',
         'pipelines',
         'webhooks',
+        'crumb',
         'stats'
     ];
 
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "bell": "^8.0.0",
     "boom": "^3.2.2",
     "config": "^1.21.0",
+    "crumb": "^6.0.3",
     "good": "^7.0.0",
     "good-console": "^6.1.1",
     "good-squeeze": "^4.0.0",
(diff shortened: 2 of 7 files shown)


stjohnjohnson:
Awesome job, @catto.

We should turn off crumb protection on the webhooks endpoints as GitHub and the Launcher do not support crumb.

How does this work for multiple replicas?  If we run two of them on kubernetes, what do they use to validate the crumb the other one issued?

catto:
Ah, I forgot webhooks completely. I have turned off crumb protection on the webhooks endpoints.

Hapi crumb plugin reads the cookie named `crumb` and compares it with the passed crumb.
https://github.com/hapijs/crumb/blob/master/lib/index.js#L167
So it works for multiple instances because of this behavior. I verified it by issuing a new crumb, restarting the server and posting it to a test endpoint.

stjohnjohnson:
👍  
giphy 5

nkatzman:
Are crumb's required when a jwt is being sent?

stjohnjohnson:
Good catch, @nkatzman.  I guess crumbs should only be required when using cookies, JWTs should be enough for CSRF.

catto:
You have a point. I have added `skip` function to check a request is under `webhook` or jwt is used. Also, added tests for it.

nkatzman:
Cool 👍

stjohnjohnson:
👍

stjohnjohnson:
Thanks @catto!  You rock!

catto:
:cat2:

PR #154, opened 2016-09-01
Get jobs for pipeline

nkatzman:
This PR adds in a route to get all jobs for a given pipeline.

It will return 404 when the pipeline does not exist, and will return an array of all jobs.

There isn't any pagination for the route which should be added, but pipeline.jobs only will return 25 jobs (including those that are disabled). 

We need to find a way to filter out all disabled jobs

### DIFF
--- a/plugins/pipelines/listJobs.js
+++ b/plugins/pipelines/listJobs.js
@@ -1,0 +1,32 @@
+'use strict';
+const boom = require('boom');
+const joi = require('joi');
+const schema = require('screwdriver-data-schema');
+const listSchema = joi.array().items(schema.models.job.get).label('List of jobs');
+
+module.exports = () => ({
+    method: 'GET',
+    path: '/pipelines/{id}/jobs',
+    config: {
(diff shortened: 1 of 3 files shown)


plugins/pipelines/listJobs.js
```diff
@@ -0,0 +1,32 @@
...
+const boom = require('boom');
+const joi = require('joi');
+const schema = require('screwdriver-data-schema');
+const listSchema = joi.array().items(schema.models.job.get).label('List of jobs');
+
+module.exports = () => ({
+    method: 'GET',
+    path: '/pipelines/{id}/jobs',
+    config: {
+        description: 'Get all jobs jobs for a given pipelines',
```
Filbird: jobs jobs

Filbird:
😢 👍

plugins/pipelines/listJobs.js
d2lam: Typical Noah... ;)))

d2lam:
Looks good 👍 
giphy

PR #155, opened 2016-09-01
Add settings for ecosystem

petey:
Need to pass ui url to build models for target_url configs.
Also add editor config to define how editors should behave:
http://editorconfig.org/
https://atom.io/packages/editorconfig


stjohnjohnson:
👍  

cat window

PR #157, opened 2016-09-02
Moving all Login stuff into a new route for /v3/auth

stjohnjohnson:
This adds new routes:
 - /v3/auth/login - login via oauth
 - /v3/auth/logout - logout cookie
 - /v3/auth/crumb - crumb for csrf
 - /v3/auth/token - JWT
 - /v3/auth/key - Public key to verify JWT

Note, major change required for your JWT secret (needs to be a RSA keypair)


plugins/auth/index.js
```diff
@@ -0,0 +1,125 @@
...
+            clientSecret: pluginOptions.oauthClientSecret,
+            scope: ['admin:repo_hook', 'read:org', 'repo:status'],
+            isSecure: pluginOptions.https
+        });
+
+        server.auth.strategy('token', 'jwt', {
+            key: pluginOptions.jwtPublicKey,
+            verifyOptions: {
+                algorithms: ['RS256'],
+                maxAge: '12h'
```
FenrirUnbound: Is this supposed to match the value set in line 68?

stjohnjohnson: Yes, it is.  I guess it could be a constant.

FenrirUnbound:
:+1:

Looks good to me

proo7jg

d2lam:
How come we don't bump the version if this is a major change?

stjohnjohnson:
@d2lam Nothing depends on this version right now, so we're not bumping versions on screwdriver main repository until we get to a RC.  After that we'll start with v4.0.0 and follow semver from there.

plugins/auth/index.js
```diff
@@ -0,0 +1,128 @@
...
+        oauthClientId: joi.string().required(),
+        oauthClientSecret: joi.string().required(),
+        jwtPrivateKey: joi.string().required(),
+        jwtPublicKey: joi.string().required(),
+        whitelist: joi.array().default([]),
+        admins: joi.array().default([])
+    }), 'Invalid config for plugin-login');
+
+    /**
+     * Generates a profile for storage in cookie and jwt
```
tkyi: should it be `JWT`? can we be consistent about it?

PR #158, opened 2016-09-03
Using new executor with Launcher docker image

stjohnjohnson:
Pulls in new change from https://github.com/screwdriver-cd/executor-k8s/pull/36


jer:
👍

Issue #148, opened 2016-08-29
Secret Management


stjohnjohnson:
> As a user, I need to have access to my secrets, so I can publish / deploy my application

One of the big blockers with continuous delivery in the existing tools out there is the
availability of secrets to the build. Secrets are needed for publishing packages or code,
deploy to services, and remotely testing if a service is working.

The systems out there usually support secrets in one of three camps:
 - A secret can be available to all people on the shared system
 - A secret can be available just to your pipeline
 - A secret can be available just to a job in your pipeline

Screwdriver should provide a combination of options two and three. This gives developers the
ability to specify secrets that all jobs need to use, and then restrict production secrets
to just the production jobs.

Additionally, developers should be able to customize what secrets are available in a pull
request, even as fine-grained as should it be available in a forked pull-request.

Rules:
 - Saved secrets cannot be viewed by developers, only the name of them should be available
 - Secrets can only be listed, added, changed, or removed by admins of the pipeline
 - Secrets can be strings or generated SSH keys
 - Secrets should be available as environment variables

---
### Preparation
 - [x] Cucumber feature file
 - [x] Secret interface

### Implementation
 - [x] Screwdriver.yaml changes
 - [x] Launcher changes
 - [x] API changes
 - [x] Store changes

### Outcome
 - [x] Screwdriver publishing itself
 - [x] Cucumber feature passing
 - [x] Documentation
 
---
## Notes
 - _nice to have: SSH keys are available in an SSH agent, **not** as environment variables_
 - [Link to Cucumber Feature](https://github.com/screwdriver-cd/screwdriver/blob/fd7f1f02523f15b74a248166717f9bc6d401c467/features/secrets.feature)

d2lam:
```Saved secrets cannot be viewed by developers, only the name of them should be available```

- So they can view the secret names even though they are not admins? 
- Also, they cannot see other fields like `id, pipelineId, allowInPR`?

stjohnjohnson:
I would say that anyone who has `write` access should be able to list the secrets.  Those with `read` should get a 403.  Users with `write` or `admin` should be able to see all properties __EXCEPT__ `value`.

PR #168, opened 2016-09-08
Add /GET secrets/{id} route

minz1027:
- Add /GET secrets/{id} route
- Only users with write permissions can GET all properties of the secret except the value
- Add helper.js with helper functions `userPermission` and `buildPermission`
- The `buildPermission` will be needed later for list

#148 feature

### DIFF
--- a/plugins/secrets/get.js
+++ b/plugins/secrets/get.js
@@ -1,0 +1,56 @@
+'use strict';
+const boom = require('boom');
+const joi = require('joi');
+const schema = require('screwdriver-data-schema');
+const getSchema = schema.models.secret.get;
+const idSchema = joi.reach(schema.models.secret.base, 'id');
+
+module.exports = () => ({
+    method: 'GET',
+    path: '/secrets/{id}',
+    config: {
+        description: 'Get a single secret',
+        notes: 'Returns a secret record',
+        tags: ['api', 'secrets'],
+        auth: {
+            strategies: ['token', 'session'],
+            scope: ['user', 'build']
+        },
+        plugins: {
+            'hapi-swagger': {
+                security: [{ token: [] }]
+            }
+        },
+        handler: (request, reply) => {
+            const secretFactory = request.server.app.secretFactory;
+            const credentials = request.auth.credentials;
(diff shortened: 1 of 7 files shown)


plugins/secrets/get.js
```diff
@@ -0,0 +1,65 @@
...
+                security: [{ token: [] }]
+            }
+        },
+        handler: (request, reply) => {
+            const userFactory = request.server.app.userFactory;
+            const pipelineFactory = request.server.app.pipelineFactory;
+            const secretFactory = request.server.app.secretFactory;
+            const username = request.auth.credentials.username;
+
+            return secretFactory.get(request.params.id)
```
stjohnjohnson: I would use a multi-promise load here instead.  Like: https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/builds/create.js#L34-L37

d2lam: Right. We didn't do that because we were handling the "build" case as well. But since we removed it we should just use Promise.all here

tkyi:
👍 after squashing commits
giphy

plugins/secrets/index.js
```diff
@@ -9,8 +11,59 @@ const createRoute = require('./create');
...
+                        throw boom.notFound(`User ${username} does not exist`);
+                    }
+
+                    return user.getPermissions(pipeline.scmUrl).then(permissions => {
+                        if (!permissions.push) {
+                            throw boom.forbidden(`User ${username} does not have `
+                                + 'write access to this repository');
+                        }
+
+                        return false;
```
minz1027: What `canAccess` actually returns is whether to show the secret value depends on the scope. However just by reading the function name and description, I will think it returns whether the credential can access the secret. Can we document this more clearly?

minz1027: What `canAccess` actually returns is whether to show the secret value depends on the scope. However just by reading the function name and description, I will think it returns whether the credential can access the secret. Can we document this more clearly?

d2lam: @minz1027  Updated

stjohnjohnson:
👍

PR #169, opened 2016-09-08
Exposing more environment variables to customize

stjohnjohnson:
Specifically to use HTTPS and the URL for the UI

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -120,4 +120,5 @@ logging:
     baseUrl: https://s3-us-west-2.amazonaws.com/logs.screwdriver.cd
 
 ecosystem:
+    # URL for the User Interface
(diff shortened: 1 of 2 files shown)


config/default.yaml
```diff
@@ -120,4 +120,5 @@ logging:
     baseUrl: https://s3-us-west-2.amazonaws.com/logs.screwdriver.cd
 
 ecosystem:
+    # URL for the User Interface
     ui: http://cd.screwdriver.cd
```
tkyi: will this need to be https as well?

stjohnjohnson: We COULD default it.  But I changed it here instead: https://github.com/screwdriver-cd/kubernetes/pull/8

jer:
👍

PR #172, opened 2016-09-09
Move CORS out of config and just use the ecosystem.ui value

stjohnjohnson:
Instead of having to change the CORS AND the ecosystem.ui config, I'm combining them.


petey:
👍

PR #173, opened 2016-09-09
Fixing HTTPS with OAuth flow by passing another flag

stjohnjohnson:
Without this, bell tries to read the server protocol (which is http because we have https on the ingress).  And we end up with this error:

> Authentication failed due to: Invalid setting  - isSecure must be set to false for non-https server


petey:
ok 👍 

loo_kat_mee

FenrirUnbound:
👍 

Looks good to me

tumblr_oblcpbcv7k1twrbr9o1_500

PR #174, opened 2016-09-09
Enabling credentials to be passed (like cookies + stuff)

stjohnjohnson:
Specifically this field: `Access-Control-Allow-Credentials`

So the UI can pass the logged in cookie to the API.


d2lam:
👍

petey:
👍

PR #179, opened 2016-09-10
Moving Coveralls and Fixing Docker Trigger

stjohnjohnson:
Publish now builds both latest and the recent tag that was built.


d2lam:
Wow cool ^

PR #181, opened 2016-09-12
Add auth/login/web route

petey:
CORS does not handle redirects well, and javascript is unable to talk to child windows easily. This adds an optional parameter to /auth/login to cause the api to send code to close the window instead of redirecting to /auth/token. The UI can watch for the child window to close, then make an ajax call to /auth/token to get the jwt.


stjohnjohnson:
OOOOOohh, nice. 👍 
giphy

FenrirUnbound:
👍 

Looks fine to me

d2lam:
👍

PR #183, opened 2016-09-13
Build triggers next job in the workflow

FenrirUnbound:
When a build has succeeded -- and is updated as such -- we look up the next build in the workflow and trigger that build prior to returning.

### DIFF
--- a/plugins/builds/update.js
+++ b/plugins/builds/update.js
@@ -20,6 +20,7 @@ module.exports = () => ({
             const factory = request.server.app.buildFactory;
             const id = request.params.id;
             const desiredStatus = request.payload.status;
+            const jobFactory = request.server.app.jobFactory;
             const username = request.auth.credentials.username;
             const scope = request.auth.credentials.scope;
             const isBuild = scope.includes('build');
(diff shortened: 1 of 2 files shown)


plugins/builds/update.js
```diff
@@ -66,10 +67,37 @@ module.exports = () => ({
...
+                        const nextJobName = workflow[workflowIndex + 1];
 
-                    // Update the model in datastore
-                    return build.update();
+                        return jobFactory.get({
+                            name: nextJobName,
+                            pipelineId: pipeline.id
+                        }).then((nextJobToTrigger) => buildFactory.create({
+                            jobId: nextJobToTrigger.id,
+                            username
```
stjohnjohnson: Should we also pass the SHA of the current build?

d2lam: Sure we can

stjohnjohnson:
Please add labels.

stjohnjohnson:
LGTM 👍

tkyi:
👍 LGTM
giphy-1

PR #185, opened 2016-09-14
Attempting to build docker using tags instead of source code

stjohnjohnson:
This uses the undocumented hook feature of Docker Hub to install from a specific version.

Hint here: https://github.com/docker/hub-feedback/issues/508#issuecomment-240616319

This will give us the correct version number when starting these containers.

### DIFF
--- a/hooks/build
+++ b/hooks/build
@@ -1,0 +1,3 @@
+#!/usr/bin/env sh
+
+docker build --build-arg=VERSION=$DOCKER_TAG -t $IMAGE_NAME .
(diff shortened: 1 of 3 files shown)


hooks/build
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env sh
+
+docker build --build-arg=VERSION=$DOCKER_TAG -t $IMAGE_NAME .
```
d2lam: should this be `--build-arg VERSION=$DOCKER_TAG`

https://docs.docker.com/engine/reference/commandline/build/

stjohnjohnson: Both are valid.

PR #186, opened 2016-09-14
Add jobs/{id}/builds route

d2lam:
- This uses the `getBuilds` method from the models
https://github.com/screwdriver-cd/models/pull/98

- Add test for pipeline create for coverage

- This will break until https://github.com/screwdriver-cd/datastore-dynamodb/pull/19 is merged.

Solves https://github.com/screwdriver-cd/screwdriver/issues/187


plugins/jobs/listBuilds.js
```diff
@@ -0,0 +1,55 @@
...
+              .catch(err => reply(boom.wrap(err)));
+        },
+        response: {
+            schema: listSchema
+        },
+        validate: {
+            query: {
+                page: joi.number().integer().positive(),
+                count: joi.number().integer().positive(),
+                sort: joi.string().lowercase().valid(['ascending', 'descending'])
```
minz1027: Would it be better to re-use part of the scan schema? In case later on there are changes on the scan schema.

d2lam: Yeah, let me modify the data-schema first.

package.json
```diff
@@ -63,7 +63,7 @@
     "screwdriver-datastore-imdb": "^1.0.0",
     "screwdriver-executor-k8s": "^7.0.0",
     "screwdriver-executor-l3l": "^1.0.0",
-    "screwdriver-models": "^10.2.0",
+    "screwdriver-models": "^10.3.0",
```
FenrirUnbound: The `^` means it pulls in the latest minor version

stjohnjohnson: True, but the .3 requires at LEAST that version.

plugins/jobs/listBuilds.js
```diff
@@ -0,0 +1,55 @@
...
+              })
+              .then(builds => reply(builds.map(b => b.toJson())))
+              .catch(err => reply(boom.wrap(err)));
+        },
+        response: {
+            schema: listSchema
+        },
+        validate: {
+            query: {
+                page: joi.number().integer().positive(),
```
stjohnjohnson: Isn't there already a data-schema for this?

plugins/jobs/listBuilds.js
```diff
@@ -0,0 +1,55 @@
...
+
+            return factory.get(request.params.id)
+              .then(job => {
+                  if (!job) {
+                      throw boom.notFound('Job does not exist');
+                  }
+
+                  let config = {};
+
+                  if (Object.keys(request.query).length !== 0) {
```
stjohnjohnson: We should never be doing things like this, instead we should be using the default feature of Joi.

plugins/jobs/listBuilds.js
d2lam: Yes, but it was exposed as an object with page and count. You can't access the page or count individually. I modified it so it has sort in there: https://github.com/screwdriver-cd/data-schema/blob/master/api/pagination.js so I'm just gonna change it to use that now

package.json
d2lam: Yup. Because 10.2 does not have `getBuilds()`.

d2lam:
Still waiting for 10.3 to be published:
https://app.wercker.com/Screwdriver/models/runs

Issue #188, opened 2016-09-15
List active jobs for a pipeline: pipelines/{id}/jobs


d2lam:
Modify the current route so that it:
- Only lists active jobs (archived = false)
- Lists jobs in the order they appear in workflow, then open PRs. 
- Archive jobs after they are renamed. Or if it's a PR job, archive if the PR is closed.

This requires changes in this order:
1) data-schema
3) datastore-dynamodb 
4) models
5) API

d2lam:
Verifying feature:
API has 1 opened PR, so it lists main, publish and PR-200
https://api.screwdriver.cd/v3/pipelines/27cec39c9baea6cf64c684479115300d22039b2a/jobs

This PR from ui closed so archived is true:
https://api.screwdriver.cd/v3/jobs/b0d44a42de7dd09c6a0a7f4a81d17d0ddc5961c7

@stjohnjohnson

d2lam:
__

PR #190, opened 2016-09-16
[5] List active jobs

d2lam:
This uses https://github.com/screwdriver-cd/models/pull/99. That needs to be merged before this PR

Solves https://github.com/screwdriver-cd/screwdriver/issues/188


d2lam:
@stjohnjohnson  Could you review again since the function takes in a config now.

PR #191, opened 2016-09-17
Pull CI from a remote gist (need to replace with shared-step)

stjohnjohnson:
Source is here: https://gist.github.com/stjohnjohnson/3d2388b2a7ba658cdcdaffa8cd874e50

This should make it easier to apply this everywhere.  Also uses ssh instead of https token for git push.


screwdriver.yaml
```diff
@@ -4,13 +4,17 @@ jobs:
...
+            - publish-npm: ./ci/publish.sh
+            - publish-docker: ./ci/docker.sh
+        environment:
+            DOCKER_REPO: screwdrivercd/screwdriver
         secrets:
             # Publishing to NPM
             - NPM_TOKEN
             # Pushing tags to Git
-            - GIT_TOKEN
+            - GIT_KEY
```
d2lam: You created this yet? because I didn't.

d2lam: Just checked and saw that you created it.

screwdriver.yaml
```diff
@@ -4,13 +4,17 @@ jobs:
         steps:
             - install: npm install
             - test: npm test
-            - coverage: ./ci/coverage
-            - publish: ./ci/publish
+            - setup-ci: git clone https://gist.github.com/3d2388b2a7ba658cdcdaffa8cd874e50.git ci
+            - coverage: ./ci/coverage.sh
+            - publish-npm: ./ci/publish.sh
```
d2lam: Should we make publish a separate job?

stjohnjohnson: Oooh, that's a GREAT idea

stjohnjohnson: Done.

PR #193, opened 2016-09-17
Adding support for badges per pipeline

stjohnjohnson:
This adds simple badges per pipeline.  It will only show the last build from the `main` job right now.

Looks something like this  and can use alternative badge services other than shields.io.

### DIFF
--- a/plugins/pipelines/badge.js
+++ b/plugins/pipelines/badge.js
@@ -1,0 +1,76 @@
+'use strict';
+const joi = require('joi');
+const schema = require('screwdriver-data-schema');
+const tinytim = require('tinytim');
+const idSchema = joi.reach(schema.models.pipeline.base, 'id');
+
+/**
+ * Generate Badge URL
+ * @method getUrl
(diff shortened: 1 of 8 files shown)


d2lam:
Nice!

plugins/pipelines/badge.js
```diff
@@ -0,0 +1,76 @@
...
+                        return main.getBuilds({
+                            paginate: {
+                                page: 1,
+                                count: 1
+                            },
+                            sort: 'descending'
+                        }).then(builds => {
+                            const build = builds.pop();
+
+                            return reply.redirect(getUrl(badgeService, build && build.status));
```
FenrirUnbound: Doesn't `build && build.status` resolves into a boolean?

FenrirUnbound: Nevermind. Dynamically-typed languages and all that jazz

FenrirUnbound:
Gif for reference: 

l3vrfimkp6fywwb7o

Issue #160, opened 2016-09-07
Store SCM as ID instead of URL


stjohnjohnson:
I propose that we store an `scm_id` in addition to an `scm_url` in the `pipeline` model.  The `id` should be the unique field and `url` just be a display for the user.

We can generate `id` by a call to `scmModel.getID(url)`.  For GitHub that would be done via a call to https://developer.github.com/v3/repos/#get and https://developer.github.com/v3/repos/branches/#get-branch with a return of `123456:master`.  We know that branches are case-sensitive and repositories have a unique identifier.  Additionally, we can access the repository later (regardless of rename) via [https://api.github.com/repositories/:id](http://stackoverflow.com/questions/28778568/how-to-find-the-new-location-of-a-repository-after-a-rename/)

This stems from https://github.com/screwdriver-cd/screwdriver/issues/82

catto:
`id` should include a unique hostname of GitHub instance if SD will support multiple SCMs like GitHub.com and GitHub Enterprise.  I think it would be better to generate `github.com:123456:master`.

stjohnjohnson:
@catto that's a really good point. :+1:

stjohnjohnson:
I love the move forward with this.  I was drafting up some changes to the data-schema that would include a new Repo, Commit, and User structure.  My idea for Repo structure was:

```javascript
const REPO = Joi.object().keys({
    id: Joi.string()
        .required()
        .label('Unique Identifier')
        .example('github.com:123456:master'),

    name: Joi.string()
        .required()
        .label('Display name')
        .example('screwdriver-cd/screwdriver'),

    url: Joi.string()
        .uri()
        .required()
        .label('Link to Repository')
        .example('https://github.com/screwdriver-cd/screwdriver/tree/master')
}).label('SCM Repository');
```

So this method you're writing should take a SCM_URI and output this structure that we can then store for later use.

stjohnjohnson:
So @catto I did some minor refactoring and moved the scmRepo storage to the sync method:
 - https://github.com/screwdriver-cd/data-schema/pull/58
 - https://github.com/screwdriver-cd/models/pull/106
 - https://github.com/screwdriver-cd/screwdriver/pull/204/files

The next steps would be to:
- [ ] Switch key over to `scmRepo.id`
- [ ] Change create to lookup by `id` when checking for duplicates
- [ ] Update GitHub webhooks to use the `id` that is hopefully in the payload
[...]

PR #194, opened 2016-09-17
Store scmId additionally when creating pipelines

catto:
This PR implements the main feature described in #160, blocked by screwdriver-cd/scm-github#8.
Needs to change [this](https://github.com/screwdriver-cd/data-schema/blob/master/models/pipeline.js#L88) if this PR works.

### DIFF
--- a/plugins/pipelines/create.js
+++ b/plugins/pipelines/create.js
@@ -80,7 +80,7 @@ module.exports = (options) => ({
                                 const pipelineConfig = hoek.applyToDefaults(request.payload, {
                                     admins,
                                     scmUrl,
-                                    scmId: repo.id
+                                    scmRepo: repo
                                 });
 
                                 return pipelineFactory.create(pipelineConfig);
--- a/plugins/pipelines/create.js
+++ b/plugins/pipelines/create.js
@@ -48,49 +48,44 @@ module.exports = (options) => ({
             const userFactory = request.server.app.userFactory;
             const scmUrl = formatScmUrl(request.payload.scmUrl);
             const username = request.auth.credentials.username;
-            let user;
 
             // fetch the user
             return userFactory.get({ username })
                 // get the user permissions for the repo
-                .then(u => {
-                    // store the user object for fetching token later
-                    user = u;
+                .then(user => user.getPermissions(scmUrl)
+                    // if the user isn't an admin, reject
+                    .then(permissions => {
+                        if (!permissions.admin) {
+                            throw boom.unauthorized(
+                              `User ${username} is not an admin of this repo`);
+                        }
+                    })
+                    // see if there is already a pipeline
+                    .then(() => pipelineFactory.get({ scmUrl }))
+                    // if there is already a pipeline for the scmUrl, reject
+                    .then(pipeline => {
+                        if (pipeline) {
+                            throw boom.conflict('scmUrl needs to be unique');
+                        }
+                    })
+                    // set up pipeline admins, and create a new pipeline
+                    .then(() => {
+                        const admins = {};
 
-                    return user.getPermissions(scmUrl);
-                })
-                // if the user isn't an admin, reject
-                .then(permissions => {
-                    if (!permissions.admin) {
-                        throw boom.unauthorized(`User ${username} is not an admin of this repo`);
-                    }
-                })
-                // see if there is already a pipeline
-                .then(() => pipelineFactory.get({ scmUrl }))
-                // if there is already a pipeline for the scmUrl, reject
-                .then(pipeline => {
-                    if (pipeline) {
-                        throw boom.conflict('scmUrl needs to be unique');
-                    }
-                })
-                // set up pipeline admins, and create a new pipeline
-                .then(() => {
-                    const admins = {};
+                        admins[username] = true;
 
-                    admins[username] = true;
+                        return user.unsealToken()
+                            .then(token => options.scmPlugin.getRepoId({ scmUrl, token }))
+                            .then(repo => {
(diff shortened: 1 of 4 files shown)


plugins/pipelines/create.js
```diff
@@ -73,12 +79,17 @@ module.exports = () => ({
...
-                        admins,
-                        scmUrl
-                    });
+                    return user.unsealToken()
+                        .then(token => options.scmPlugin.getRepoId({ scmUrl, token }))
+                        .then(repo => {
+                            const pipelineConfig = hoek.applyToDefaults(request.payload, {
+                                admins,
+                                scmUrl,
+                                scmId: repo.id
```
catto: Do I need to store other values such as `name` and `url` from `repo`?

plugins/pipelines/create.js
```diff
@@ -48,11 +48,17 @@ module.exports = () => ({
...
             const username = request.auth.credentials.username;
+            let user;
 
             // fetch the user
             return userFactory.get({ username })
                 // get the user permissions for the repo
-                .then(user => user.getPermissions(scmUrl))
+                .then(u => {
+                    // store the user object for fetching token later
+                    user = u;
```
d2lam: You don't need to do this. You can have the rest nested inside `.then(user =>` scope and `user` will be available.

plugins/pipelines/create.js
d2lam: Don't think so. Wouldn't that information in the scmUrl anyway?

catto: 🆗

plugins/pipelines/create.js
catto: I changed to use `user` object at [this line](https://github.com/screwdriver-cd/screwdriver/pull/194/files#diff-1eb6375f360812e82460d42f926123b1R82). I think it’s out of scope.

d2lam: I mean you can change it to be in the same scope so that you don't need to store the user object. Something like this: 

```
.then(user => user.getPermissions(scmUrl)
    // user is available in this whole scope
    .then(permissions => { ... })
    .
    .
    .
    .then(() => {
       const admins = {};

       admins[username] = true;

       return user.unsealToken()
           .then(token => options.scmPlugin.getRepoId({ scmUrl, token }))
           .then(repo => { ... })
       })
// this is where it ends
)
```

catto: Oh I see. I'll fix it!

plugins/pipelines/create.js
```diff
@@ -52,34 +52,40 @@ module.exports = () => ({
...
-                        admins,
-                        scmUrl
-                    });
+                        return user.unsealToken()
+                            .then(token => options.scmPlugin.getRepoId({ scmUrl, token }))
+                            .then(repo => {
+                                const pipelineConfig = hoek.applyToDefaults(request.payload, {
+                                    admins,
+                                    scmUrl,
+                                    scmId: repo.id
```
stjohnjohnson: See, this is where I would change it to something like `scmRepo: repo`  Store the whole thing.

plugins/pipelines/create.js
```diff
@@ -52,34 +52,40 @@ module.exports = () => ({
...
+                .then(user => user.getPermissions(scmUrl)
+                    // if the user isn't an admin, reject
+                    .then(permissions => {
+                        if (!permissions.admin) {
+                            throw boom.unauthorized(
+                              `User ${username} is not an admin of this repo`);
+                        }
+                    })
+                    // see if there is already a pipeline
+                    .then(() => pipelineFactory.get({ scmUrl }))
```
stjohnjohnson: _not for this pr_ what I would want to do next is use the `scmRepo.id` as the unique hash system next.

plugins/pipelines/create.js
catto: I think we still need to store `repo.id` directly to `pipelineConfig` because the factory [generates `id`](https://github.com/screwdriver-cd/models/blob/master/lib/baseFactory.js#L38) using the [key](https://github.com/screwdriver-cd/data-schema/blob/master/models/pipeline.js#L88).
Is it okay to not change `scmId: repo.id` but add `scmRepo: repo`?

stjohnjohnson: So, yes.  We could store both.  We could also fix the model to support `hoek.reach` for that.  So instead the key is `['scmRepo.id']`.  Also, if you were to merge this, I think the field needs to be in the data-schema: https://github.com/screwdriver-cd/data-schema/blob/master/models/pipeline.js#L7-L35

catto: I god it. I'm going to change this line to store scmRepo and fix the model to support `hoek.reach`.

catto:
Hmm.. tests passed but the build marked as failed

d2lam:
@catto  That's a bug. Because there is a change on the master since this PR was created, and we don't do sync for PR. 
@stjohnjohnson  do you have an issue to track this?

catto:
@d2lam I see 🙆

plugins/pipelines/create.js
stjohnjohnson: Since you changed this in data-schema, you'll prob. want to change this here as well.  Sadly this PR is getting bigger and bigger.  Sorry @catto  :(

stjohnjohnson:
Wait, scratch my previous idea @catto.  Leave this PR alone.  I approve it.

Instead, update https://github.com/screwdriver-cd/data-schema/pull/56 to not change the key.  That way it won't be backwards incompatible and we can merge both and then start on the migration.

catto:
Okay, I'll do it!

PR #196, opened 2016-09-19
No longer return 404 if pipeline doesn't exist

stjohnjohnson:
This fixes the constant error state that the webhooks are in:

screen shot 2016-09-19 at 9 45 09 am

### DIFF
--- a/test/plugins/webhooks/github.test.js
+++ b/test/plugins/webhooks/github.test.js
@@ -14,6 +14,8 @@ const PARSED_CONFIG = require('../data/github.parsedyaml.json');
 
 sinon.assert.expose(assert, { prefix: '' });
 
+require('sinon-as-promised');
+
 describe('github plugin test', () => {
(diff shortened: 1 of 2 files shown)


d2lam:
In what kind of event that webhooks hits this endpoint and cannot find a pipeline?
Like the pipeline was not created yet, but there is a PR for it?

test/plugins/webhooks/github.test.js
```diff
@@ -14,6 +14,8 @@ const PARSED_CONFIG = require('../data/github.parsedyaml.json');
 
 sinon.assert.expose(assert, { prefix: '' });
 
+require('sinon-as-promised');
```
FenrirUnbound: Did omitting this not cause a test failure...?

stjohnjohnson:
@d2lam the most basic example is a push event to a non-master branch.

PR #202, opened 2016-09-21
[prom-5] Bumping to Promise versions and removing Local support

stjohnjohnson:
Depends on the previous tagged builds.  Also removing local as we don't have log-service or promises in there yet.


package.json
```diff
@@ -64,9 +64,8 @@
     "screwdriver-data-schema": "^12.0.0",
     "screwdriver-datastore-dynamodb": "^2.1.0",
     "screwdriver-datastore-imdb": "^1.0.0",
-    "screwdriver-executor-k8s": "^7.0.0",
-    "screwdriver-executor-l3l": "^1.0.0",
-    "screwdriver-models": "^10.3.0",
+    "screwdriver-executor-k8s": "^8.0.0",
+    "screwdriver-models": "^11.0.0",
```
d2lam: Why not 12?

stjohnjohnson: Because there isn't one?  I'm just bumping to 11 in this PR: https://github.com/screwdriver-cd/models/pull/104

d2lam: Lol sorry I think it was kinda late at night. Not sure how I saw "12"... ?!?

PR #210, opened 2016-09-22
Allow auth_token login alternative

FenrirUnbound:
Enable an alternative to the OAuth flow for acquiring a valid JWT. This PR is a start, where there is only a single `access_token`. Ideally, we want to allocate `access token`s per user.

Looking for feedback on the direction of the work. This is a frankenstein of a couple of PoCs, so it's rough on a few details.

### DIFF
--- a/config/custom-environment-variables.yaml
+++ b/config/custom-environment-variables.yaml
@@ -15,6 +15,12 @@ auth:
     oauthClientId: SECRET_OAUTH_CLIENT_ID
     # The client secret used for OAuth with github
     oauthClientSecret: SECRET_OAUTH_CLIENT_SECRET
+    # The access token to use on behalf of a user to access the API. Used as
+    # an alternative to the OAuth flow
+    temporaryAccessKey: SECRET_ACCESS_KEY
+    # The user name associated with the temproary access token. Used as a
+    # means of functionally testing the API
+    temporaryAccessUser: SECRET_ACCESS_USER
     # A password used for encrypting session, and OAuth data.
     # **Needs to be minimum 32 characters**
     password: SECRET_PASSWORD
--- a/plugins/auth/index.js
+++ b/plugins/auth/index.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const authToken = require('hapi-auth-bearer-token');
 const bell = require('bell');
 const sugar = require('hapi-auth-cookie');
 const authjwt = require('hapi-auth-jwt');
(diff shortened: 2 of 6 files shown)


plugins/auth/index.js
```diff
@@ -112,6 +116,19 @@ exports.register = (server, options, next) => {
...
+        server.auth.strategy('auth_token', 'bearer-access-token', {
+            accessTokenName: 'access_token',
+            allowCookieToken: false,
+            allowQueryToken: true,
+            validateFunc: (token, cb) => {
+                if (token !== pluginOptions.temporaryAccessToken) {
+                    return cb(null, false, {});
+                }
+
+                return cb(null, true, { username: '???????', scope: ['user'] });
```
FenrirUnbound: * Would it be appropriate to use our team user?
* What would be an appropriate scope for them?

stjohnjohnson: I would have this customizable via the config

plugins/auth/index.js
```diff
@@ -112,6 +116,19 @@ exports.register = (server, options, next) => {
             }
         });
 
+        server.auth.strategy('auth_token', 'bearer-access-token', {
+            accessTokenName: 'access_token',
```
stjohnjohnson: token seems overloaded, can we use `access_key` instead?

FenrirUnbound: Updated everything to access key 👍

plugins/auth/index.js
FenrirUnbound: Configured the config to be configurable by the local configuration files

config/custom-environment-variables.yaml
```diff
@@ -15,6 +15,12 @@ auth:
     oauthClientId: SECRET_OAUTH_CLIENT_ID
     # The client secret used for OAuth with github
     oauthClientSecret: SECRET_OAUTH_CLIENT_SECRET
+    # The access token to use on behalf of a user to access the API. Used as
+    # an alternative to the OAuth flow
+    temporaryAccessToken: SECRET_ACCESS_KEY
+    # The user name associated with the temproary access token. Used as a
+    # means of functionally testing the API
+    temporaryAccessToken: SECRET_ACCESS_USER
```
stjohnjohnson: Name is wrong here

config/custom-environment-variables.yaml
```diff
@@ -15,6 +15,12 @@ auth:
     oauthClientId: SECRET_OAUTH_CLIENT_ID
     # The client secret used for OAuth with github
     oauthClientSecret: SECRET_OAUTH_CLIENT_SECRET
+    # The access token to use on behalf of a user to access the API. Used as
+    # an alternative to the OAuth flow
+    temporaryAccessToken: SECRET_ACCESS_KEY
```
stjohnjohnson: Name is wrong here as well

FenrirUnbound:
I'll confirm when I edit the configuration for our API deployment 😺

FenrirUnbound:
API deployment configuration has been updated!

PR #218, opened 2016-09-27
Authorize user "calvin" for functional test

tkyi:
Using new auth_token/access_key login alternative

Relates to https://github.com/screwdriver-cd/screwdriver/pull/210


d2lam:
So we are using sd-buildbot's access token, which has all permissions to our repo
`{ admin: true, push: true, pull: true }`
, but the test requires `susie` to have `read-only` permission: https://github.com/screwdriver-cd/screwdriver/blob/master/features/authorization.feature#L19-L25

How do we ensure that? I also don't know how we can make `calvin` an `admin`, `hobbes` a `contributor` since they are supposed to be github users.

FenrirUnbound:
@d2lam I don't think that's the case at all. Currently `sd-buildbot` for [screwdriver-cd-test/functional-git](https://github.com/screwdriver-cd-test/functional-git) only has read-only access.

As for the other users, we most likely will need users that fulfill the criteria.

d2lam:
Are you sure? I tried:
```
github.authenticate({
     type: "oauth",
     token: "....sd build bot token..."
 });

  github.repos.get({
      'user': 'screwdriver-cd-test',
      'repo': 'functional-git'
  }, function(err, res) {
      console.log(res.permissions);
  });
```

that gave me `{ admin: true, push: true, pull: true }`

features/step_definitions/authorization.js
```diff
@@ -0,0 +1,38 @@
...
+        if (user !== 'susie') {
+            throw new Error('User is not "susie"');
+        }
+
+        return request({
+            method: 'GET',
+            url: `https://api.screwdriver.cd/v4/auth/token?access_key=${this.accessKey}`,
+            followAllRedirects: true
+        }).then((response) => {
+            accessToken = JSON.parse(response.body).token;
```
d2lam: You won't need to do `JSON.parse` if you put `json:true` in the request

FenrirUnbound:
@d2lam yeah, that's because we had to add it for the `git-flow` functional tests. Initially, the build bot was not added to the repository as an owner, so it was completely valid at the time of writing.

features/step_definitions/authorization.js
FenrirUnbound: Addressed

PR #220, opened 2016-09-28
Add functional test for secrets scenario 1

minz1027:
- Add cucumber test for secrets.feature scenario 1
- [Repo for testing](https://github.com/screwdriver-cd-test/functional-secrets)
- Remove `And the build succeeded` since it is already checked in `Then the "foo" secret should be available in the build`
- PipelineId is hardcoded until we find a better way to do it. 

issue [#148](https://github.com/screwdriver-cd/screwdriver/issues/148)


features/secrets.feature
```diff
@@ -36,7 +36,6 @@ Feature: Secrets
         When a secret "foo" is added globally
         And the "main" job is started
         Then the "foo" secret should be available in the build
-        And the build succeeded
```
FenrirUnbound: Why is the success of the build not important?

features/step_definitions/secrets.js
```diff
@@ -0,0 +1,131 @@
...
+    return err || body.status === 'QUEUED' || body.status === 'RUNNING';
+}
+
+module.exports = function server() {
+    this.setDefaultTimeout(60000);
+    let waitForBuild;
+    let options;
+
+    // eslint-disable-next-line new-cap
+    this.Before(() => {
```
FenrirUnbound: This runs before all scenarios, not just the ones defined for secrets

features/step_definitions/secrets.js
```diff
@@ -0,0 +1,131 @@
...
+                uri: `${this.instance}/${this.namespace}/builds/${buildID}`,
+                method: 'GET',
+                maxAttempts: 10,
+                retryDelay: 5000,
+                retryStrategy: buildRetryStrategy
+            }));
+    });
+
+    // eslint-disable-next-line new-cap
+    this.After(() =>
```
FenrirUnbound: This `.After()` hook runs after all scenarios; even the ones that are not defined in secrets.feature

features/step_definitions/secrets.js
```diff
@@ -0,0 +1,131 @@
...
+        })).then((response) => {
+            this.pipelineId = response.body.id;
+
+            Assert.equal(response.statusCode, 201);
+        }).then(() => table)
+    );
+
+    this.Given(/^an existing pipeline with that repository with the workflow:$/, (table) => table);
+
+    this.Given(/^"calvin" is logged in$/, () => null);
```
FenrirUnbound: The scenarios are defined with a matrix table. You should expand this to capture all the users defined in the matrix table instead of defining each one individually

features/step_definitions/secrets.js
minz1027: removed

features/secrets.feature
d2lam: We said in the description of the PR:
`Remove And the build succeeded since it is already checked in Then the "foo" secret should be available in the build`

features/step_definitions/secrets.js
minz1027: removed that line

features/step_definitions/secrets.js
minz1027: fixed

d2lam:
Closing since this branch is hella messed up and we don't know how to fix it :(

PR #223, opened 2016-09-29
Add functional test for secrets scenario 1 and 4

minz1027:
- Add cucumber test for secrets.feature scenario 1 and 4 (need more users for other scenarios)
- In scenario 1, remove `And the build succeeded` since it is already checked in `Then the "foo" secret should be available in the build`
- PipelineId is hardcoded until we find a better way to do it
- [Repo used for testing](https://github.com/screwdriver-cd-test/functional-secrets)

issue [#148](https://github.com/screwdriver-cd/screwdriver/issues/148)

### DIFF
--- a/features/step_definitions/git-flow.js
+++ b/features/step_definitions/git-flow.js
@@ -157,7 +157,7 @@ function waitForBuild(screwdriverInstance, desiredSha) {
 module.exports = function server() {
     // eslint-disable-next-line new-cap
     this.Before(() => {
-        this.instance = 'http://api.screwdriver.cd';
+        this.instance = 'https://api.screwdriver.cd';
         this.repositoryOwner = 'screwdriver-cd';
         this.repository = 'garbage-repository-ignore-this';
         this.testBranch = 'testBranch';
--- a/features/step_definitions/secrets.js
+++ b/features/step_definitions/secrets.js
@@ -3,8 +3,6 @@ const Assert = require('chai').assert;
 const request = require('../support/request');
 
 module.exports = function server() {
-    this.setDefaultTimeout(60000);
-
     this.Given(/^an existing repository for secret with these users and permissions:$/, (table) => {
         this.pipelineId = '4de9888518fd74beb336eb6e36f68b25697c219f';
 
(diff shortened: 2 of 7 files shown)


features/step_definitions/secrets.js
```diff
@@ -0,0 +1,128 @@
+'use strict';
+const Assert = require('chai').assert;
+const request = require('../support/request');
+
+module.exports = function server() {
+    this.setDefaultTimeout(60000);
```
FenrirUnbound: This affects timeouts for ***all*** tests

features/support/before-hook.js
```diff
@@ -10,6 +23,18 @@ function beforeHooks() {
...
+        this.namespace = 'v4';
+        this.waitForBuild = (buildID) =>
+            requestretry({
+                uri: `${this.instance}/${this.namespace}/builds/${buildID}`,
+                method: 'GET',
+                maxAttempts: 10,
+                retryDelay: 5000,
+                retryStrategy: buildRetryStrategy,
+                json: true
+            });
```
tkyi: 👍 nice!

features/step_definitions/git-flow.js
```diff
@@ -157,7 +157,7 @@ function waitForBuild(screwdriverInstance, desiredSha) {
 module.exports = function server() {
     // eslint-disable-next-line new-cap
     this.Before(() => {
-        this.instance = 'http://api.screwdriver.cd';
```
FenrirUnbound: No need to modify this feature set

d2lam: Had to. Its `Before` applies to all the scenarios. We could move this whole block to `before-hooks` but didn't want to mess with git-flow

features/step_definitions/secrets.js
d2lam: Thanks for telling me how to fix this 👍

features/step_definitions/git-flow.js
FenrirUnbound: 👍

PR #226, opened 2016-09-30
Removing /v4/builds and /v4/jobs as requesting everything isn't needed

stjohnjohnson:
All of our requests for builds should go through /v4/jobs/:id/builds
All of our requests for jobs should go through /v4/pipelines/:id/jobs


tkyi:
So we'll always need `pipelineId` and/or `jobId` to get jobs or builds from now on?

stjohnjohnson:
Yup!

tkyi:
Can you remove the routes from the READMEs too?

PR #230, opened 2016-09-30
DELETE should return 204 because of no content

d2lam:
- `200` expects some content there, so we should return `204` instead
- Update old tests to promise form

### DIFF
--- a/plugins/pipelines/remove.js
+++ b/plugins/pipelines/remove.js
@@ -43,7 +43,8 @@ module.exports = () => ({
                     })
 
                     // user has good permissions, remove the pipeline
-                    .then(() => reply(pipeline.remove()));
+                    .then(() => pipeline.remove())
+                    .then(() => reply().code(204));
             })
             .catch(err => reply(boom.wrap(err)));
         },
--- a/test/plugins/pipelines.test.js
+++ b/test/plugins/pipelines.test.js
@@ -189,10 +189,10 @@ describe('pipeline plugin test', () => {
             };
         });
 
-        it('returns 200 and all pipelines', (done) => {
+        it('returns 200 and all pipelines', () => {
             pipelineFactoryMock.list.resolves(getPipelineMocks(testPipelines));
 
-            server.inject(options, (reply) => {
+            return server.inject(options).then(reply => {
                 assert.equal(reply.statusCode, 200);
                 assert.deepEqual(reply.result, testPipelines);
                 assert.calledWith(pipelineFactoryMock.list, {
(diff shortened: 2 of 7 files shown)


plugins/pipelines/remove.js
```diff
@@ -43,7 +43,7 @@ module.exports = () => ({
                     })
 
                     // user has good permissions, remove the pipeline
-                    .then(() => reply(pipeline.remove()));
+                    .then(() => reply(pipeline.remove()).code(204));
```
stjohnjohnson: Shouldn't this be something like:
`.then(pipeline.remove).then(() => reply().code(204));`

test/plugins/pipelines.test.js
```diff
@@ -290,9 +290,9 @@ describe('pipeline plugin test', () => {
             pipelineFactoryMock.get.withArgs(id).resolves(pipeline);
         });
 
-        it('returns 200 when delete successfully', (done) => {
+        it('returns 204 when delete successfully', (done) => {
             server.inject(options, (reply) => {
-                assert.equal(reply.statusCode, 200);
+                assert.equal(reply.statusCode, 204);
                 done();
```
stjohnjohnson: plz test that delete was called

test/plugins/pipelines.test.js
```diff
@@ -290,9 +290,9 @@ describe('pipeline plugin test', () => {
             pipelineFactoryMock.get.withArgs(id).resolves(pipeline);
         });
 
-        it('returns 200 when delete successfully', (done) => {
+        it('returns 204 when delete successfully', (done) => {
             server.inject(options, (reply) => {
```
stjohnjohnson: can you use the promise form instead of the callback?

test/plugins/secrets.test.js
```diff
@@ -267,9 +267,9 @@ describe('secret plugin test', () => {
             });
         });
 
-        it('returns 200 if remove successfully', (done) => {
+        it('returns 204 if remove successfully', (done) => {
             server.inject(options, (reply) => {
```
stjohnjohnson: can you use the promise form instead of the callback?

plugins/pipelines/remove.js
d2lam: Fixed

test/plugins/pipelines.test.js
d2lam: 👍  I fixed other tests too

test/plugins/pipelines.test.js
d2lam: Added

PR #234, opened 2016-10-03
Support overriding the dynamodb prefix via environment variables

stjohnjohnson:
This allows us to use `beta_` for our beta environment


jer:
👍

PR #236, opened 2016-10-03
Deploy to beta first, then production

stjohnjohnson:
This ensures we test beta environment first and don't accidentally break production


jer:
👍

Issue #232, opened 2016-10-03
Update a secret


minz1027:
- [x] Add secret `update` schema to [data-schema](https://github.com/screwdriver-cd/data-schema)
- [x] Add `update` route to [Screwdriver API](https://github.com/screwdriver-cd/screwdriver)
- [x] Modify current `update`  to save dirty fields only in [UI](https://github.com/screwdriver-cd/ui)

PR #238, opened 2016-10-03
 Add update secret route

minz1027:
- Add `update` route for secrets
- Only user can `update` a secret
- Modify `canAccess` helper function to take in required permission
- Update README

Issue #232

### DIFF
--- a/plugins/secrets/README.md
+++ b/plugins/secrets/README.md
@@ -26,7 +26,11 @@ server.register({
 ```
 
 ### Routes
-#### Get a single secret
+`Get` requires **write** permission to the repository.
+
+`Create`, `Remove` and `Update` require **admin** permission to the repository.
+
+#### Get a secret
 
 `GET /secrets/{id}`
 
@@ -46,6 +50,28 @@ Example payload:
 {
   "pipelineId": "d398fb192747c9a0124e9e5b4e6e8e841cf8c71c",
   "name": "NPM_TOKEN",
+  "value": "batman",
+  "allowInPR": true
+}
+```
+#### Remove a secret
+
+`DELETE /secrets/{id}`
+
+#### Update a secret
+
+`PUT /secrets/{id}`
+
+**Arguments**
+
+Only `value` and `allowInPR` can be updated.
+
+* `value` - Value of the secret.
+* `allowInPR` - Flag to denote if this secret can be shown in PR builds.
+
+Example payload:
+```json
+{
   "value": "batman",
   "allowInPR": true
 }
--- a/plugins/secrets/index.js
+++ b/plugins/secrets/index.js
@@ -2,6 +2,7 @@
 const createRoute = require('./create');
 const getRoute = require('./get');
 const removeRoute = require('./remove');
+const updateRoute = require('./update');
 const boom = require('boom');
 
 /**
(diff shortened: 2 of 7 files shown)


d2lam:
Blocked by https://github.com/screwdriver-cd/data-schema/pull/62

plugins/secrets/update.js
```diff
@@ -0,0 +1,54 @@
...
+            const credentials = request.auth.credentials;
+            const canAccess = request.server.plugins.secrets.canAccess;
+
+            return factory.get(request.params.id)
+                .then(secret => {
+                    if (!secret) {
+                        throw boom.notFound('Secret does not exist');
+                    }
+
+                    Object.keys(request.payload).forEach(key => {
```
d2lam: Should probably move this to after canAccess, right before `secret.update()`. No point of doing this if user doesn't have permission.

minz1027: fixed

plugins/secrets/update.js
```diff
@@ -0,0 +1,56 @@
...
+            const credentials = request.auth.credentials;
+            const canAccess = request.server.plugins.secrets.canAccess;
+
+            return factory.get(request.params.id)
+                .then(secret => {
+                    if (!secret) {
+                        throw boom.notFound('Secret does not exist');
+                    }
+
+                    // Make sure that user/build has permission before updating
```
stjohnjohnson: builds cannot update secrets

plugins/secrets/update.js
```diff
@@ -0,0 +1,56 @@
...
+                    return canAccess(credentials, secret)
+                        .then(() => {
```
stjohnjohnson: You're not doing anything with the result.  But even so, that check only checks for write access (which is not the same as admin for create).  I would assume you would need admin to update if you need admin to create.

plugins/secrets/update.js
minz1027: fixed

plugins/secrets/update.js
stjohnjohnson: What do you think should be the appropriate permission level?  Who should be creating/updating/deleting secrets?  Whichever we pick, we should document.

d2lam: Hmmm I think admin access for create, update and remove. write access for get. We can modify canAccess to do this. 
If we want to simplify we can make everything requires admin access.

minz1027: @stjohnjohnson I agree with admin access for `create`, `update` and `remove`. write access for `get` and `list`. Have modified the `canAccess` function to take care of these 2 permission levels. Plz help review again. :)

plugins/secrets/index.js
```diff
@@ -42,6 +49,11 @@ exports.register = (server, options, next) => {
                     }
 
                     return user.getPermissions(pipeline.scmUrl).then(permissions => {
+                        if (permission === 'admin' && !permissions.admin) {
```
d2lam: You can do this so you don't need to split into multiple checks for admin and push:
```js 
if (!permisions[requiredPerm]) {
   throw boom.forbidden(`User ${username} does not have ${requiredPerm} access to this repo`)
``` 
where `requiredPerm` is the param passed in.

minz1027: fixed

plugins/secrets/README.md
```diff
@@ -26,7 +26,11 @@ server.register({
 ```
 
 ### Routes
-#### Get a single secret
+`Get` requires **write** permission.
```
d2lam: permission to the repository to be more clear?

plugins/secrets/README.md
```diff
@@ -26,7 +26,11 @@ server.register({
...
+
+`Create`, `Remove` and `Update` require **admin** permission.
```
d2lam: same here

plugins/secrets/README.md
minz1027: fixed

PR #239, opened 2016-10-03
PR sync reads from its branch

d2lam:
- This allows synchronize event to read from PR ref instead of master. 
- Remove `getBuildsForJobsId` and replace with `job.getRunningBuilds()`
- Still need to keep this check: https://github.com/screwdriver-cd/screwdriver/blob/da6cf99d835bc8750fe4e6e6792f1b176cd944d8/plugins/webhooks/github.js#L14-L16 just in case the build finishes in between the time it gets the running builds and time stopRunningBuild is called. 
- Blocked by https://github.com/screwdriver-cd/models/pull/113
- Relates to https://github.com/screwdriver-cd/screwdriver/issues/241

### DIFF
--- a/plugins/webhooks/github.js
+++ b/plugins/webhooks/github.js
@@ -5,23 +5,34 @@ const hoek = require('hoek');
 const boom = require('boom');
 
 /**
+ * Stop a job by stopping all the builds associated with it
  * If the build is running, set state to ABORTED
- * @method stopRunningBuild
- * @param  {Build}   build Build to stop
+ * @method stopJob
+ * @param  {Job}    job     Job to stop
  * @return {Promise}
  */
-function stopRunningBuild(build) {
-    if (build.isDone()) {
-        return Promise.resolve();
+function stopJob(job) {
+    if (!job) {
+        throw boom.notFound('Job does not exist');
     }
-    build.state = 'ABORTED';
 
-    return build.update();
+    const stopRunningBuild = (build) => {
+        if (build.isDone()) {
+            return Promise.resolve();
+        }
+        build.state = 'ABORTED';
+
+        return build.update();
+    };
+
+    return job.getRunningBuilds()
+        // Stop running builds
+        .then(builds => Promise.all(builds.map(stopRunningBuild)));
 }
 
 /**
- * Create a new job and start the build for an opened pull-request
- * @method pullRequestOpened
+ * Run pull request's main job
+ * @method startPRJob
  * @param  {Object}       options
  * @param  {String}       options.eventId       Unique ID for this GitHub event
  * @param  {String}       options.pipelineId    Identifier for the Pipeline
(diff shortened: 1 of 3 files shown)


plugins/webhooks/github.js
```diff
@@ -38,6 +38,7 @@ function pullRequestOpened(options, request, reply) {
     const buildFactory = request.server.app.buildFactory;
     const eventId = options.eventId;
     const pipelineId = options.pipelineId;
+    const jobId = options.jobId;
```
petey: I'm not clear how the jobId is passed here when the job hasn't been created yet.

plugins/webhooks/github.js
```diff
@@ -85,69 +84,91 @@ function pullRequestOpened(options, request, reply) {
...
-        .then(builds => Promise.all(builds.map(stopRunningBuild)))
-        // log build stoppage
+            return job.getRunningBuilds()
+                // Stop running builds
+                .then(builds => Promise.all(builds.map(stopRunningBuild)));
+        })
         .then(() => {
             request.log(['webhook-github', eventId, jobId], `${name} stopped`);
         })
+        .then(() => pipeline.getConfiguration(ref))
```
petey: This block of code looks identical to the code in `pullRequestOpened`. Refactor?

plugins/webhooks/github.js
d2lam: `jobId` is just the hash of `pipelineId` and `name`. Since we already generate here: https://github.com/screwdriver-cd/screwdriver/blob/da6cf99d835bc8750fe4e6e6792f1b176cd944d8/plugins/webhooks/github.js#L234

we might as well pass it in?

plugins/webhooks/github.js
d2lam: Done!

plugins/webhooks/github.js
```diff
@@ -30,14 +30,15 @@ function stopRunningBuild(build) {
  * @param  {String}       options.username      User who created the PR
  * @param  {String}       options.prRef         Reference to pull request
  * @param  {Pipeline}     options.pipeline      Pipeline model for the pr
- * @param  {Hapi.request} request Request from user
- * @param  {Hapi.reply}   reply   Reply to user
+ * @param  {Hapi.request} request               Request from user
+ * @return {Promise}
  */
-function pullRequestOpened(options, request, reply) {
+function runMainJob(options, request) {
```
petey: I think the name of this function is a little unfortunately confusing, since we actually require a `main` job in every project, and this isn't actually running that job. `startPRJob` might be more accurate.

plugins/webhooks/github.js
```diff
@@ -85,71 +96,72 @@ function pullRequestOpened(options, request, reply) {
...
     const name = options.name;
-    const username = options.username;
-    const sha = options.sha;
     const jobId = options.jobId;
 
-    return buildFactory.getBuildsForJobId({ jobId })
-        // stop all running builds
-        .then(builds => Promise.all(builds.map(stopRunningBuild)))
-        // log build stoppage
+    return jobFactory.get(jobId)
```
petey: Like `runMainJob` or `startPRJob` or whatever it gets named, the code for stopping all running builds in the PR is the same in both sync and closed handlers.

d2lam:
@petey I refactored it further. Can you please review again? :)

plugins/webhooks/github.js
```diff
@@ -50,24 +62,33 @@ function pullRequestOpened(options, request, reply) {
...
-        .then(jobId => buildFactory.create({ jobId, sha, username }))
-        // log the build created/started
-        .then(build => {
-            request.log(['webhook-github', eventId, build.jobId, build.id], `${name} started `
-                + `${build.number}`);
-        })
+        .then(() => buildFactory.create({ jobId, sha, username }))
+        .then(build =>
+            request.log(['webhook-github', options.eventId, build.jobId, build.id],
+            `${options.name} started ${build.number}`));
```
minz1027: why not just `${name}`?

PR #247, opened 2016-10-06
Adding IMDB back into the mix

stjohnjohnson:
Since @FenrirUnbound was kind enough to fix the IMDB class, we can import it again.


d2lam:
Need to make it use model v13 https://github.com/screwdriver-cd/datastore-imdb/blob/master/package.json#L43

stjohnjohnson:
Ugh, good catch.  I can wait.  The docker image just doesn't work out of the box anymore because of it.

PR #256, opened 2016-10-10
Use new Datastore setup as well as no `plugin`

stjohnjohnson:
Uses the new `setup` from Datastore, removes the `Plugin` in `scmPlugin`, refactors some internal callbacks as promises, and migrates away from the custom log url into ecosystem.

This requires:
 - https://github.com/screwdriver-cd/models/pull/118
 - https://github.com/screwdriver-cd/datastore-dynamodb/pull/29
 - https://github.com/screwdriver-cd/datastore-base/pull/17

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -49,7 +49,7 @@ function prettyPrintErrors(request, reply) {
  * @param  {Function}    callback                   Callback to invoke when server has started.
  * @return {http.Server}                            A listener: NodeJS http.Server object
  */
-module.exports = (config, callback) => {
+module.exports = (config) => {
     // Hapi Cross-origin resource sharing configuration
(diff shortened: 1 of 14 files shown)


lib/server.js
```diff
@@ -91,23 +91,18 @@ module.exports = (config, callback) => {
...
+                    server.plugins.auth.generateToken(
+                        server.plugins.auth.generateProfile(buildId, ['build'], metadata)
+                    );
+            }
 
-        return 0;
-    });
+            // Start the server
+            return server.start().then(() => server);
+        });
```
minz1027: Why we do not need any error handling here?

stjohnjohnson: It's a promise, so the rejection is passed all the way back.

PR #257, opened 2016-10-10
Add working auth tests for user Calvin

tkyi:
We only have an access key working for the user Calvin at the moment, so adding a fully working scenario for that user here.

**Not sure where pipeline creation should happen.

### DIFF
--- a/features/step_definitions/authorization.js
+++ b/features/step_definitions/authorization.js
@@ -4,6 +4,15 @@ const request = require('../support/request');
 const jwt = require('jsonwebtoken');
 
 module.exports = function server() {
+    // eslint-disable-next-line new-cap
+    this.Before({
+        tags: ['@auth']
+    }, () => {
+        this.repoOrg = 'screwdriver-cd-test';
+        this.repoName = 'functional-auth';
+        this.pipelineId = null;
+    });
+
     this.Given(/^an existing repository with these users and permissions:$/, (table) => table);
 
     this.Given(/^an existing pipeline with that repository$/, () => null);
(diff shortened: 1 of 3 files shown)


features/step_definitions/authorization.js
```diff
@@ -37,4 +46,82 @@ module.exports = function server() {
             return null;
         });
     });
+
+    this.Then(/^they can see the project$/, () =>
+        request({                           // make sure pipeline exists
```
FenrirUnbound: > make sure the pipeline exists

Wouldn't this be better placed in the *Given an existing pipeline with that repository* step definition?

tkyi: I wasn't sure how to create a pipeline any earlier without the user's jwt.
After discussing with @FenrirUnbound , we decided it would make sense to do it in the _Given an existing pipeline with that repository_ in the future. We will need to make the call to get a jwt with the admin's **accessKey** before creating a pipeline.

Since we don't have any other users yet, we will wait to do this refactoring.

features/step_definitions/authorization.js
```diff
@@ -37,4 +46,82 @@ module.exports = function server() {
...
+            }).then((resp) => {
+                this.buildId = resp.body.id;
+
+                Assert.equal(resp.statusCode, 201);
+            })
+        )
+    );
+
+    this.Then(/^they can delete the pipeline$/, { timeout: 60 * 1000 }, () =>
+        this.waitForBuild(this.buildId).then(response => {
```
d2lam: Why do we need this for deleting pipeline?

tkyi: I was worried things wouldn't be cleaned up properly we tried to delete the pipeline when the build was still building.
Will remove since it doesn't seem to be an issue.

PR #275, opened 2016-10-14
requesting access to tonytamsf @ Splunk

tonytamsf:
Just to poke around and kick the tires


stjohnjohnson:
Hey @tonytamsf, sure.  We'd be happy to let you on our hosted instance.  Be aware, we're doing some major changes in the next few days (to support BitBucket) and it will cause pipelines to be recreated.

PR #280, opened 2016-10-15
Refactor API to use scmUri

stjohnjohnson:
This is so @tkyi can continue working on moving to scmUri


tkyi:
Need to migrate all the tables to use the new data-schema first

minz1027:
Fixed the functional test to create pipelines dynamically!

features/step_definitions/authorization.js
```diff
@@ -38,7 +33,7 @@ module.exports = function server() {
 
             switch (user) {
             case 'calvin':
-                Assert.strictEqual(decodedToken.username, 'sd-buildbot');
+                Assert.strictEqual(decodedToken.username, 'minz1027');
```
d2lam: 'sd-buildbot'

tkyi:
Cloned prod `pipeline` and `secret` tables to `rc2_`

features/step_definitions/git-flow.js
```diff
@@ -200,4 +186,22 @@ module.exports = function server() {
             Assert.oneOf(data.state, ['success', 'pending']);
         })
     );
+
+    // eslint-disable-next-line new-cap
+    this.After({
```
FenrirUnbound: Just keep in mind that the `.After` block is not guaranteed

minz1027: Cannot figure out a better way to do it. :(  We don't know the pipelineId before we create it in the `.Before` hook

FenrirUnbound: Do your tests care whether a pipeline was recently created or already exists? Can they tell the difference?

tkyi: Our tests don't care, but we need the `pipelineId`. The alternative (to creating a pipeline every time) would be searching through a list of pipelines for a particular `scmUri` after making a call to the scm to get a `repoId` to construct an `scmUri`.

PR #286, opened 2016-10-19
Fix functional tests to create pipelines dynamically

tkyi:
Currently, `pipelineId` is hardcoded. This PR will create pipelines dynamically instead.

Some caveats:
- deletes and creates a new pipeline with each scenario; not sure if there's a better way to handle this issue???

Blocks https://github.com/screwdriver-cd/screwdriver/pull/280


tkyi:
Moving contents of this PR to other PR (https://github.com/screwdriver-cd/screwdriver/pull/280) which pulls in correct package versions

Closing this PR

PR #287, opened 2016-10-19
Updating requirements for in-a-box

stjohnjohnson:
Just adding OS X and Docker as required


tkyi:
Failed at "duplicate" step
Will be fixed in https://github.com/screwdriver-cd/screwdriver/pull/280

PR #288, opened 2016-10-19
Pull all logs that are available at a given time

stjohnjohnson:
With the new paging, log loading is a little slower from the UI as it polls every second.  This would load ALL the logs that are available after line X for each poll.

### DIFF
--- a/plugins/builds/steps/logs.js
+++ b/plugins/builds/steps/logs.js
@@ -6,6 +6,44 @@ const request = require('request');
 const ndjson = require('ndjson');
 const MAX_LINES = 100;
 
+/**
+ * Load all pages that are available
+ * @method loadAllLines
+ * @param  {String}     baseUrl   URL to load from (without the .$PAGE)
+ * @param  {Integer}    linesFrom What line number are we starting from
+ * @return {Promise}              Array of log lines
+ */
+function loadAllLines(baseUrl, linesFrom) {
+    return new Promise((resolve) => {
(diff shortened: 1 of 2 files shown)


plugins/builds/steps/logs.js
```diff
@@ -6,6 +6,44 @@ const request = require('request');
...
+ * @param  {String}     baseUrl   URL to load from (without the .$PAGE)
+ * @param  {Integer}    linesFrom What line number are we starting from
+ * @return {Promise}              Array of log lines
+ */
+function loadAllLines(baseUrl, linesFrom) {
+    return new Promise((resolve) => {
+        const page = Math.floor(linesFrom / MAX_LINES);
+        const output = [];
+
+        request
```
d2lam: do you need `return` here?

d2lam: nvm i don't think so

plugins/builds/steps/logs.js
```diff
@@ -6,6 +6,44 @@ const request = require('request');
...
+                if (line.n >= linesFrom) {
+                    output.push(line);
+                }
+            })
+            .on('end', () => resolve(output));
+    }).then((lines) => {
+        const linesCount = lines.length;
+
+        // Load from next log if we got stuff AND we reached the edge of a page
+        if (linesCount > 0 && (linesCount + linesFrom) % MAX_LINES === 0) {
```
FenrirUnbound: Would there ever be a case where `linesCount + linesFrom` would exceed `MAX_LINES`?

PR #291, opened 2016-10-20
Upgrading a couple packages

tkyi:
Upgrading a couple packages:
 - [github@5](https://github.com/mikedeboer/node-github/blob/master/CHANGELOG.md#500)
 - [chai-as-promised@6](https://github.com/domenic/chai-as-promised/releases/tag/v6.0.0)
 - [nock@9](https://github.com/node-nock/nock/blob/2a1bc2d68cb5e36bc4a1501b5cbe9fe3cbcd576f/CHANGELOG.md#900--2016-10-23)
 - mockery@2

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "pretest": "eslint .",
     "test": "jenkins-mocha --recursive",
     "start": "./bin/server",
-    "functional": "cucumber-js --format=pretty ./features/server.feature ./features/secrets.feature ./features/git-flow.feature ./features/authorization.feature"
+    "functional": "cucumber-js --format=pretty"
   },
   "bin": {
     "screwdriver-api": "./bin/server"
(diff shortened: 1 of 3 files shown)


stjohnjohnson:
So meta

package.json
```diff
@@ -91,17 +91,17 @@
   },
   "devDependencies": {
     "chai": "~3.5.0",
-    "chai-as-promised": "^5.3.0",
+    "chai-as-promised": "^6.0.0",
```
petey: Is chai-as-promised actually used? There is one file that requires it, but I don't think the features of chai-as-promised are even used in it.

tkyi: Looks like we use chai-as-promised in one test for its `isRejected` function here: https://github.com/screwdriver-cd/screwdriver/blob/master/test/plugins/auth.test.js#L108

petey: fair enough.

d2lam:
Let's not merge anything into the pipeline until it's fixed.

PR #305, opened 2016-10-24
Switching to pluggable auth

stjohnjohnson:
Blocked on https://github.com/screwdriver-cd/scm-github/pull/25

- Can configure to use GitHub Enterprise
- Can configure to use another SCM other than GitHub
- Split password up into its two forms (encryption and cookie signing)

_note I tried to keep the environment variables backwards compatible, but `local.yaml` is no longer the same_

### DIFF
--- a/config/custom-environment-variables.yaml
+++ b/config/custom-environment-variables.yaml
@@ -10,20 +10,18 @@ auth:
     # Generate one by running
     # $ openssl rsa -in jwt.pem -pubout -out jwt.pub
     jwtPublicKey: SECRET_JWT_PUBLIC_KEY
-    # The client id used for OAuth with github. Look up GitHub OAuth for details
-    # https://developer.github.com/v3/oauth/
-    oauthClientId: SECRET_OAUTH_CLIENT_ID
-    # The client secret used for OAuth with github
-    oauthClientSecret: SECRET_OAUTH_CLIENT_SECRET
     # The access token to use on behalf of a user to access the API. Used as
     # an alternative to the OAuth flow
     temporaryAccessKey: SECRET_ACCESS_KEY
     # The user name associated with the temproary access token. Used as a
     # means of functionally testing the API
     temporaryAccessUser: SECRET_ACCESS_USER
-    # A password used for encrypting session, and OAuth data.
+    # A password used for encrypting session data.
     # **Needs to be minimum 32 characters**
-    password: SECRET_PASSWORD
+    cookiePassword: SECRET_COOKIE_PASSWORD
+    # A password used for encrypting stored secrets.
+    # **Needs to be minimum 32 characters**
+    encryptionPassword: SECRET_PASSWORD
     # A flag to set if the server is running over https.
     # Used as a flag for the OAuth flow
     https: IS_HTTPS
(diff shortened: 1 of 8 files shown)


plugins/auth/login.js
```diff
@@ -48,12 +48,9 @@ module.exports = config => ({
                     if (!model) {
                         return factory.create({
                             username,
-                            token: githubToken,
-                            password: config.password
```
stjohnjohnson: _note: password hasn't been required here since we setup password on the UserFactory in `bin/server.js`_

plugins/auth/login.js
```diff
@@ -48,12 +48,9 @@ module.exports = config => ({
...
+                            token: githubToken
```
d2lam: we should rename this to accessToken since it's not always github

stjohnjohnson: Done

config/default.yaml
```diff
@@ -14,14 +14,12 @@ auth:
...
-    # The client secret used for OAuth with github
-    oauthClientSecret: AGAIN-SOMETHING-HERE-IS-USEFUL
-    # A password used for encrypting session, and OAuth data.
+    # A password used for encrypting session data.
     # **Needs to be minimum 32 characters**
-    password: WOW-ANOTHER-INSECURE-PASSWORD!!!
+    cookiePassword: WOW-ANOTHER-INSECURE-PASSWORD!!!
+    # A password used for encrypting stored secrets.
+    # **Needs to be minimum 32 characters**
+    encryptionPassword: WOW-ANOTHER-MORE-INSECURE-PASSWORD!!!
```
FenrirUnbound: > Another More

kgisi63

package.json
```diff
@@ -83,7 +83,7 @@
     "screwdriver-executor-k8s": "^9.0.0",
     "screwdriver-executor-docker": "^1.1.2",
     "screwdriver-models": "^16.0.0",
-    "screwdriver-scm-github": "^2.0.0",
+    "screwdriver-scm-github": "^3.0.0",
```
FenrirUnbound: Nevermind, I found the PR buried in the stack 👍

config/custom-environment-variables.yaml
```diff
@@ -10,20 +10,18 @@ auth:
...
     # an alternative to the OAuth flow
     temporaryAccessKey: SECRET_ACCESS_KEY
     # The user name associated with the temproary access token. Used as a
     # means of functionally testing the API
     temporaryAccessUser: SECRET_ACCESS_USER
-    # A password used for encrypting session, and OAuth data.
+    # A password used for encrypting session data.
     # **Needs to be minimum 32 characters**
-    password: SECRET_PASSWORD
+    cookiePassword: SECRET_COOKIE_PASSWORD
```
minz1027: Have we set it up yet?
Do we need to update this? https://github.com/screwdriver-cd/kubernetes/blob/master/api.yaml

stjohnjohnson:
Ugh.
```
  1) github plugin test "before each" hook for "returns 204":
     Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.
      at server.register (test/plugins/webhooks/github.test.js:109:13)
      at process.nextTick (node_modules/hapi/node_modules/hoek/lib/index.js:854:22)
      at _combinedTickCallback (internal/process/next_tick.js:67:7)
      at process._tickDomainCallback (internal/process/next_tick.js:122:9)
```

d2lam:
@stjohnjohnson  Ughhh I tried looking into it but don't see anything wrong. And it works fine locally every time. Is it possible that we register 2 plugins in the github test, so it takes a bit longer?

FenrirUnbound:
@d2lam Even the `screwdriver-store` hit timeouts with the server starting up. And that's a lightweight API in comparison to this one. The timeout issues with both of them may be related.

d2lam:
Bumping the timeout to 3000 for now.

PR #307, opened 2016-10-25
Forgot required flag about HTTPS

stjohnjohnson:
Without this, OAuth flow fails due to trying to redirect to HTTP not HTTPS


d2lam:
Don't we set it to false by default?

d2lam:
Oh i see

d2lam:
Ughhh that github before hook timeout error again ~.~

PR #328, opened 2016-10-28
Fix eslint rules

minz1027:
- Fixed to adapt to new eslint rules

### DIFF
--- a/test/plugins/pipelines.test.js
+++ b/test/plugins/pipelines.test.js
@@ -162,10 +162,7 @@ describe('pipeline plugin test', () => {
             options: {
                 password
             }
-        }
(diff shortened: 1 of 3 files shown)


test/plugins/pipelines.test.js
```diff
@@ -162,8 +162,7 @@ describe('pipeline plugin test', () => {
             options: {
                 password
             }
-        }
-    ], (err) => {
+        }], (err) => {
```
petey: This should be equivalent:

```
}], done);
```

minz1027: Nice! Fixed

PR #336, opened 2016-11-01
pr jobs can be started by non-screwdriver users

### LINKED ISSUE #335: PR builds created by non-user do not run
PR builds are not being created if the user that creates the PR is not in our database:

It is checking if user exists here: https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/webhooks/index.js#L187-L193

And user is from `parsed.username`, which is `pull_request.user.login` from the payload: https://github.com/screwdriver-cd/scm-github/blob/master/index.js#L448

```
{
  "action": "reopened",
  "number": 327,
  "pull_request": {
    "url": "https://api.github.com/repos/screwdriver-cd/screwdriver/pulls/327",
    "id": 91439458,
    "html_url": "https://github.com/screwdriver-cd/screwdriver/pull/327",
    "diff_url": "https://github.com/screwdriver-cd/screwdriver/pull/327.diff",
    "patch_url": "https://github.com/screwdriver-cd/screwdriver/pull/327.patch",
    "issue_url": "https://api.github.com/repos/screwdriver-cd/screwdriver/issues/327",
    "number": 327,
    "state": "open",
    "locked": false,
    "title": "Update github to version 6.0.1 🚀",
    "user": {
      "login": "greenkeeperio-bot",
```

FenrirUnbound:
The Github API has request-threshold quotas in place. Requests are tracked by either IP or API token. During the PR job creation, Screwdriver uses the PR users's access token to communicate with the Github API. If a user is not registered in Screwdriver, this operation will fail and the job will not run.

This PR proposes a solution that fixes #335 . If a PR user does not have an API token registered in Screwdriver, then it will use a default, generic user's API token instead.

### DIFF
--- a/test/plugins/webhooks/github.test.js
+++ b/test/plugins/webhooks/github.test.js
@@ -253,6 +253,16 @@ describe('github plugin test', () => {
                     assert.equal(reply.statusCode, 500);
                 });
             });
+
+            it('handles checkouting when given a non-listed user', () => {
+                userFactoryMock.get.resolves(null);
+                userFactoryMock.get.withArgs({ username: 'sd-buildbot' }).resolves(userMock);
+
+                return server.inject(options)
+                .then((response) => {
+                    assert.equal(response.statusCode, 201);
+                });
+            });
         });
 
         describe('pull-request event', () => {
@@ -386,6 +396,16 @@ describe('github plugin test', () => {
                             sha,
                             username
                         });
+                    });
+                });
+
+                it('handles checkout when given a non-listed user', () => {
+                    userFactoryMock.get.resolves(null);
+                    userFactoryMock.get.withArgs({ username: 'sd-buildbot' }).resolves(userMock);
+
+                    return server.inject(options)
+                    .then((response) => {
+                        assert.equal(response.statusCode, 201);
                     });
                 });
             });
--- a/plugins/webhooks/index.js
+++ b/plugins/webhooks/index.js
@@ -2,6 +2,8 @@
 
 const boom = require('boom');
 
+const GENERIC_SCM_USER = 'sd-buildbot';
+
 /**
  * Stop a job by stopping all the builds associated with it
  * If the build is running, set state to ABORTED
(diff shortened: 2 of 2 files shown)


test/plugins/webhooks/github.test.js
```diff
@@ -253,6 +253,16 @@ describe('github plugin test', () => {
...
                 });
             });
+
+            it('handles checkouting when given a non-listed user', () => {
+                userFactoryMock.get.resolves(null);
+                userFactoryMock.get.withArgs({ username: 'sd-buildbot' }).resolves(userMock);
+
+                return server.inject(options)
+                .then((response) => {
+                    assert.equal(response.statusCode, 201);
```
d2lam: We should probably assert `userFactoryMock.get` is called with `sd-buildbot`

FenrirUnbound: You can't get to this point if `userFactoryMock.get` was not called with `sd-buildbot`

FenrirUnbound: The stub is guarded in a way that you cannot get a `userMock` if you don't provide `sd-buildbot` as a user: https://github.com/screwdriver-cd/screwdriver/pull/336/files#diff-844b8ba64bd36a7e8d37f9f4d803295dR259

d2lam: I agree, but it's always better to assert. If somehow the code was okay with `null` user. It's better to have this check. :)

FenrirUnbound: This test would fail if the code was passed with `null` user

d2lam: I mean with that argument we can remove all the `calledWith` in our tests where `statusCode` is already checked. 
One example is here (and throughout this file): https://github.com/screwdriver-cd/screwdriver/blob/master/test/plugins/builds.test.js#L336-L337

I understand that technically it won't pass the test if that function wasn't called. But since the code says it calls it, we should assert that it's actually called. That way if **somehow** it returns `201` without calling that function, it will fail here. The error message  will be more accurate as well.

FenrirUnbound: You still need the `calledWith` in the other tests. The way we always have setup the stubs is like this:

``` js
   // https://github.com/screwdriver-cd/screwdriver/pull/336/files#diff-844b8ba64bd36a7e8d37f9f4d803295dR170

   userFactoryMock.get.resolves(userMock);
```

This is saying that, given any kind of input, always return `userMock`. This means that, no matter what the input is, you will always get `userMock` in return. This is why we use `assert.calledWith` in our tests: to make sure that the stub was invoked with the correct parameters because we configured the stub to not make that assertion.

Now the way that this test is setup is this:

```js
   // https://github.com/screwdriver-cd/screwdriver/pull/336/files#diff-844b8ba64bd36a7e8d37f9f4d803295dR258

  userFactoryMock.get.resolves(null);
  userFactoryMock.get.withArgs({ username: 'sd-buildbot' }).resolves(userMock);
```

This is saying that, given any kind of input, always return `null`. But, if and only if we get `{username: 'sd-buildbot}`, we return the `userMock`. Using an `assert.calledWith` is redundant here because you could never get `userMock` if the parameters were incorrect.

d2lam: Ok. I'm fine with it.

PR #337, opened 2016-11-01
Create an event when creating a build via api

minz1027:
- Create an event when creating a build via build create API endpoint
- Will add creating events for webhook events in another PR
- Need to DELETE all the builds before this gets merged

Related to issue #221


plugins/builds/create.js
```diff
@@ -49,7 +51,20 @@ module.exports = () => ({
                         })
                         // user has good permissions, sync and create a build
                         .then(() => pipeline.sync())
-                        .then(() => buildFactory.create(payload))
+                        .then(() => user.unsealToken())
```
d2lam: These can probably be in an outer scope.

Issue #347, opened 2016-11-03
Environment override for K8S_HOST is broken


jer:
If you set K8S_HOST, nothing gets overridden. @d2lam found that the override happens on options.kubernetes.host, while 'kubernetes' there is not actually valid.

d2lam:
@jer  Does it work now?

jer:
Yep, just confirmed that the environment override is working. Thanks!

PR #349, opened 2016-11-03
Pass in executor config correctly

d2lam:
`executor-k8s` expects options to be: ``` { kubernetes: { host, token... } } ``` 
`executor-docker` expects options to be: ``` { docker: { host, port, ... } }```

But we've been passing them in without `kubernetes` or `docker` so it's been taking default values. This also explains why k8s never worked locally. This pass the options in as ``` { executor: { ... } } ```

Relates to https://github.com/screwdriver-cd/screwdriver/issues/347


d2lam:
Closing. Found an easier way to fix this without changing the executor plugins.

PR #350, opened 2016-11-03
Pass in kubernetes config correctly

d2lam:
They should be inside `kubernetes`: https://github.com/screwdriver-cd/executor-k8s/blob/master/index.js#L19-L22

Relates to https://github.com/screwdriver-cd/screwdriver/issues/347


jer:
👍

Issue #284, opened 2016-10-18
Create duplicate secrets


minz1027:
- In the secret create route, we don't check duplicates:
- With dynamo, it will just do an update on the existing record
- With sequelize, it will throw an validation error
- With sequelize, the secret functional tests will fail since we are creating same secrets again and again

PR #353, opened 2016-11-05
Prevent creating duplicate secrets

d2lam:
Resolves https://github.com/screwdriver-cd/screwdriver/issues/284

### DIFF
--- a/plugins/secrets/create.js
+++ b/plugins/secrets/create.js
@@ -45,7 +45,19 @@ module.exports = () => ({
                                 `User ${username} is not an admin of this repo`);
                         }
                     })
-                    .then(() => secretFactory.create(request.payload))
+                    // check if secret already exists
+                    .then(() => secretFactory.get({
+                        pipelineId: request.payload.pipelineId,
+                        name: request.payload.name
+                    }))
+                    // if secret already exists, reject
+                    .then((secret) => {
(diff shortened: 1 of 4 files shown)


test/plugins/secrets.test.js
```diff
@@ -175,6 +175,16 @@ describe('secret plugin test', () => {
             });
         });
 
+        it('returns 409 when the pipieline already exists', () => {
```
tkyi: returns 409 when the `secret` already exists

plugins/secrets/create.js
```diff
@@ -45,7 +45,19 @@ module.exports = () => ({
...
-                    .then(() => secretFactory.create(request.payload))
+                    // check if secret already exists
+                    .then(() => secretFactory.get({
+                        pipelineId: request.payload.pipelineId,
+                        name: request.payload.name
+                    }))
+                    // if secret already exists, reject
+                    .then((secret) => {
+                        if (secret) {
+                            throw boom.conflict(`Secret already exists: ${secret.id}`);
```
minz1027: To make it more clear, can we put something like `Secret already exists with the id: ${secret.id}` ?

d2lam: Sure. Added.

PR #355, opened 2016-11-08
Add GET event/:id/builds endpoint

minz1027:
- Add `GET event/:id/builds` and `GET event/:id` endpoints
- So that the UI can query all builds associated with a event

### DIFF
--- a/plugins/events/get.js
+++ b/plugins/events/get.js
@@ -3,23 +3,23 @@
 const boom = require('boom');
 const joi = require('joi');
 const schema = require('screwdriver-data-schema');
-const getSchema = schema.models.build.get;
-const idSchema = joi.reach(schema.models.build.base, 'id');
+const getSchema = schema.models.event.get;
+const idSchema = joi.reach(schema.models.event.base, 'id');
 
 module.exports = () => ({
     method: 'GET',
-    path: '/builds/{id}',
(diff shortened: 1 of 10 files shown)


plugins/events/get.js
```diff
@@ -0,0 +1,38 @@
...
+    method: 'GET',
+    path: '/events/{id}',
+    config: {
+        description: 'Get a single event',
+        notes: 'Returns a event record',
+        tags: ['api', 'events'],
+        handler: (request, reply) => {
+            const factory = request.server.app.eventFactory;
+
+            factory.get(request.params.id)
```
d2lam: missing `return` here

minz1027: fixed

petey: you fixed the one in builds, but not here.

test/plugins/jobs.test.js
```diff
@@ -138,7 +138,7 @@ describe('job plugin test', () => {
         });
     });
 
-    describe('/jobs/{id}', () => {
+    describe('PUT /jobs/{id}', () => {
```
tkyi: nice! 👍

plugins/events/get.js
minz1027: LOL, finally fixed!

Issue #312, opened 2016-10-25
Scm modules output the commands to checkout


d2lam:
Currently, the launcher does `git fetch origin prRef` to merge the PR branch to master: https://github.com/screwdriver-cd/launcher/blob/ca5d3e7cb5eaf20e373b236dc636c86bd866e5c6/git/git.go#L105-L107

Scm modules should output the shell command to check out, which will be stored in the job. For example:

```
{ 
  name: "sd-setup",
  command: "git checkout git@github.com:screwdriver-cd/screwdriver.git"
}
```

Until this is done, PR builds for bitbucket won't work. 

---
## Work
- [x] Data schema
- [x] Scm-base
- [x] Scm-github
- [x] Scm-bitbucket
- [x] Models
- [x] Launcher
- [x] API
- [x] Integrate new models into API
- [x] Deploy new launcher

d2lam:
Done!

<img width="1425" alt="screen shot 2016-11-17 at 1 15 14 pm" src="https://cloud.githubusercontent.com/assets/3401924/20407847/eae1df74-acc7-11e6-8a1c-93b164323909.png">

PR #356, opened 2016-11-10
Pass in prRef when creating a PR build

tkyi:
The `prRef` will be needed in models when calling the new `getCheckoutCommand` method defined in scm plugins

Blocked by https://github.com/screwdriver-cd/scm-github/pull/39
Blocked by https://github.com/screwdriver-cd/scm-bitbucket/pull/24
Related to https://github.com/screwdriver-cd/screwdriver/issues/312


plugins/webhooks/index.js
```diff
@@ -83,7 +83,7 @@ function startPRJob(options, request) {
             });
         })
         // create a build
-        .then(event => buildFactory.create({ jobId, sha, username, eventId: event.id }))
+        .then(event => buildFactory.create({ jobId, sha, username, eventId: event.id, prRef }))
```
d2lam: I don't think this would work. `prRef` is not an attribute of `build`. It won't get saved.

tkyi: We're not going to save `prRef` to the table, we just want to use it in models to call `getCheckoutCommand`

d2lam: OK nvm 👍

PR #374, opened 2016-11-29
Allow users to customize TLS settings via a JSON stringified environment variable `HTTPD_TLS`

stjohnjohnson:
So you can do something like:
```
HTTPD_TLS='{"key":"`cat ./cert.key`","cert":"`cat ./cert.cert`"}'
```

More options are here: https://nodejs.org/api/tls.html#tls_tls_createserver_options_secureconnectionlistener


outofjungle:
👍

Issue #400, opened 2016-12-20
Mass create Screwdriver pipelines


tkyi:
Since we are frequently moving/taking down databases and recreating them, it would be nice to have a script/tool to easily regenerate Screwdriver pipelines with secrets

- [x] semantic release for all node  repos
- [x] script for mass create all pipelines
  - [x] node repos with 2 secrets: NPM_TOKEN, GH_TOKEN
  - [x] node repos(screwdriver, UI) with more secrets
  - [x] go repos
  - [x] no-secret repos
- [x] GIT_KEY for all repos that need it
- [x] Test out script using beta tables
- [x] Create pipelines in production postgres tables

List of all the repos:
https://gist.github.com/tkyi/e5870adcd13163d16d236a4cb387a3b4

Script:
https://gist.github.com/minz1027/81f9c43a87fab3e1878166d5f3d36ebf

Problems:
1. Hard to automate the creation of `GIT_KEY` secret. Can we get rid of `GIT_KEY` and use `GITHUB_TOKEN` instead?

stjohnjohnson:
Problem #1, yes please.  Who still uses GIT_KEY?

minz1027:
@stjohnjohnson `ui`, `homepage`, `guide` and the three GO repos: `gitversion`, `launcher`, `log-service`.

tkyi:
**Status:** Waiting on #399 before trying to recreate all the pipelines.

tkyi:
Pipelines have been recreated successfully.

- [ ] [optional] move mass-create pipelines script to central tools repo

PR #402, opened 2016-12-22
Change secret GIT_TOKEN to GITHUB_TOKEN to be consistent other repos

tkyi:
Trying to make the secret key match the UI's (https://github.com/screwdriver-cd/ui/blob/master/screwdriver.yaml#L34) so the mass-create script will be easier to write

Related to https://github.com/screwdriver-cd/screwdriver/issues/400

### DIFF
--- a/features/support/before-hook.js
+++ b/features/support/before-hook.js
@@ -37,7 +37,7 @@ function beforeHooks() {
     this.Before((scenario, cb) => {
         const host = process.env.SD_API;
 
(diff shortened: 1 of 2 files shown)


petey:
Looks like you are going to have to pin some of the module version to prevent pulling in newer versions of data-schema.

tkyi:
Eh it was more for consistency but I'm okay with closing this for now. It's not really an urgent thing

PR #401, opened 2016-12-22
feat(399): update API to use numeric ids instead of hashes

petey:
- [x] Update dependencies
- [x] Fix tests to use numeric ids
- [x] Update K8s config to use new PostgreSQL db

Epic: https://github.com/screwdriver-cd/screwdriver/issues/399

----

BREAKING CHANGES: 
 - Removes all usage of hash based ids in favor of numeric ids
 - Removes DynamoDB support

Issue #379, opened 2016-12-01
Repository webhooks


FenrirUnbound:
## Scenario

When creating a Screwdriver pipeline for my repository, a webhook is not added to the target repository. This means that when the target branch is updated either by directly pushing to it or merging in a Pull-Request, a new build has to be manually initiated. 

Adding a webhook to the repository solves this, but is also a manual task.

Automating this manual work would ease the on-boarding process.

## Tasks

- [x] Update `scm-base` to expose an endpoint for adding a webhook
   - [x] Implement the functionality in `scm-github`
      - [ ] Clean up/Refactor to not test against pseudo-private functions
   - [x] Implement the functionality in `scm-bitbucket`
      - [x] determine permissions scope
      - [ ] Clean up/Refactor to not test against pseudo-private functions
- [x] Integrate the functionality in the `Pipeline`/`PipelineFactory` model.
   - [x] Decide on whether it happens during `create` or `sync`
   - [x] Clean up/Refactor the `.sync` operation

## Related Issues

- screwdriver-cd/screwdriver#219

### 2016-12-6
* Additional item for `scm-bitbucket`. There was some work with exploring which permissions were needed to modify webhooks for a repository.

### 2016-12-12
* As part of PR feedback, we included tasks for refactoring and cleaning up.

FenrirUnbound:
PoC

```javascript
'use strict';

const Github = require('github');
const client = new Github({
  debug: false,
  protocol: 'https',
  host: 'api.github.com',
  headers: {
    'user-agent': 'FenrirUnbound-client'
  }
});

client.authenticate({
  type: 'oauth',
  token: process.env.AUTH_TOKEN || ''
});

const events = ['pull_request', 'push'];
const config = {
  url: 'https://api.screwdriver.cd/v4/webhooks',
  content_type: 'json',
  secret: 'someSecretGoesHere'
};
client.repos.createHook({
  owner: 'FenrirUnbound',
  repo: 'wallball',
  name: 'web',  // defined by list - https://api.github.com/hooks
  config,
  events,
  active: true
}).then(console.log);

```

FenrirUnbound:
Instead of guessing what the API is supposed to be in `scm-base`, we're implementing it in both `scm-github` and `scm-bitbucket` first. Once we have a baseline for what inputs each of these implementations require, we'll refactor it up into the `scm-base`.

FenrirUnbound:
PoC for bitbucket.org usage.

[...]

PR #403, opened 2016-12-22
feat(379) Add webhook during pipeline recreation

d2lam:
- Since we [moved addWebhook out of sync](https://github.com/screwdriver-cd/models/pull/146), we should now call it inside the `POST /pipelines`. 
- Also added a new endpoint to sync webhooks if it's not correct. We can add a button for it in UI later.

Blocked by: https://github.com/screwdriver-cd/screwdriver/pull/401
Related: https://github.com/screwdriver-cd/screwdriver/issues/379

### DIFF
--- a/plugins/pipelines/syncWebhooks.js
+++ b/plugins/pipelines/syncWebhooks.js
@@ -7,10 +7,10 @@ const idSchema = joi.reach(schema.models.pipeline.base, 'id');
 
 module.exports = () => ({
     method: 'GET',
-    path: '/pipelines/{id}/sync',
+    path: '/pipelines/{id}/sync/webhooks',
     config: {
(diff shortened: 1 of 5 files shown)


plugins/pipelines/syncWebhooks.js
```diff
@@ -0,0 +1,58 @@
...
+                }
+                if (!user) {
+                    throw boom.notFound(`User ${username} does not exist`);
+                }
+
+                // ask the user for permissions on this repo
+                return user.getPermissions(pipeline.scmUri)
+                    // check if user has read access
+                    .then((permissions) => {
+                        if (!permissions.pull) {
```
minz1027: push permission will be better?

Issue #399, opened 2016-12-19
Use Relational Database instead of NoSQL


stjohnjohnson:
Right now we're using DynamoDB to store our data.  It worked well for us at first, but we've encountered two things that make it hard to move forward:
 - Our IDs are hashes, which are hard to remember.  _"Oh yeah, project 1fec866296c698141c2e8e0ed21da0ba23c848f1"_  Implementing auto-incrementing IDs in NoSQL is difficult.
 - Pagination doesn't work as one would expect (no way to do page X, limit Y).

Instead, we're going to move our system to use relational databases (like PostgreSQL) and switch to numeric IDs.

---

- [x] Update Data Schema
- [x] Update Models
- [x] Update Sequelize
- [x] Remove DynamoDB
- [x] Fix all the API stuff
- [x] Delete Everything
- [x] Recreate all the things

stjohnjohnson:
EOD Summary:
 - Made changes to Data Schema to use new ID
 - Found out that keeping Hash was complicated and didn't provide any value
 - Started upgrade on Datastore Base and Sequelize
 - Working on change to model to support get by ID OR unique keys

stjohnjohnson:
Right now we're just waiting on the API rollout: https://github.com/screwdriver-cd/screwdriver/pull/401

tkyi:
Disabled prod job for API and pointed beta job to new postgres DB. Merged #401 

Currently debugging acceptance test failure to create pipelines.

tkyi:
Found a bunch of bugs:
- [ ] Need to change datastore-sequelize to use `upsert` instead of `create`. Dynamodb allowed for overwriting of table entries, postgres does not. (Seeing a SequelizeUniqueConstraintError)
- [ ] Need to update sdstore docker image. Seeing validation errors `\"Build ID\" fails because [\"Build ID\" length must be 40 characters long]"`
- [ ] Need to investigate error. Seeing `Error: Failed to create job: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"jobs.batch \"10\" already exists","reason":"AlreadyExists","details":{"name":"10","group":"batch","kind":"jobs"},"code":409}`

stjohnjohnson:
@d2lam for the upsert, why are we creating the same thing twice?  Shouldn't it be updates? https://github.com/screwdriver-cd/datastore-sequelize/blob/master/index.js#L281

[...]

PR #408, opened 2017-01-03
fix(399): update job for PR sync

d2lam:
- For PR sync event, the job already exists, so we do an `update` instead of a `create`
- Revert https://github.com/screwdriver-cd/screwdriver/pull/407/commits/aa33dff74ee523b214d2fb734d41a8b01f2086b5 since it was meant to archive the old PR job, and create another PR job. But that doesn't really make sense.

Related: https://github.com/screwdriver-cd/screwdriver/issues/399

### DIFF
--- a/plugins/webhooks/index.js
+++ b/plugins/webhooks/index.js
@@ -1,7 +1,6 @@
 'use strict';
 
 const boom = require('boom');
-
 const GENERIC_SCM_USER = 'sd-buildbot';
 
 /**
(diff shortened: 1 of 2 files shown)


plugins/webhooks/index.js
```diff
@@ -65,9 +59,24 @@ function startPRJob(options, request) {
...
-        .then(permutations => jobFactory.create({ pipelineId, name, permutations }))
+        .then(([permutations, jobs]) => {
+            const jobNamesArray = jobs.map(j => j.name);
+            const jobIndex = jobNamesArray.indexOf(name);
+
+            // if PR job already exists, update
+            if (jobIndex > -1) {
+                const job = jobs[jobIndex];
+
+                hoek.merge(job, { pipelineId, name, permutations });
```
minz1027: Probably just need the `permutations`.

PR #414, opened 2017-01-05
fix: set archived to true if PR is reopened

minz1027:
- set archived to true if PR is reopened after closed


d2lam:
Add test please :)

Issue #417, opened 2017-01-10
Syncing PRs


d2lam:
Currently, if webhook fails to send, PR jobs will be at a wrong state and there is no way to correct this since our current `sync` does not take care of PRs. 

- [x]  Add `getOpenedPRs()` to scm-base  --- https://github.com/screwdriver-cd/scm-base/pull/37
- [x]  Add `getOpenedPRs()` to scm-github: [example endpoint](https://api.github.com/repos/screwdriver-cd/models/pulls)
- [x]  Add `getOpenedPRs()` to scm-bitbucket: [example endpoint](https://api.bitbucket.org/2.0/repositories/d2lam2/%7Bde7d7695-1196-46a1-b87d-371b7b2945ab%7D/pullrequests?state=OPEN)
- [x]  Add `syncPRs()` to models      --- https://github.com/screwdriver-cd/models/pull/148
- [x]  Add `sync/pr` endpoint to API --- https://github.com/screwdriver-cd/screwdriver/pull/418
- [x]  Add a buttons to sync in UI
   - [x] Sync pipeline, PR,  webhooks
   - [x] Add tests

Related: https://github.com/screwdriver-cd/screwdriver/issues/389

d2lam:
**Update as of 1/17:** 
The backend work is done. Sync PRs can be accessed through: `/v4/sync/pullrequests`
We are working on adding a button to the UI for the users to sync their pipeline.

tkyi:
The UI pipeline is broken so the changes cannot be pushed out and verified. Looking into that today.

d2lam:
This feature is done.

PR #418, opened 2017-01-10
feat: Add an endpoint to sync PRs

d2lam:
- Add an endpoint to sync PR jobs
- This will call `syncPRs()` from the model

Blocked by: https://github.com/screwdriver-cd/models/pull/148
Related: https://github.com/screwdriver-cd/screwdriver/issues/417


d2lam:
Tested this locally with sd-in-a-box. Worked!

plugins/pipelines/syncPRs.js
```diff
@@ -0,0 +1,54 @@
+'use strict';
+
+const boom = require('boom');
+const joi = require('joi');
+const schema = require('screwdriver-data-schema');
+const idSchema = joi.reach(schema.models.pipeline.base, 'id');
+
+module.exports = () => ({
+    method: 'GET',
```
FenrirUnbound: This question is more about semantics than the logic. Is a `GET` operation appropriate for syncing the pull requests?

d2lam:
@minz1027  Added to README

plugins/pipelines/syncPRs.js
d2lam: I think it could be either a GET or a PUT. But since it's been `GET` for `sync` and `syncWebhooks`, I keep it the same to be consistent.

PR #419, opened 2017-01-10
fix: Whitelist Jerry

tkyi:
Need to add our new intern to the whitelist

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -37,6 +37,7 @@ auth:
         - d2lam
         - dvdizon
         - jer
+        - jerryzhang222
         - minz1027
         - nicolaifsf
         - nkatzman


config/default.yaml
```diff
@@ -37,6 +37,7 @@ auth:
         - d2lam
         - dvdizon
         - jer
+        - jerryzhang
```
d2lam: Isn't it `jerryzhang222`?

tkyi: oops

config/default.yaml
```diff
@@ -37,6 +37,7 @@ auth:
         - d2lam
         - dvdizon
         - jer
+        - jerryzhang222
```
FenrirUnbound: https://github.com/jerryzhang222

Looks legit 👍

PR #428, opened 2017-01-18
docs: refer to the correct docker image

FenrirUnbound:
The `README.md` refers to the wrong docker image. It's referring to `screwdrivercd/api` when it should be `screwdrivercd/screwdriver`.

Reference: https://news.ycombinator.com/item?id=13424592
Docker Image: https://hub.docker.com/r/screwdrivercd/screwdriver/

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ info: Server running at http://localhost:8080
 
 ```bash
 $ vim ./local.yaml # See below for configuration
-$ docker run --rm -it --volume=`pwd`/local.yaml:/config/local.yaml -p 8080 screwdrivercd/api:latest
+$ docker run --rm -it --volume=`pwd`/local.yaml:/config/local.yaml -p 8080 screwdrivercd/screwdriver:stable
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,7 +23,7 @@ info: Server running at http://localhost:8080
 
 ```bash
 $ vim ./local.yaml # See below for configuration
-$ docker run --rm -it --volume=`pwd`/local.yaml:/config/local.yaml -p 8080 screwdrivercd/api:latest
+$ docker run --rm -it --volume=`pwd`/local.yaml:/config/local.yaml -p 8080 screwdrivercd/screwdriver:latest
```
tkyi: Should it be `stable` version?

d2lam: probably stable.

FenrirUnbound: Amended.

Last time we updated the `stable` tags, it was manually done. We should re-visit our policy on how we treat how we tag docker images, especially if that's going to be a way users interact with the system.

d2lam: 👍

PR #437, opened 2017-01-26
feat: Made setup and teardown plugins configurable

jerryzhang222:
Small changes to allow users to set setup and teardown plugins as environment variables.

### DIFF
--- a/config/custom-environment-variables.yaml
+++ b/config/custom-environment-variables.yaml
@@ -112,3 +112,7 @@ ecosystem:
     store: ECOSYSTEM_STORE
     # Badge service (needs to add a status and color)
     badges: ECOSYSTEM_BADGES
+    # List of module names, or objects { name, config } for instantiation to use in sd-setup
+    setup: SD_SETUP_PLUGINS
+    # List of module names, or objects { name, config } for instantiation to use in sd-teardown
+    teardown: SD_TEARDOWN_PLUGINS


config/custom-environment-variables.yaml
```diff
@@ -112,3 +112,7 @@ ecosystem:
     store: ECOSYSTEM_STORE
     # Badge service (needs to add a status and color)
     badges: ECOSYSTEM_BADGES
+    # List of module names, or objects { name, config } for instantiation to use in sd-setup
+    setup: SD_SETUP_PLUGINS
```
FenrirUnbound: Based on this, anything that's set to that environment variable will override the current `setup`. This means that the `sd-scm` will be overridden. 

I would have expected something closer to an append, rather than a complete overwrite.

d2lam: Actually sd-scm is the default config already (1st param that gets passed into Bookend constructor), so this will have no effects

d2lam: Actually sd-scm is the default config already (1st param that gets passed into Bookend constructor), so this will have no effects

FenrirUnbound: The first parameter is a map of all the default modules. The second parameter defines which plugins to actually use.

If you have the `screwdriver-scm-github` in the first parameter, it won't be loaded unless you specify it in the setup list.

Reference: https://github.com/screwdriver-cd/build-bookend/blob/master/index.js#L84-L86

d2lam: Oh I think you're right. But since `SD_SCM_SETUP` is not configured, it will use the value in `default.yaml`. So in this case, nothing will be affected.

d2lam: Oh actually I get what you mean. We should keep sd-scm as required always 👍

config/custom-environment-variables.yaml
```diff
@@ -112,3 +112,9 @@ ecosystem:
     store: ECOSYSTEM_STORE
     # Badge service (needs to add a status and color)
     badges: ECOSYSTEM_BADGES
+    # List of module names, or objects { name, config } for instantiation to use in sd-setup
+    setup:
+        - sd-scm
+        - SD_SETUP_PLUGINS
```
FenrirUnbound: If SD_SETUP_PLUGINS is set to "my_custom_plugin", this seems like it would work. Are we only allowing a single plugin to be added?

FenrirUnbound: Nevermind! We're going back!

d2lam:
@FenrirUnbound  please re-review :)

stjohnjohnson:
@FenrirUnbound @d2lam @jerryzhang222 this is incorrect, it's not going to work as you expect.

You need to use [this](https://github.com/lorenwest/node-config/wiki/Environment-Variables#custom-environment-variables) feature:
```
 setup:
  __name: BOOKENDS_SETUP
  __format: json
```

Also, the environment name is very different from other names.  Why can't we do something like bookend_setup?

d2lam:
@stjohnjohnson  Oh I see. Thanks! I thought you could just set the environment variable to an array. Will make a fix tomorrow. This should not break anything.

PR #442, opened 2017-01-27
fix: Kubernetes ConfigMap-compatible mounting

mgoodness:
Afraid #440 didn't fix my problem (mounting a ConfigMap that contains a `local.yaml` entry at `/config`). This does, though.

Also tested it with "native" Docker bind-mounting, i.e. `docker run -v $(pwd)/local.yaml:/config/local.yaml screwdrivercd/screwdriver ...`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ RUN npm install screwdriver-api@$VERSION
 WORKDIR /usr/src/app/node_modules/screwdriver-api
 
 # Setup configuration folder
-RUN ln -s /usr/src/app/node_modules/screwdriver-api/config /config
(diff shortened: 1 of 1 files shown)


yahoocla:
Thank you for submitting this pull request, however I do not see a valid CLA on file for you. Before we can merge this request please visit https://yahoocla.herokuapp.com/ and agree to the terms. Thanks! :smile:

stjohnjohnson:
Hmm, that should be the correct symlink direction.  I just tested it as well:
```
$ docker run --rm -it -v `pwd`/local.yaml:/config/local.yaml screwdrivercd/screwdriver:v0.4.83
```

Which version of the container did you test it on?  We haven't retagged 0.4.83 as stable.

petey:
Please reopen with comments if you are still having trouble.

PR #443, opened 2017-01-29
fix: Obtain IP for Docker-Machine users

bdangit:
This allows users who create docker-machines as virtual machines.  These machines have IPs that are created on virtualized networks on the host.

### DIFF
--- a/in-a-box.py
+++ b/in-a-box.py
@@ -1,7 +1,9 @@
 # coding=utf-8
+import os
 import socket
 import sys
 import distutils.spawn
+from urlparse import urlparse
 from string import Template
 from subprocess import check_output, call, STDOUT, PIPE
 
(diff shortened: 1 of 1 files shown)


stjohnjohnson:
So will this give you a routable IP for the Virtual Machine from the base host?

bdangit:
@stjohnjohnson, yes that is correct.  Virtualbox, for instance, will create virtualized networks and they all are interfaceable via these "virtualized" nics.  You can see these interfaces via `ifconfig`.

example:

```
$ ifconfig | grep -A 2 vmnet
vmnet1: flags=8863<UP,BROADCAST,SMART,RUNNING,SIMPLEX,MULTICAST> mtu 1500
        ether 00:50:56:c0:00:01
        inet 172.16.56.1 netmask 0xffffff00 broadcast 172.16.56.255
--
vmnet2: flags=8863<UP,BROADCAST,SMART,RUNNING,SIMPLEX,MULTICAST> mtu 1500
        ether 00:50:56:c0:00:02
        inet 192.168.32.1 netmask 0xffffff00 broadcast 192.168.32.255
--
vmnet8: flags=8863<UP,BROADCAST,SMART,RUNNING,SIMPLEX,MULTICAST> mtu 1500
        ether 00:50:56:c0:00:08
        inet 192.168.90.1 netmask 0xffffff00 broadcast 192.168.90.255
```

While the IPs above represent what the NIC has, your VMs may have an IP allocation that is within any of these networks above.  Regardless if you configure your virtualized networks in NAT or Host-only, the base host is able to route traffic to the IP that has been allocated to it.  You can think of the base host as the default gateway or router.

This image below is a good diagram of what is happening.

img
ref: http://1001notes.blogspot.com/2016/07/networking-in-virtualbox.html

Issue #436, opened 2017-01-25
Cannot use Screwdriver with a private repository


bitemyapp:
Currently get this error despite the application being authorized
![screenshot from 2017-01-25 16-06-05](https://cloud.githubusercontent.com/assets/320177/22311249/34adb19c-e318-11e6-9216-c19c3e542754.png)

d2lam:
Hey @bitemyapp, I just looked into this more. It should have worked even if it's a private repo as long as you have access to it. Looking into the server log,  the error came from here: https://github.com/screwdriver-cd/scm-github/blob/master/index.js#L432-L451

I think what happens is it couldn't find a `screwdriver.yaml` in your repo. Can you make sure that you have that file? If my theory is right, the error message here is confusing and inaccurate and we should definitely fix it.

bitemyapp:
I think the issue was that the `screwdriver.yaml` existed in a `screwdriver` branch I had PR'd, but not in master.

bitemyapp:
![screenshot from 2017-01-26 14-51-35](https://cloud.githubusercontent.com/assets/320177/22349799/f4b1d234-e3d6-11e6-881b-c8c52a1403da.png)
![screenshot from 2017-01-26 14-50-30](https://cloud.githubusercontent.com/assets/320177/22349800/f4c0854a-e3d6-11e6-8afd-9ec2be60569c.png)

Same result for OAuth application under my name (I have access to the repository) and for OAuth application under `betterteamapp`.

bitemyapp:
![screenshot from 2017-01-26 14-52-21](https://cloud.githubusercontent.com/assets/320177/22349835/107e6630-e3d7-11e6-916a-7a7013a72559.png)

in master.

stjohnjohnson:
Ahh, so the problem lies here: https://github.com/screwdriver-cd/scm-github/blob/master/index.js#L701

We're not asking for read access to private repos. We can solve this by adding `repo` to the list: https://developer.github.com/v3/oauth/#scopes

I wish GitHub allowed us to ask for read only access to private repositories, because this grants our token WAY too much power:

> Grants read/write access to code, commit statuses, invitations, collaborators, adding team memberships, and deployment statuses for public and private repositories and organizations.

d2lam:
So this seems like a github limitation. Travis-pro, which supports private repos would also need to ask for write access:
[...]

PR #448, opened 2017-02-03
[WIP] feat: Add read protection for pipelines

d2lam:
Some work for https://github.com/screwdriver-cd/screwdriver/issues/436
- When get a pipeline, if no access return `404`
- When list pipelines, only returns the ones that user has access to. 

Please give some feedback :)


stjohnjohnson:
I have a much larger idea for how this should work.  We should not be doing the check on each API call but instead use specific JWTs for this case.  I'm going to write up an issue to describe it.

d2lam:
@stjohnjohnson  Sure!

PR #452, opened 2017-02-08
fix: can settings username and email for checkout

wahapo:
Add environment variables `SCM_USERNAME` and `SCM_EMAIL` with use custom user for checkout.
https://github.com/screwdriver-cd/scm-github/pull/52
https://github.com/screwdriver-cd/scm-bitbucket/pull/34

As a further modification, if a user does not have a valid SCM token registered with Screwdriver, `SCM_USERNAME` will use this user's token instead.

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -116,6 +116,9 @@ scm:
         oauthClientSecret: AGAIN-SOMETHING-HERE-IS-USEFUL
         # You can also configure for use with GitHub enterprise
         #gheHost: github.screwdriver.cd
+        # The username and email used for checkout with github
+        #username: sd-buildbot
+        #email: dev-null@screwdriver.cd
         # Secret to add to GitHub webhooks so that we can validate them
         secret: SUPER-SECRET-SIGNING-THING
         # Whether it supports private repo: boolean value.
@@ -125,6 +128,13 @@ scm:
     bitbucket:
         oauthClientId: YOUR-BITBUCKET-OAUTH-CLIENT-ID
         oauthClientSecret: YOUR-BITBUCKET-OAUTH-CLIENT-SECRET
+        # The username and email used for checkout with bitbucket
+        #username: sd-buildbot
+        #email: dev-null@screwdriver.cd
+
+webhooks:
+    # Obtains the SCM token for a given user. If a user does not have a valid SCM token registered with Screwdriver, it will use this user's token instead.
+    #username: sd-buildbot
 
 bookends:
     # Plugins for build setup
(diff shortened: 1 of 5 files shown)


config/default.yaml
```diff
@@ -116,6 +116,9 @@ scm:
         oauthClientSecret: AGAIN-SOMETHING-HERE-IS-USEFUL
         # You can also configure for use with GitHub enterprise
         #gheHost: github.screwdriver.cd
+        # The username and email used for checkout with github
+        #username: sd-buildbot
```
stjohnjohnson: I wouldn't comment these out, let them be there

plugins/webhooks/index.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
 const boom = require('boom');
+const joi = require('joi');
 const GENERIC_SCM_USER = 'sd-buildbot';
```
stjohnjohnson: I would remove this, as you can default it in the default.yaml

config/default.yaml
```diff
@@ -125,6 +128,13 @@ scm:
     bitbucket:
         oauthClientId: YOUR-BITBUCKET-OAUTH-CLIENT-ID
         oauthClientSecret: YOUR-BITBUCKET-OAUTH-CLIENT-SECRET
+        # The username and email used for checkout with bitbucket
+        #username: sd-buildbot
+        #email: dev-null@screwdriver.cd
+
+webhooks:
```
stjohnjohnson: I think you need to load this in the `/bin/server.js`

config/default.yaml
```diff
@@ -125,6 +128,13 @@ scm:
...
+    # Obtains the SCM token for a given user. If a user does not have a valid SCM token registered with Screwdriver, it will use this user's token instead.
+    #username: sd-buildbot
```
stjohnjohnson: I wouldn't leave this commented out

config/default.yaml
d2lam: Shouldn't matter though...

config/default.yaml
wahapo: Thank you for pointing out. Added.

config/default.yaml
wahapo: modified

plugins/webhooks/index.js
wahapo: removed.

config/default.yaml
wahapo: Modified.

PR #451, opened 2017-02-07
Make work with Python3 and Python2

dwighthubbard:

PR #456, opened 2017-02-13
fix: Correct padding of JWT secrets

stjohnjohnson:
Looks like the [Python2/3 change](https://github.com/screwdriver-cd/screwdriver/pull/451) broke how secrets are set in the template.  It's adding an extra newline before and after the JWT keys.


d2lam:
@stjohnjohnson  Which change? Link in PR description please?

PR #459, opened 2017-02-15
Fix: make the badge show status of the latest event

minz1027:
- Previously the badge only reflect the build status for main job
- Change it to get the last build status of the latest event

screen shot 2017-02-16 at 3 32 24 pm
screen shot 2017-02-16 at 3 35 00 pm

related to https://github.com/screwdriver-cd/screwdriver/issues/460

### DIFF
--- a/plugins/pipelines/badge.js
+++ b/plugins/pipelines/badge.js
@@ -9,23 +9,45 @@ const idSchema = joi.reach(schema.models.pipeline.base, 'id');
  * Generate Badge URL
  * @method getUrl
  * @param  {string}  badgeService    Template URL for badges - needs {{status}} and {{color}}
- * @param  {string}  [currentStatus] Current Build Status
+ * @param  {Array}  [buildsStatus]   Current status of all builds in the same event
  * @return {string}                  URL to redirect to
  */
-function getUrl(badgeService, currentStatus) {
-    const status = (currentStatus || 'UNKNOWN').toLowerCase();
+function getUrl(badgeService, buildsStatus = []) {
+    const counts = {};
+    const parts = [];
+    let worst = 'lightgrey';
+
     const statusColor = {
-        unknown: 'lightgrey',
         success: 'green',
-        failure: 'red',
-        aborted: 'red',
         queued: 'blue',
-        running: 'blue'
+        running: 'blue',
+        unknown: 'lightgrey',
+        failure: 'red',
+        aborted: 'red'
     };
-    const color = statusColor[status];
+
+    const levels = [
+        'success',
+        'queued',
+        'running',
+        'unknown',
+        'failure',
+        'aborted'
+    ];
+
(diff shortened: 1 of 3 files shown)


plugins/pipelines/badge.js
```diff
@@ -9,23 +9,36 @@ const idSchema = joi.reach(schema.models.pipeline.base, 'id');
...
-    const color = statusColor[status];
+
+    buildsStatus.forEach((status) => {
+        counts[status] = (counts[status] || 0) + 1;
+    });
+
+    Object.keys(statusColor).forEach((s) => {
+        if (counts[s]) {
+            parts.push(`${counts[s]} ${s}`);
+            worst = statusColor[s];
```
FenrirUnbound: This just sets the status to the last one checked. If `abort` came first and then `success` came second, then the status will be set to `success`.

FenrirUnbound: An Ojbect is unordered, so you can't necessarily rely on that. If you need order, consider using a map https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#Objects_and_maps_compared.

stjohnjohnson: I agree, I would do something more like:
```js
const LEVELS = [
  'success',
  'queued',
  'running',
  'unknown',
  'failure',
  'aborted'
];

worst = Math.max(LEVELS.indexOf(s), worst);
worstColor = statusColor[worst];
```

plugins/pipelines/badge.js
```diff
@@ -9,23 +9,36 @@ const idSchema = joi.reach(schema.models.pipeline.base, 'id');
...
+        failure: 'red',
+        aborted: 'red'
     };
-    const color = statusColor[status];
+
+    buildsStatus.forEach((status) => {
+        counts[status] = (counts[status] || 0) + 1;
+    });
+
+    Object.keys(statusColor).forEach((s) => {
```
stjohnjohnson: Please don't have vague variable names, this should really be `statusName` instead of `s`

plugins/pipelines/badge.js
minz1027: addressed. loop through the levels array instead

plugins/pipelines/badge.js
```diff
@@ -43,26 +65,29 @@ module.exports = () => ({
...
-                            const build = builds.pop();
+                        return lastEvent.getBuilds().then((builds) => {
+                            if (!builds || builds.length < 1) {
+                                return reply.redirect(getUrl(badgeService));
+                            }
+
+                            const buildsStatus = builds.reverse()
+                                .map(build => build.status.toLowerCase());
+
+                            for (let i = builds.length; i < lastEvent.workflow.length; i += 1) {
```
stjohnjohnson: This is not a good longterm approach, workflow is not always in sequential order.  But it will work for now.

Issue #458, opened 2017-02-15
Upload artifacts to store


d2lam:
Users want to be able to look at the artifacts after their builds finishes. Currently, we write artifacts to artifact folder (`/sd/workspace/artifacts`). However, this is destroyed after the build finishes. We should upload them to store, similar to how we upload the build logs. 

Related components:
https://github.com/screwdriver-cd/launcher
https://github.com/screwdriver-cd/log-service
https://github.com/screwdriver-cd/store

d2lam:
Confirmed this is working

PR #461, opened 2017-02-17
feat: register artifact bookend to upload artifacts

d2lam:
Registers artifact bookend, which uploads artifacts to S3
Blocked by: https://github.com/screwdriver-cd/artifact-bookend/pull/1
Related: https://github.com/screwdriver-cd/screwdriver/issues/458

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -140,6 +140,8 @@ bookends:
     # Plugins for build setup
     setup:
         - scm
+    teardown:
+        - screwdriver-artifact-bookend
 
 ecosystem:
     # Externally routable URL for the User Interface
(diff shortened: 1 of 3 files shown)


d2lam:
@FenrirUnbound  Thanks, I missed that.  :)

config/default.yaml
```diff
@@ -140,6 +140,8 @@ bookends:
     # Plugins for build setup
     setup:
         - scm
+    teardown:
+        - screwdriver-artifact-bookend
```
stjohnjohnson: Why is the one above called `scm`?

d2lam: I think because of this: https://github.com/screwdriver-cd/screwdriver/blob/master/bin/server#L39-L52

scm is actually `screwdriver-scm-${scmConfig.plugin}`, depending on what users configure as their scm plugin.

stjohnjohnson: That's upsetting :(  This is now confusing.

d2lam: Haha you should ask stjohn. I believe he wrote that piece of code :p

Issue #465, opened 2017-02-23
Email Notifications


d2lam:
> Users want to be able to be notified when their builds finish.

By default, only send when build fails
If status is specified, notifies based on the status. 

This should be configured per pipeline:
```yaml
shared:
   settings:
         email: foo@yahoo.com
```

Or per job:
```yaml
jobs: 
   component:
      settings:
             email: bar@yahoo.com
      steps: 
          ....
```

Email could be a string for simple configuration
```yaml
   email: bar@yahoo.com
```

or an object for more complicated configuration
```yaml
   email:
        addresses: 
              - foo@yahoo.com
              - bar@yahoo.com
        statuses:
              - success
              - failure
```

**Work:**
- [x] Data schema: accept `settings`: https://github.com/screwdriver-cd/data-schema/pull/107
- [x] Config-parser: to parse the new `settings` field inside `screwdriver.yaml`: https://github.com/screwdriver-cd/config-parser/pull/29
- [x] API: emits event when a build status updated
- [x] Notification Plugin: listens to the notify event, and sends email
- [x] Add documentation to the guide
- [x] Blog post

PR #469, opened 2017-02-27
feat(465): support notification plugins

d2lam:
Emits an event called `build_status` when a build status updated. 
Notification plugins will subscribe to this event and perform some notification actions.

Should be able to configure multiple notifications like:
```
notification:
   email: ...
   slack: ...
   hipchat: ...
```
Related: https://github.com/screwdriver-cd/screwdriver/issues/465
~Blocked by: https://github.com/screwdriver-cd/data-schema/pull/108~
Blocked by: `screwdriver-notifications-email` to be done so we can test it.


config/default.yaml
```diff
@@ -143,6 +143,12 @@ bookends:
     teardown:
         - screwdriver-artifact-bookend
 
+notification:
```
FenrirUnbound: If this is supposed to be a stub setting, I think YAML expects you to initialize it as an empty object `{}`

package.json
```diff
@@ -76,8 +76,8 @@
     "requestretry": "^1.12.0",
     "screwdriver-artifact-bookend": "^1.0.1",
     "screwdriver-build-bookend": "^2.0.1",
-    "screwdriver-config-parser": "^3.1.0",
-    "screwdriver-data-schema": "^16.0.1",
+    "screwdriver-config-parser": "^3.3.0",
```
FenrirUnbound: Until we have a more formalized process, we should at least note these versions in the compatibility document

config/default.yaml
d2lam: @FenrirUnbound  It is set as `{}` in registerPlugins.js. Is that what you meant?

d2lam:
@FenrirUnbound  @petey  Sorry but I updated the PR to send more fields. The change is mainly in `build/update.js`. Could you guys please re-review! :)

config/default.yaml
FenrirUnbound: Kinda like how this is set to an empty object: https://github.com/screwdriver-cd/screwdriver/blob/2af654612965eeaac5719b0cff6c3f1c5393d3a8/config/default.yaml#L104

Issue #470, opened 2017-02-28
Job Templates


d2lam:
> As a template owner, I want to be able to publish a template for other people to use. 

Example `screwdriver-template.yaml`:
```yaml
name: test
version: 1.3
description: template for testing
maintainer: foo@bar.com
config:
  image: node:6
  steps:
    - install: npm install 
    - test: npm test
    - echo: echo $FOO
  environment:
    FOO: bar
  secrets:
     - NPM_TOKEN
```

To publish a template, I can edit the `screwdriver-template.yaml`, and run the pipeline to publish a new version.

Publishing a new version should do these steps:
- Validate the template
- Check if permission is good
- Publish the template

> As a user, I want to be able to reuse existing configuration for a job. 

In user's `screwdriver.yaml`:
```yaml
jobs:
   main:
      template: test@1.3
```

Would be translated to:
```yaml
jobs:
   main: 
        image: node:6
        steps:
          - install: npm install 
          - test: npm test
          - echo: echo $FOO
        environment:
           FOO: bar
        secrets:
          - NPM_TOKEN
```

# Implementation Details
## First pass: MVP

*Work to be done:*
### Template owner
- [x] Validate expected fields in a `screwdriver-template.yaml`
    - We have added validation into the data-schema repo for the required fields in a template.yaml
    - We have created a module for validating templates that will use the changes made in the data-schema
   - [x] Data schema (https://github.com/screwdriver-cd/data-schema/pull/109)
   - [x] Template validation module (https://github.com/screwdriver-cd/template-validator/pull/3)
   - [x] API new endpoint to validate template (screwdriver-cd/screwdriver#480)
- [x] Publish template (PR open, https://github.com/screwdriver-cd/template-main/pull/2)
- [x] Store template metadata in RDS
      2 new endpoints, GET and POST for getting and creating templates: name, version, description, maintainer, config, label, scmURI
    - `label` is an array. It could be anything like `['test', 'stable', 'v123']`
    - scmURI is there so that we could check if the correct pipeline is publishing the template
    - [x] Data-schema
    - [x] Models (https://github.com/screwdriver-cd/models/pull/149)
    - [x] API (https://github.com/screwdriver-cd/screwdriver/pull/475)

### Template user
- [x] Validate template syntax in a `screwdriver.yaml`
- [x] Merge into jobs

### Others
- [x] Cucumber feature file

## Second pass:

*Need more ideas/ architecture designs*
[...]

PR #475, opened 2017-03-01
feat(470): Add template endpoints

d2lam:
~~Blocked by data-schema~~
Related https://github.com/screwdriver-cd/screwdriver/issues/470

This adds endpoints to create, get, and list templates. 
Create a template will store the template data (`config`, `name`, `version`, `description`, `maintainer`, `labels`) into the datastore.

- Template does not exist: create a new entry
- Template with same name, new version: create a new entry
- Template with same name, same version: ~update the labels~ throw 409
- Bad permission: will return 401.
- This endpoint is only accessible in `build` scope right now.


package.json
```diff
@@ -76,12 +76,13 @@
...
     "screwdriver-build-bookend": "^2.0.1",
-    "screwdriver-config-parser": "^3.1.0",
-    "screwdriver-data-schema": "^16.0.1",
+    "screwdriver-config-parser": "^3.3.0",
+    "screwdriver-data-schema": "^16.1.1",
     "screwdriver-datastore-sequelize": "^2.0.0",
     "screwdriver-executor-docker": "^2.0.0",
     "screwdriver-executor-k8s": "^10.0.0",
     "screwdriver-models": "^21.2.0",
+    "screwdriver-notifications-email": "^1.0.3",
```
d2lam: I'm not sure what's up with `git`, but this is not correct. This line is already in master ?!?

FenrirUnbound:
```
$ npm view screwdriver-notifications-email versions
[ '0.0.1', '1.0.0', '1.0.1', '1.0.2', '1.0.3', '1.1.0' ]
```

d2lam: From master branch: https://github.com/screwdriver-cd/screwdriver/blob/master/package.json#L85

FenrirUnbound: > I'm not sure what's up with git, but this is not correct. This line is already in master ?!?

Maybe I'm misunderstanding what you're comment is. I thought you were asking if version `1.0.3` exists for `screwdriver-notifications-email`. When I checked the NPM Registry, it says that it has version `1.0.3` in the list of published versions.

Were you referring to something else that does not have version `1.0.3`?

d2lam: Hmmm I mean git is saying that I added this line `"screwdriver-notifications-email": "^1.0.3"` in this PR, but I didn't add it. This line is already there in the master branch.

plugins/templates/create.js
```diff
@@ -0,0 +1,92 @@
...
+            const pipelineFactory = request.server.app.pipelineFactory;
+            const templateFactory = request.server.app.templateFactory;
+            const token = request.auth.credentials;
+            const pipelineId = token.pipelineId;
+            const name = request.payload.name;
+            const labels = request.payload.labels || [];
+
+            return Promise.all([
+                pipelineFactory.get(pipelineId),
+                templateFactory.get({ name })
```
d2lam: This always returns null for some reason. Even if the template with that name already exists

package.json
FenrirUnbound: ❓ Super weird.

Just a wild guess, but it *might* correct itself after rebasing with master. If you want to pursue it on your local machine, I would suggest doing it on a separate branch so you don't lose all your hard work. Someone in the group has already lost their hard work to a git rebase gone wrong; don't be the next victim!

plugins/templates/create.js
d2lam: Thanks to @FenrirUnbound 
This is because https://github.com/screwdriver-cd/models/blob/master/lib/baseFactory.js#L75

It expects all the keys to be there.

d2lam:
Not sure why it says `Coverage decreased (-0.08%) to 99.916%` 😢 
Coverage stays the same as before.

d2lam:
@FenrirUnbound  Can you re-review please :D

plugins/templates/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+**Arguments**
+
+'name', 'version', 'description', 'maintainer', labels
+
+* `name` - Name of the template
+* `version` - Version of the template
+* `description` - Description of the template
+* `maintainer` - Maintainer of the template
+* `labels` - Labels of the template. This field is optional and should be an array.
```
FenrirUnbound: If the list of labels is finite, should we consider documenting that list here?

plugins/templates/create.js
```diff
@@ -0,0 +1,84 @@
...
+                    labels
+                });
+
+                // If template doesn't exist yet, just create a new entry
+                if (templates.length === 0) {
+                    return templateFactory.create(templateConfig);
+                }
+
+                // If template exists, but this build's scmUri is not the same as template's scmUri
+                // Then this build does not have permission to publish
```
FenrirUnbound: The scenario of "what happens when the SCM URI changes" didn't seem to come to a final solution. Do we intend on either documenting this scenario, or plan on addressing it in a future pass?

plugins/templates/create.js
```diff
@@ -0,0 +1,84 @@
...
+                if (pipeline.scmUri !== templates[0].scmUri) {
+                    throw boom.unauthorized('Not allowed to publish this template');
+                }
+
+                // If template exists and has good permission, check the exact version
+                return templateFactory.get({ name, version })
+                    .then((template) => {
+                        // If the version doesn't exist, create a new entry
```
FenrirUnbound: From a REST perspective, I would have expected a `PUT` for being able to specifically create a resource. Looking at how the endpoint responds to the same input, I would also expect it to be a `PUT` because it's an idempotent operation.

For instance, with jobs or builds, we allow you to update the resource directly with a `PUT`. By specifying which build or job (e.g. `/builds/{id}` or `/jobs/{id}`), I can update them.

plugins/templates/README.md
d2lam: Yeah, we should. But it's not finite though.

plugins/templates/README.md
```diff
@@ -0,0 +1,75 @@
...
+### Routes
+
+#### Get all templates
+`page` and `count` optional
+
+`GET /templates?page={pageNumber}&count={countNumber}`
+
+#### Get single template
+
+`GET /templates/{id}`
```
FenrirUnbound: For builds and jobs, the `id` is a number. What's the `id` for a template: is also a number, or is it the name of the template+version number?

plugins/templates/create.js
d2lam: We didn't come to a conclusion for that. I personally think that if the repo changes, it should be a different template. Otherwise, I don't know what to use for permission control. But we can talk about that more in the 2nd pass.

plugins/templates/README.md
d2lam: `id` is still a number. `template + version` is the primary keys. Similar to how `scmUri` is the primary key for pipeline, but GET for pipelines is still via `id`

plugins/templates/README.md
FenrirUnbound: Ah ok. So there's no limitation on what label an owner tags their template as 👍

plugins/templates/create.js
```diff
@@ -0,0 +1,84 @@
...
+            })
+            .then((template) => {
+                const location = urlLib.format({
+                    host: request.headers.host,
+                    port: request.headers.port,
+                    protocol: request.server.info.protocol,
+                    pathname: `${request.path}/${template.id}`
+                });
+
+                return reply(template.toJson()).header('Location', location).code(201);
```
FenrirUnbound: Is it a `201` regardless if a template is updated or created?

plugins/templates/create.js
d2lam: This endpoint is more for `CREATE`, so that's why I made it a `POST`. 
The only time it does an update is when the exact same name and version already exist. And I added that guard because of what we discussed here: https://github.com/screwdriver-cd/screwdriver/issues/470#issuecomment-283182395 on how to publish a template. We would always call a `POST /templates` in the yaml. It's a little confusing to do an update inside a create, but I think `POST` is more appropriate since it's supposed to create, not update. And the same version is a special case. Also, we don't have `id` so we can't do`PUT /templates/{id}`.

plugins/templates/create.js
d2lam: It's for `create`. Maybe I should return `200` if it's an update.

plugins/templates/create.js
FenrirUnbound: In the end, a `PUT` and a `POST` can create. The difference is the side effects they cause when you call them twice. A `PUT` operation is expected to have the same outcome regardless if it's called once or twice. The way this endpoint is structure, that seems to be the case.

> Also, we don't have id so we can't do `PUT /templates/{id}`

As an alternative to `PUT /templates/{id}`, couldn't there be a more RESTful endpoint along the lines of `PUT /templates/<templateName>/<version>`? The caller still has all the appropriate information (name of the template and version). 

In addition, it allows us to have `GET /templates/<templateName>/<version>` in the future to query for a specific version.

plugins/templates/README.md
tkyi: Or should we limit it to like 5 or 10? Since anything more than that might just get ridiculous.

d2lam: Why do we want to limit it? It's just an array.

plugins/templates/create.js
d2lam: I agree with you that `PUT` and `POST` can do a create. Both can do an update as well. However, if you look at the way we are planning to use this endpoint at https://github.com/screwdriver-cd/screwdriver/issues/470#issuecomment-283182395, I imagine this endpoint will be called inside a step in `screwdriver.yaml` for publishing template. Currently, it will just need to call `POST` and pass the whole template in as the payload. If we change it to `PUT/templateName/version`, we would need to do extra steps to parse the name and version before calling this.

tkyi: Yeah to add to Dao's statement, the path would not be consistent with all the other existing API calls we have

FenrirUnbound: The other API calls we have don't have an alternative way of calling them. There's no alternative to fetching the 15th job other than calling `/jobs/15`.

For templates, a caller cannot interact with the API without the `id`. The caller gives the API the template name and version number. If the API infers information from the body anyways, why not just make the URI reflect the same information?

d2lam: I'm simplifying this logic to just `CREATE` and throw `409` if already exists. 
We will have separate endpoints that allow users to add/remove labels for a template. We need to think about permission control here. Will do this in 2nd pass. Right now it's not blocking.

plugins/templates/create.js
```diff
@@ -0,0 +1,81 @@
+'use strict';
+
+const boom = require('boom');
+const schema = require('screwdriver-data-schema');
+const hoek = require('hoek');
+const urlLib = require('url');
+
+module.exports = () => ({
+    method: 'POST',
+    path: '/templates',
```
FenrirUnbound: This path still seems very contrived. What does it look like to fetch a template? Does the user need to hit an endpoint like `/templates/{id}`? Does the endpoint look like `/templates/<templateName>/<version>`?

FenrirUnbound: What makes a template so special that it needs to be by {id} and not by it's name and version?

There are numerous other services in the real world that follow a more RESTful pattern. Docker registries (like Docker Hub) enable users to query an image by its name and not some SHA representation of it.

d2lam: User should be able to do a `GET templates/{id}`, similar to `GET pipelines/{id}`

If they want to fetch by `name`, or `name+version`, they would use a query like `GET templates?name=node&version=123` or `GET templates?name=node`. It's more like a LIST. This is not implemented in this PR. It's the 3rd bullet point here (under using templates): https://github.com/screwdriver-cd/screwdriver/issues/470#issuecomment-285238348

FenrirUnbound: If you know exactly what you're looking for, why be opposed to being a RESTful API and not do `/templates/<templateName>/version/<version>`?

What makes that different than `/templates/<templateName>` for listing all the versions and `/templates/<templateName>?version=2` for listing all versions >= 2.0? Or even `/templates/<templateName>?labels=stable`? 

This is nothing special to what other services we own. Many services cover this already; what makes *this* API so special that it's breaking from generally accepted principles?

tkyi: Can you provide some links to examples?

d2lam: @FenrirUnbound  Yeah. I'm open to having that as well. We need to think more about it. We might even need to change the naming convention for templates, since currently we allow `/` in the name. But I would say it doesn't belong to the scope of this PR. This PR provides the base `POST` to create, and `GET /id` to get which might be helpful in some cases (and we do have that consistently for other models). 

But we do need to implement the `/templates/name/versions/version`, or `templates?name=name&version=version`, or whichever way we decide to implement.

FenrirUnbound: {id} coincidentally is a number as well as what distinguishes uniqueness. There isn't an alternative way to distinguish one unique build from another, besides its build number. There cannot be two screwdriver-cd/screwdriver/builds/1. 

Templates are different. The characteristics that define a template from another are know before they do the create. What makes a template unique is the name and the version; there cannot be another instance of the same combination.

Look at this repository: screwdriver-cd/screwdriver. Github associates an {id} number to this repository too; where is it?

d2lam: Template is not any different from `pipeline` or other models. For example, `scmUri` is a unique field of a pipeline. There cannot be any two pipelines with the same `scmUri`. Same as `job`. There cannot be any two jobs with the same `pipelineId +  name`

I'm not opposing to having an endpoint to get by name and version (we need it). But I don't think there is any harm in having a `GET /{id}` since we do have this endpoint consistently for ALL the models.

d2lam:
@FenrirUnbound  Please re-review? :D

plugins/templates/create.js
FenrirUnbound: > For example, scmUri is a unique field of a pipeline.
`scmuri` is not a consistent form. It can be `https` or `git` protocols. 

> There cannot be any two jobs with the same pipelineId + name

The initial design for the job was to have it canonical and use a `/pipelines/{id}/jobs/{job}`. We then switched it to `/pipelines/{id}/jobs/{id}`. Do you recall the problems we had with it?
 
Templates with different names are unique. Just because you have an `id` associated with all the other URIs, it doesn't mean this one has to also. Github repositories have an associated `id` yet that is not exposed to the user. 

We've already solved this problem so I'm confused why this is a contention point. What makes this so unique that it can't follow the same pattern?

@stjohnjohnson even echoes a similar concern here: https://github.com/screwdriver-cd/config-parser/pull/32#discussion_r105335166 A POST is never cached anyways, but the endpoint structure is still applicable here.

d2lam: Not commenting on the `GET` as it's not part of this PR. 
For the `CREATE`, if we choose to do `PUT /templates/:name/:version` instead of `POST /templates`, if the `name` in the `param` is not the same as the `name` in the `payload`, what is the expected behavior? Do we throw an error? Do we favor the name in the `param`, or the `payload`? We need to make this clear to the users. Same question for `version`.

plugins/templates/README.md
tkyi: Just to force people to only choose the most relevant labels and not get too crazy with them

plugins/templates/create.js
d2lam: Actually the more I think about it, the more I incline to make this `POST /templates`. 

**The only time a new version should be published is when something in `template.yaml` changes.** (Please correct me if I'm wrong @stjohnjohnson, I remember this from our conversation) If this is true, if they want to publish a new version, they need to change the `template.yaml` anyway. In that case, they can change the `version` (major, minor, and patch) in the yaml as well. We don't need the "smart detection" that @stjohnjohnson  mentioned about not publishing if nothing in `template.yaml` changes. 

I'm not in favor of `PUT /templates/:name/:version` because we need to detect what exact version to put. If we don't want to do the auto-bumping version for the users, then the user needs to know exactly what version it is. And "user" in this case is actually `template-template` (our template to publish templates), which means it would need to change every time, and that doesn't make sense to me. 

@FenrirUnbound  @petey  @stjohnjohnson

Issue #431, opened 2017-01-19
Executor Queue MVP


stjohnjohnson:
> As a user, I want my build to enqueue for later execution if all the execution engines are offline or full, so that I do not have to worry about forgotten builds

Right now, Screwdriver directly calls the executor when a build is to be started.  If the executor target is offline or full, the build is generally dropped with an error message.  This isn't a good user experience, especially if the error is hidden due to the build being started from a PR or commit.

Instead, Screwdriver should be pushing planned builds into a queue.  And all available executors should be listening on that queue and taking items off when they have capacity.  This will allow us to distribute between N number of Docker Swarm clusters for example.

Rules:
 - Executors can determine if they are capable of running a new build.
 - Multiple instances of the same executor can connect to the queue.
 - Multiple executor types can connect to the same queue.
 - JWT is not created until build is sent to the executor.
 - Users know how far in the queue a build is.

---

Notes:
 - We should investigate what other technologies like Travis or Wercker are using for this internally.

--- 
Edit (petey): Removed work items - See list below

d2lam:
@tk3fftk  @catto  It would be great if you guys could update the issue with your findings, decisions, progress, etc. Also feel free to let us know if you have any questions/concerns! :)

catto:
@d2lam 🙆  We have started designing an architecture of the executor queue and are going to share it with you when completed.

d2lam:
@catto awesome! Thanks for the update!

catto:
I talked to @d2lam today, and we have decided that we are going to work on #300 instead of this because we don't have enough time to do.

jerryzhang222:
[Travis queuing](https://github.com/travis-ci/travis-scheduler/)

From [what I've seen](https://github.com/travis-ci/travis-scheduler/blob/76ca2e2d9f382751981dbd33bb7c6be626bd38c0/lib/travis/scheduler/worker.rb), Travis uses [Sidekiq](https://github.com/mperham/sidekiq) to queue up builds and perform tasks. They have [some code](https://github.com/travis-ci/travis-scheduler/blob/76ca2e2d9f382751981dbd33bb7c6be626bd38c0/lib/travis/scheduler/service/notify.rb) that uses RabbitMQ (AMQP), but I don't see this code used anywhere and it appears that they are trying to [phase RabbitMQ out](https://github.com/travis-ci/travis-scheduler/tree/amqp-removal).

petey:
[...]

PR #477, opened 2017-03-01
feat(431): Add executor queue feature file

kumada626:
Add cucumber feature file of executor queue.
How does it look?

Related:#431

### DIFF
--- a/features/executor-queue.feature
+++ b/features/executor-queue.feature
@@ -21,10 +21,10 @@ Feature: Executor Queue
         Then that build is pushed into the queue
         
         Examples:
-            | type   |
-            | k8s    |
-            | docker |
-            | j5s    |
+            | type    |
+            | k8s     |
+            | docker  |
+            | jenkins |
 
     @ignore
     Scenario Outline: Run a queued build
@@ -35,10 +35,10 @@ Feature: Executor Queue
         And that build is running
 
         Examples:
-            | type   |
-            | k8s    |
-            | docker |
-            | j5s    |
+            | type    |
+            | k8s     |
+            | docker  |
+            | jenkins |
     
     @ignore
     Scenario Outline: No executors can handle a new build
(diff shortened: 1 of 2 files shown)


kumada626:
Fixed validator.output.json because of updating data-schema.

features/executor-queue.feature
```diff
@@ -0,0 +1,33 @@
...
+    Users want their build to enqueue for later execution if all the execution engines are
+    offline or full, so that they do not have to worry about forgotten builds.
+
+    Screwdriver should be pushing planned builds into a queue. And all available executors
+    should be listening on that queue and taking items off when they have capacity.
+
+    Rules:
+        - Executors can determine if they are capable of running a new build.
+        - Multiple instances of the same executor can connect to the queue.
+        - Multiple executor types can connect to the same queue.
```
FenrirUnbound: It looks like these rules were inherited from feature #431. I don't see this specific scenario being covered. Is it safe to assume that this scenario won't be covered yet?

petey: There should probably be a test for when an executor `can not` handle a new build.

kumada626: Added a test case when an executor can not handle a new build, and fixed all tests cope with multiple executor types.

features/executor-queue.feature
```diff
@@ -0,0 +1,68 @@
...
+    Scenario Outline: Push a build into the queue
+        Given an executor type is <type>
+        When a new build is created
+        Then that build is pushed into the queue
+        
+        Examples:
+            | type   |
+            | k8s    |
+            | docker |
+            | j5s    |
```
stjohnjohnson: On a side note (not a blocker here) I really want to rename `j5s` into `jenkins`

kumada626: Renamed j5s into jenkins.

PR #480, opened 2017-03-03
feat(470): Template validation as an API endpoint

FenrirUnbound:
## Context

Template owners are responsible for creating and maintaining their templates for others to use. Along with descriptive and meaningful documentation, we would like to empower them with tools that will help them in developing and updating their templates. One tool that we currently would offer is an API endpoint for template validation, which an owner can leverage to make sure their template is syntactically valid.

## Objective

~Create a feature scenario regarding how a user could use the API to validate their template.~ (Deferred to another PR)

Create an endpoint that allows users to validate and verify a template configuration.

## Progress

- [x] Endpoint for template validation
- [x] Enforcing format of user input
- [x] Feature file
- ~Refactor other validation endpoint?~
   - ~May~ To be performed in a separate PR as a clean-up

## Miscellaneous 

* Relates to screwdriver-cd/screwdriver#470
* Requires updates made to `template-validator` via screwdriver-cd/template-validator#6
* Requires updates made to `data-schema`
   * screwdriver-cd/data-schema#111
   * screwdriver-cd/data-schema#112
* Split the functional testing and feature file work to PR screwdriver-cd/screwdriver#490

EDIT1: Refined scope of PR
EDIT2: More notes


plugins/template-validator.js
```diff
@@ -0,0 +1,31 @@
...
+        path: '/validator/template',
+        config: {
+            handler: (request, reply) => {
+                // rip eslint
+                const result = validator(request.payload.yaml)
+                .then((stuff) => {
+                    reply(stuff);
+                }, (validationErr) => {
+                    reply({
+                        error: validationErr
```
FenrirUnbound: Point of feedback: What does the error case here look like for the caller?

The validator for a `screwdriver.yaml` returns a default configuration file when an error is encountered. We don't have a good default configuration for an invalid `sd-template.yaml` file.

plugins/template-validator.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
+
+const validator = require('screwdriver-template-validator');
+
+exports.register = (server, options, next) => {
+    server.route({
+        method: 'POST',
+        path: '/validator/template',
```
FenrirUnbound: Point of feedback: What does the endpoint for the template validator look like?

Alternative suggestions were:

1. Overload the `/validator` endpoint

To distinguish between validating a `sd-template.yaml` and a standard `screwdriver.yaml`, the caller could pass a parameter to distinguish a type. For instance, it could be `/validator?type=template` to communicate that the caller wants to validate a template.

This felt too ambiguous and overloading an endpoint for different configuration types could lead to confusion on the part of the user.

2. Use a `/validator/<type>` endpoint

This is the current implementation. If this is the preferred route, we would want to revisit the current `/validator` endpoint and refactor it. It could be refactored into `/validator/screwdriver`, `/validator/user`, or whatever type makes sense to describe it.

d2lam: I think `/validator/template` makes sense. 
Kinda like how we have `sync/webhooks` and `sync/pullrequests`

plugins/template-validator.js
d2lam: The `screwdriver.yaml` needed a default case because otherwise the job won't even start. For this, why do we need a default config?

FenrirUnbound: We were using the `config-parser` as a guide for how to implement the `template-validator`. As you pointed out, this is one case where modeling the `template-validator` after the `config-parser` does not make sense. So we're looking for what is better for the caller.

After some thought, one alternative could be that this endpoint returns the following payload format: 

```
{
  data: {
    config: {
      // The template configuration object, or an empty object if it can't be parsed 
      ...
    },
    errors: [
      // The number of errors encountered while validating the template configuration
      // It would be empty if no errors were detected
      ...
    ]
  }
}
```

This payload has the benefit of giving the caller:

1. What the template looks like in Screwdriver. 
1. A list of errors that the validator found

Given these two things, the template owner has a minimum amount of information to determine why their template is wrong or is formatted differently than expected.

plugins/template-validator.js
```diff
@@ -0,0 +1,47 @@
...
+        config: {
+            description: 'Validate a given sd-template.yaml',
+            notes: 'returns the parsed config, validation errors, or both',
+            tags: ['api', 'validation', 'yaml'],
+            handler: (request, reply) => {
+                const templateString = request.payload.yaml;
+
+                return validator(templateString)
+                .then(reply, err =>
+                    reply(boom.badRequest(err))
```
d2lam: Just formatting: can we have this on the same line as above? It would be easier to read.

plugins/template-validator.js
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Hapi Template Validator Plugin
+ *  - Validates sd-template.yaml and returns the parsed template with any
+ *    errors associated with it
+ * @method register
+ * @param  {Hapi.Server}    server
+ * @param  {Object}         options
+ * @param  {Function} next
+ */
+exports.register = (server, options, next) => {
```
d2lam: Need to register this in https://github.com/screwdriver-cd/screwdriver/blob/master/lib/registerPlugins.js#L12-L20

PR #490, opened 2017-03-08
feat(470): cucumber tests for templates

FenrirUnbound:
## Context

Functional testing helps us assert that new functionality works as expected, in addition to ensuring that it does not negatively affect other features. For the template validation feature, we want to include a set of test scenarios that represent how a user would utilize the functionality.

## Objective

A cucumber test to cover the scenarios for a template owner to validate their template.

## Miscellaneous

* Contributes to screwdriver-cd/screwdriver#470 
* Extended form PR screwdriver-cd/screwdriver#480

EDIT 1: Included the step definitions along with the feature file

### DIFF
--- a/features/step_definitions/templates.js
+++ b/features/step_definitions/templates.js
@@ -1,0 +1,70 @@
+'use strict';
+
+const fs = require('mz/fs');
+const request = require('../support/request');
+const path = require('path');
+const Assert = require('chai').assert;
+
+module.exports = function () {
+    this.Given(/^a (valid|invalid)\b job-level template$/, (templateType) => {
+        let targetFile = '';
+
+        switch (templateType) {
+        case 'valid':
+            targetFile = path.resolve(__dirname, '../data/valid-template.yaml');
+            break;
+        case 'invalid':
+            targetFile = path.resolve(__dirname, '../data/invalid-template.yaml');
+            break;
+        default:
+            return Promise.reject('Template type is neither valid or invalid');
+        }
+
+        return fs.readFile(targetFile, 'utf8')
+        .then((contents) => {
+            this.templateContents = contents;
+        });
+    });
+
+    this.When(/^they submit it to the API$/, () =>
+        this.getJwt(this.accessKey)
+        .then((response) => {
+            const jwt = response.body.token;
+
+            return request({
+                uri: `${this.instance}/${this.namespace}/validator/template`,
+                method: 'POST',
+                auth: {
+                    bearer: jwt
+                },
+                body: {
+                    yaml: this.templateContents
+                },
+                json: true
+            });
+        })
+        .then((response) => {
+            Assert.equal(response.statusCode, 200);
+
+            this.body = response.body;
+        })
+    );
+
+    this.Then(/^they are notified it has (no|some) errors$/, (quantity) => {
+        switch (quantity) {
+        case 'no':
+            Assert.equal(this.body.errors.length, 0);
+            break;
+        case 'some':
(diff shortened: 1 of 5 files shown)


FenrirUnbound:
> @tkyi 
> Should we put a summarized description for the templates?

I would like there to be one, but I haven't been able to come up with a good enough summary. I was hoping @d2lam would chime in with her input 😸

features/templates.feature
```diff
@@ -0,0 +1,19 @@
+@templates
+Feature: Templates
+
+    Description from screwdriver-cd/screwdriver#470
+
+    Background:
+        Given "calvin" has admin permission to the template
+
+    @ignore
+    Scenario Outline: A template owner wants to validate their template
```
minz1027: hmm, from this PR: https://github.com/screwdriver-cd/screwdriver/pull/480/files, looks like we don't enforce any permission control for call validate endpoint. Looks like we don't need `Backgournd: Given "calvin" has admin permission to the template` for these scenarios?

FenrirUnbound: You're right. The scenario does state that Calvin has admin permissions, even though we don't enforce it. It's also intentional that we don't describe a scenario where a non-admin user interacts with the endpoint.

features/step_definitions/templates.js
```diff
@@ -0,0 +1,70 @@
+'use strict';
+
+const fs = require('mz/fs');
+const request = require('../support/request');
+const path = require('path');
+const Assert = require('chai').assert;
+
+module.exports = function server() {
```
FenrirUnbound: fix

features/templates.feature
```diff
@@ -0,0 +1,14 @@
+@templates
+Feature: Templates
+
+    Description from screwdriver-cd/screwdriver#470
```
r3rastogi: Should we put the actual description in here?

FenrirUnbound: idk what parts to copy-paste from the issue

features/step_definitions/templates.js
```diff
@@ -0,0 +1,70 @@
+'use strict';
+
+const fs = require('mz/fs');
+const request = require('../support/request');
+const path = require('path');
+const Assert = require('chai').assert;
+
+module.exports = function () {
+    this.Given(/^a (.*) job-level template$/, (templateType) => {
```
tkyi: maybe `/^a (valid|invalid) job-level template$/` to be more specific?

features/step_definitions/templates.js
```diff
@@ -0,0 +1,70 @@
...
+            });
+        })
+        .then((response) => {
+            Assert.equal(response.statusCode, 200);
+
+            this.body = response.body;
+        })
+    );
+
+    this.Then(/^they are notified it has (.*) errors$/, (quantity) => {
```
tkyi: `/^they are notified it has (no|some) errors$/`

features/step_definitions/templates.js
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        return fs.readFile(targetFile, 'utf8')
+        .then((contents) => {
+            this.templateContents = contents;
+        });
+    });
+
+    this.When(/^they submit it to the API$/, () =>
+        this.getJwt(this.accessKey)
```
minz1027: u do not need the jwt right, since the enpoint requires no auth?
https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/template-validator.js

FenrirUnbound: It returns a 403 without the JWT

PR #492, opened 2017-03-08
fix: Whitelist jpae

tkyi:
We would like jpae to get access to the screwdriver cluster so he can run through our demo (https://github.com/screwdriver-cd/fuji-demo)


r3rastogi:
👋 jpae :)

Issue #293, opened 2016-10-20
Metadata


stjohnjohnson:
> As a user, I want to pass structured data between my jobs, so that I can make informed decisions about what to deploy or test

Currently, there are almost no built-in ways to pass information between builds. The most common way is to use the value of the current Git tag.  But that data cannot be guaranteed to be correct.

Given this simple pipeline:

> main -> deploy -> promote

The `deploy` job would need to know the version of the application that was published in the `main` job, while the `promote` job would need to know the version that was tested in the `deploy` job.  This information is not available with just that git tag.

Instead, Screwdriver should provide a simple interface for reading/writing values that can be passed between builds in the same event.

Rules:
- `meta` is a command-line client (written in go) that is made available to the build container.
- Metadata should be passed from one job to the next in succession (in the same event) (see appendix A).
- When combining the results of matrix builds, they should be merge/replaced in a random order (to ensure users don't depend on a side-effect).
- When the event is done, a record of the metadata should be stored there as well.
- Metadata can contain strings, numbers, boolean, objects, and arrays.  The interface should account for this (see appendix B).
- `meta set` should not output anything, except during failures (`STDERR`).
- `meta get` should output in string representation with no newlines (see appendix C).

Stretch (gold bar):
- `meta` can read in from a file or STDIN.
- `meta get --previous` can read from the previous successful event.

---
### Preparation
- [x] Cucumber feature file
- [ ] Meta interface (discussed in this issue)
### Implementation
- [ ] Screwdriver.yaml changes
- [ ] Launcher changes
- [ ] API changes
- [ ] UI changes
- [ ] Store changes
- [ ] Schema changes
- [ ] Model changes
- [ ] Datastore changes
- [ ] Executor changes
### Outcome
- [ ] Cucumber feature passing
- [ ] Documentation

---
### A - Metadata passing

Example pipeline:

> main -> package -> deploy
#### main
- starts with:
  `{}`
- runs:
  `$ meta set coverage 99.52`
- results in:
  `{ "coverage": 99.52 }`
#### package
- starts with:
  `{ "coverage": 99.52 }`
- runs:
  `$ meta set version v3.1.2`
- results in:
  `{ "coverage": 99.52, "version": "v3.1.2" }`
#### deploy
- starts with:
  `{ "coverage": 99.52, "version": "v3.1.2" }`
- runs:
  `$ meta set url http://example.com`
- results in:
[...]

PR #494, opened 2017-03-09
fix(293): update current build before triggering next build

minz1027:
Move build.update() before triggering next build, otherwise the next build won't be able to get the `meta` from previous build when it starts.

Related to https://github.com/screwdriver-cd/screwdriver/issues/293

### DIFF
--- a/plugins/builds/update.js
+++ b/plugins/builds/update.js
@@ -68,62 +68,63 @@ module.exports = () => ({
                     build.status = desiredStatus;
 
                     // Only trigger next build on success
-                    return build.job.then(job => job.pipeline.then((pipeline) => {
-                        request.server.emit('build_status', {
-                            settings: job.permutations[0].settings,
-                            status: build.status,
-                            pipelineName: pipeline.scmRepo.name,
-                            jobName: job.name,
-                            buildId: build.id,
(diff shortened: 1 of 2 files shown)


FenrirUnbound:
Is there really no test associated with the order?

plugins/builds/update.js
```diff
@@ -68,62 +68,63 @@ module.exports = () => ({
...
-                            buildLink: `${buildFactory.uiUri}/builds/${id}`
-                        });
-
-                        // Guard against triggering non-successful builds
-                        if (desiredStatus !== 'SUCCESS') {
-                            return null;
-                        }
-
-                        const workflow = pipeline.workflow;
+                    return build.update()
```
tk3fftk: In my local test environment, parent build status is still RUNNING at the beginning of `sd-setup-builder`.
How is your local environment?
I think that It looks need waiting status change.

minz1027: The status is success for me. U need to change ur docker-compose file to pull in the api on this branch instead of pulling an image. Let me send u my docker-compose file on slack

minz1027:
@FenrirUnbound added a line in the test to assert the order of function calls.. But i felt it's weird though. Seems like that is testing if promise...then is executing in order?

PR #496, opened 2017-03-13
feat(470): Add feature file for templates

d2lam:
- Feature file for templates

### DIFF
--- a/features/templates.feature
+++ b/features/templates.feature
@@ -1,7 +1,42 @@
 @templates
 Feature: Templates
 
-    Description from screwdriver-cd/screwdriver#470
+    Templates are simply an existing configuration for a job that can be reused.
+    A basic example would be something like this:
+        name: node/test
+        version: 1.0.0
+        description: Template for testing
+        maintainer: foo@bar.com
+        config:
+            image: node:6
+            steps:
+                - install: npm install
+                - test: npm test
+
+    To use a template, users would consume it like this in their screwdriver.yaml:
+        jobs:
+            main:
+                template: node/test@1
+
+    The user's config would be translated to:
+        jobs:
+            main:
+                image: node:6
+                steps:
+                    - install: npm install
+                    - test: npm test
+
+    Rules:
+        - Template owner can publish a template if
+              - The format of template is valid
+              - The build that is used to publish has the right permissions
+              - No existing version already exists
+        - Publishing a template should be stored to be retrieved later.
+        - User can use a template by specifying it in their job. If they specify
+              - test/template:            pulls the latest version of test/template
+              - test/template@1:          pulls the latest version of major 1 for test/template
+              - test/template@1-stable:   pulls the latest version of major 1 with label "stable" for test/template
+              - test/template@1.2:        pulls the latest patch of 1.2 of test/template
 
     Scenario Outline: A template owner wants to validate their template
         Given a <template> job-level template
@@ -12,3 +47,35 @@ Feature: Templates
           | template | number |
           |  valid   |   no   |
           |  invalid |  some  |
+
+    @ignore
+    Scenario Outline: Merges template into user's config
+        Given an existing pipeline with job main
+        And an existing template called Test
+        And user specifies template Test in job main
+        When user starts the pipeline
+        Then job main executes what is specified in template
+
+    @ignore
+    Scenario Outline: Job config takes precedence over template config
+        Given an existing pipeline with job main
+        And an existing template called Test
+        And user specifies template Test in job main
+        And user has some settings defined
+        And template Test has the same settings with different values
+        When user starts the pipeline
+        Then settings is the job settings
+
+    @ignore
+    Scenario Outline: Publish template stores the template
+        Given a template with a(n) <format> format
+        And the template does not exist
+        When a project with the <permission> permissions publishes the template
+        Then the template <stored> stored
+
+        Examples:
+          | format  | permission  | stored
+          | valid   | right       | is
+          | valid   | wrong       | is not
+          | invalid | right       | is not
+          | invalid | wrong       | is not


features/templates.feature
```diff
@@ -1,7 +1,42 @@
...
+            main:
+                image: node:6
+                steps:
+                    - install: npm install
+                    - test: npm test
+
+    Rules:
+        - Template owner can publish a template if
+              - The format of template is valid
+              - The build that is used to publish has good permission
```
tkyi: `The build that is used to publish has the right permissions`

features/templates.feature
```diff
@@ -12,3 +47,36 @@ Feature: Templates
...
           |  valid   |   no   |
           |  invalid |  some  |
+
+    @ignore
+    Scenario Outline: Merges template into user's config
+        Given an existing pipeline with job main
+        And an existing template called Test
+        And user only specifies template Test in job main
+        When user starts the pipeline
+        Then job main executes what is specified in <template>
```
tkyi: `template`

features/templates.feature
```diff
@@ -12,3 +47,36 @@ Feature: Templates
...
+          And user specifies template Test in job main
+          And template Test has environment {TEST: FOO}
+          And job main has environment {TEST: BAR}
+          When user starts the pipeline
+          Then TEST=BAR
+
+    @ignore
+    Scenario Outline: Publish template saves template into Datastore
+        Given a template with a(n) <format> format
+        And it has <permission> permission to publish
```
tkyi: maybe this?
```
And it has the <permission> permission to publish

| permission |
| right |
| wrong |
| right |
| wrong |
```

features/templates.feature
```diff
@@ -1,7 +1,42 @@
...
+                steps:
+                    - install: npm install
+                    - test: npm test
+
+    Rules:
+        - Template owner can publish a template if
+              - The format of template is valid
+              - The build that is used to publish has the right permissions
+              - No existing version already exists
+        - Publishing a template should store the template into Datastore.
```
FenrirUnbound: > Publishing a template should store the template into Datastore

Saying that it stores into the Datastore is a bit too implementation-specific. Is there a scenario that checks the contents of the Datastore?

Another way of stating this is "Publishing a template should be stored to be retrieved later", or something along those lines. Then the scenario can be relaxed and only needs to retrieve the template at a future time.

features/templates.feature
```diff
@@ -12,3 +47,36 @@ Feature: Templates
           | template | number |
           |  valid   |   no   |
           |  invalid |  some  |
+
+    @ignore
+    Scenario Outline: Merges template into user's config
+        Given an existing pipeline with job main
+        And an existing template called Test
+        And user only specifies template Test in job main
```
FenrirUnbound: You may want this to match https://github.com/screwdriver-cd/screwdriver/pull/496/files#diff-2d113dc19bf2d9aa7571e6c04e00dfeaR63 for reusability

features/templates.feature
```diff
@@ -12,3 +47,36 @@ Feature: Templates
...
+        When user starts the pipeline
+        Then job main executes what is specified in template
+
+      @ignore
+      Scenario Outline: Job config takes precedence over template config
+          Given an existing pipeline with job main
+          And an existing template called Test
+          And user specifies template Test in job main
+          And template Test has environment {TEST: FOO}
```
FenrirUnbound: This is really implementation-specific. What about other user-defined settings? It's not described elsewhere, so what happens if the user specifies an `image`?

It may be better to describe this scenario more generally, and say that "a user has some settings defined" and verify that "those settings were preserved/persisted".

features/templates.feature
```diff
@@ -12,3 +47,36 @@ Feature: Templates
...
+          When user starts the pipeline
+          Then settings is the job settings
+
+    @ignore
+    Scenario Outline: Publish template saves template into Datastore
+        Given a template with a(n) <format> format
+        And it has the <permission> permissions to publish
+        And the template does not exist
+        When template owner runs the build to publish the template
+        Then the template is stored into Datastore
```
tkyi:
```
Then the template <stored> stored into Datastore

| stored |
| is |
| is not |
| is not |
| is not |
```

d2lam: Addressed

features/templates.feature
```diff
@@ -12,3 +47,36 @@ Feature: Templates
...
+          And user has some settings defined
+          And template Test has the same settings with different values
+          When user starts the pipeline
+          Then settings is the job settings
+
+    @ignore
+    Scenario Outline: Publish template saves template into Datastore
+        Given a template with a(n) <format> format
+        And it has the <permission> permissions to publish
+        And the template does not exist
```
FenrirUnbound: > And the template does not exist

Is this referring to a completely brand new template, or a template+version does not exist?

features/templates.feature
```diff
@@ -12,3 +47,36 @@ Feature: Templates
...
+        When template owner runs the build to publish the template
```
FenrirUnbound: This is slightly conflicts with the line "*And it has the <permission> permissions to publish*". A template owner should be able to publish, since they own the template. 

One suggestion I would offer is to combine the statements to something like "When a project with <permission> permissions publishes the template"

features/templates.feature
d2lam: Either. Do I need to make it clearer?

features/templates.feature
```diff
@@ -1,7 +1,42 @@
...
+                image: node:6
+                steps:
+                    - install: npm install
+                    - test: npm test
+
+    Rules:
+        - Template owner can publish a template if
+              - The format of template is valid
+              - The build that is used to publish has the right permissions
+              - No existing version already exists
```
minz1027: We will auto bump the version and ignore the patch version, so this will not happen

features/templates.feature
```diff
@@ -12,3 +47,35 @@ Feature: Templates
...
+        And user specifies template Test in job main
+        And user has some settings defined
+        And template Test has the same settings with different values
+        When user starts the pipeline
+        Then settings is the job settings
+
+    @ignore
+    Scenario Outline: Publish template stores the template
+        Given a template with a(n) <format> format
+        And the template does not exist
```
minz1027: same as the previous comment

features/templates.feature
FenrirUnbound: If that's the case, it makes this a valid rule

d2lam: Yeah, actually, the feature file defines the rules, regardless of the implementation details.

PR #497, opened 2017-03-14
fix(470): autobump version

d2lam:
Auto bump version:
For example, if latest version is 1.7.3

If template.yaml says 1, it will create 1.7.4
If template.yaml says 1.7, it will create 1.7.4
If template.yaml says 1.7.3, it will create 1.7.4 (does not matter what the patch is)
If template.yaml says 1.8, it will create 1.8.0
If template.yaml says 2, it will create 2.0.0

Blocked by https://github.com/screwdriver-cd/models/pull/151
https://github.com/screwdriver-cd/models/pull/152

### DIFF
--- a/plugins/templates/create.js
+++ b/plugins/templates/create.js
@@ -4,9 +4,6 @@ const boom = require('boom');
 const schema = require('screwdriver-data-schema');
 const hoek = require('hoek');
 const urlLib = require('url');
-const MAJOR_MATCH = 1;
-const MINOR_MATCH = 2;
-const PATCH_MATCH = 3;
 
 module.exports = () => ({
     method: 'POST',
@@ -35,16 +32,9 @@ module.exports = () => ({
                 pipelineFactory.get(pipelineId),
                 templateFactory.list({ name })
             ]).then(([pipeline, templates]) => {
-                const VERSION_REGEX = schema.config.regex.VERSION;
-                const match = VERSION_REGEX.exec(request.payload.version);
-                const major = match[MAJOR_MATCH];
-                const minor = match[MINOR_MATCH];
-                const version = minor ? `${major}${minor}.0` : `${major}.0.0`;
-
                 const templateConfig = hoek.applyToDefaults(request.payload, {
                     scmUri: pipeline.scmUri,
-                    labels,
-                    version
+                    labels
                 });
 
                 // If template name doesn't exist yet, just create a new entry
@@ -58,28 +48,9 @@ module.exports = () => ({
                     throw boom.unauthorized('Not allowed to publish this template');
                 }
 
-                // If template name exists and has good permission, get the latest version for major and minor
-                const searchVersion = minor ? `${major}${minor}` : major;
-
-                return templateFactory.getTemplate({
-                    name,
-                    version: searchVersion,
-                    labels
-                }).then((latest) => {
-                    if (!latest) {
-                        templateConfig.version = version;
-                    } else {
-                        const latestMatch = VERSION_REGEX.exec(latest.version);
-                        const latestMinor = latestMatch[MINOR_MATCH];
-                        const patch = parseInt(
-                            latestMatch[PATCH_MATCH].slice(1), 10) + 1;
-
-                        templateConfig.version = `${major}${latestMinor}.${patch}`;
-                    }
-
-                    // If the exact version doesn't exists, create a new entry
-                    return templateFactory.create(templateConfig);
-                });
+                // If template name exists and has good permission, then create
+                // Create would automatically bump the patch version
+                return templateFactory.create(templateConfig);
             })
             .then((template) => {
                 const location = urlLib.format({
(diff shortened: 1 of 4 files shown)


FenrirUnbound:
> We should only ship new versions if the template.yaml changes... Templates are just static content, there's no reason to rebuild

Ok, that makes a lot of sense. I was thinking along the lines of a shared step like `npm_install@1`. Assuming that it will work similar to how we intend templates to function, I can totally see this format working.

> We should be using the same learnings we got from Node.js and users should never have to bump their own patch version if they are changing something. That would make PRs impossible to merge one after another as each one would create conflicts with the next one.

Making the patch version automatic solves a lot of other edge cases, including multiple PRs to amend a template file. 

A semver-related question then: does this mean that any change to a template would either be a major change or a patch change? Phrased another way: what does a minor version bump look like?

d2lam:
@stjohnjohnson  @FenrirUnbound  We implemented auto bump. 
For example, if latest version is `1.7.3`
If `template.yaml` says `1`, it will create `1.7.4`
If `template.yaml` says `1.7`, it will create `1.7.4`
If `template.yaml` says `1.8`, it will create `1.8.0`
If `template.yaml` says `2`, it will create `2.0.0`

plugins/templates/create.js
```diff
@@ -26,18 +29,22 @@ module.exports = () => ({
...
 
             return Promise.all([
                 pipelineFactory.get(pipelineId),
                 templateFactory.list({ name })
             ]).then(([pipeline, templates]) => {
+                const VERSION_REGEX = schema.config.regex.VERSION;
+                const match = VERSION_REGEX.exec(request.payload.version);
+                const major = match[MAJOR_MATCH];
+                const minor = match[MINOR_MATCH];
+                const version = minor ? `${major}${minor}.0` : `${major}.0.0`;
```
FenrirUnbound: It looks like the Regex match is wrong. Why isn't it `${major}.${minor}.0` ?

plugins/templates/create.js
```diff
@@ -51,17 +58,28 @@ module.exports = () => ({
...
 
-                // If template name exists and has good permission, check the exact version
-                return templateFactory.get({ name, version })
-                    .then((template) => {
-                        // If the exact version exists, throw 409
-                        if (template) {
-                            throw boom.conflict(`Template ${name}@${version} already exists`);
-                        }
+                // If template name exists and has good permission, get the latest version for major and minor
+                const searchVersion = minor ? `${major}${minor}` : major;
```
FenrirUnbound: `version` could be used instead of this.

plugins/templates/create.js
d2lam: second group includes the `.`: https://github.com/screwdriver-cd/data-schema/blob/master/config/regex.js#L11

FenrirUnbound: The regex should return you just the minor version

plugins/templates/create.js
d2lam: `searchVersion` is used to get the latest of major & minor, it shouldn't include the patch. 
For example,
payload.version = 1.2.0
version = 1.2.0
searchVersion = 1.2

FenrirUnbound: Does it not default to a 0 for the patch?

FenrirUnbound:
This is a lot of logic in a Controller. Why have it all handled here instead of in a Model?

plugins/templates/create.js
d2lam: with `1.2`, `getTemplate`returns `1.2.latestPatch`, 
with `1.2.0`, `getTemplate` returns the exact version `1.2.0`

plugins/templates/create.js
d2lam: I'm not too good with regex, can you suggest a regex that would give that?

d2lam:
@FenrirUnbound  That's possible. You mean to just have a helper inside templateFactory to get the correct version to create?

plugins/templates/create.js
FenrirUnbound: It just looks like a bug on the Regex matcher. I would have expected it to match *only* the version number and not include the delimiter

FenrirUnbound:
> That's possible. You mean to just have a helper inside templateFactory to get the correct version to create?

Yeah. Some of the checks make sense to make here, like the permissions lookup. The business logic should be placed in the Models. A lot of the version patching and compiling could leverage the `semver` module.

d2lam:
Yeah that makes sense. I will move some logic to models

FenrirUnbound:
>  I will move some logic to models

👍  A lot of my concerns will vanish when the `semver` module is utilized. All of the version number handling goes away, and the logic will focus solely on the edge cases you're trying to capture

test/plugins/templates.test.js
```diff
@@ -245,15 +237,6 @@ describe('template plugin test', () => {
             });
         });
 
-        it('returns 409 it is an existing version', () => {
```
d2lam: Nope, since `create` will always bump version now.

PR #499, opened 2017-03-15
feat: Add cucumber test for metadata

wahapo:
Add cucumber test for metadata.feature.

Repository for testing is [here](https://github.com/wahapo/functional-metadata). I hope it will be relocated to screwdriver-cd-test Organization.

### DIFF
--- a/features/metadata.feature
+++ b/features/metadata.feature
@@ -35,7 +35,6 @@ Feature: Metadata
     Background:
         Given a metadata starts with an empty object
 
-    @ignore
     Scenario Outline: Adding some data to metadata
         Given an existing pipeline with the workflow:
             | job | triggers |
@@ -60,7 +59,6 @@ Feature: Metadata
             | ["arrg"]     | ["ARRG"]     |
             | { "x": "y" } | { "w": "z" } |
     
-    @ignore
     Scenario: Store a record of the metadata when the event is done
         Given an existing pipeline with the workflow:
             | job | triggers |
--- a/features/metadata.feature
+++ b/features/metadata.feature
@@ -37,11 +37,11 @@ Feature: Metadata
 
     Scenario Outline: Adding some data to metadata
         Given an existing pipeline with the workflow:
-            | job | triggers |
-            | FOO | BAR      |
-            | BAR | BAZ      |
-        When the "FOO" job is started
-        Then add the { "foo": <foobar> } to metadata in the "FOO" build container
+            | job  | triggers |
+            | main | BAR      |
+            | BAR  | BAZ      |
+        When the "main" job is started
+        Then add the { "foo": <foobar> } to metadata in the "main" build container
         And the build succeeded
         And the "BAR" job is started
         Then in the build, the { "foo": <foobar> } is available from metadata
@@ -50,6 +50,7 @@ Feature: Metadata
         And the "BAZ" job is started
         Then in the build, the { "foo": <foobar> } is available from metadata
         And in the build, the { "bar": <barbaz> } is available from metadata
+        And the build succeeded
 
         Examples:
             | foobar       | barbaz       |
@@ -61,10 +62,10 @@ Feature: Metadata
     
     Scenario: Store a record of the metadata when the event is done
         Given an existing pipeline with the workflow:
-            | job | triggers |
-            | FOO | BAR      |
-        When the "FOO" job is started
-        Then add the { "foo": "foobar" } to metadata in the "FOO" build
+            | job  | triggers |
+            | main | BAR      |
+        When the "main" job is started
+        Then add the { "foo": "foobar" } to metadata in the "main" build
         And the build succeeded
         And the "BAR" job is started
         And add the { "bar": "barbaz" } to metadata in the "BAR" build
@@ -74,12 +75,12 @@ Feature: Metadata
     @ignore
     Scenario: Combining the results of matrix builds
         Given an existing pipeline with the workflow:
-            | job | triggers |
-            | FOO | BAR      |
-        And "FOO" is a matrix job with BAZ:[1,2]
-        When the "FOO" job is started
-        Then add the { "foo": "foobar" } to metadata in the BAZ = 1, "FOO" build
-        And add the { "foo": "barbaz" } to metadata in the BAZ = 2, "FOO" build
+            | job  | triggers |
+            | main | BAR      |
+        And "main" is a matrix job with BAZ:[1,2]
+        When the "main" job is started
+        Then add the { "foo": "foobar" } to metadata in the BAZ = 1, "main" build
+        And add the { "foo": "barbaz" } to metadata in the BAZ = 2, "main" build
         And the build succeeded
         And the "BAR" job is started
         Then in the build, the value of "foo" is either "foobar" or "barbaz" from metadata
--- a/features/metadata.feature
+++ b/features/metadata.feature
@@ -51,27 +51,17 @@ Feature: Metadata
         Then in the build, the { "foo": <foobar> } is available from metadata
         And in the build, the { "bar": <barbaz> } is available from metadata
         And the build succeeded
+        And the event is done
+        Then a record of the metadata is stored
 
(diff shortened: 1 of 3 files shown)


tkyi:
Hey @wahapo ,

I created the functional-metadata repo for you in screwdriver-cd-test and added you as an Admin.
https://github.com/screwdriver-cd-test/functional-metadata

features/metadata.feature
```diff
@@ -35,7 +35,6 @@ Feature: Metadata
     Background:
         Given a metadata starts with an empty object
 
-    @ignore
     Scenario Outline: Adding some data to metadata
```
minz1027: `Adding some data to metadata` and `Store a record of the metadata when the event is done` look very similar. They are all checking if the meta is set in the subsequent builds. Can we combine them somehow?

features/step_definitions/metadata.js
```diff
@@ -0,0 +1,157 @@
...
+        this.repoName = 'functional-metadata';
+        this.pipelineId = null;
+        this.eventId = null;
+        this.meta = null;
+        this.jwt = null;
+    });
+
+    this.Given(/^a metadata starts with an empty object$/);
+
+    this.Given(/^an existing pipeline with the workflow:$/, { timeout: TIMEOUT }, table =>
```
d2lam: This is already here: https://github.com/screwdriver-cd/screwdriver/blob/master/features/step_definitions/events.js#L21-L53

It will match already.

features/step_definitions/metadata.js
```diff
@@ -0,0 +1,157 @@
...
+                const id = str.split(': ')[1];
+
+                this.pipelineId = id;
+            }
+
+            return table;
+        })
+    );
+
+    this.When(/^the "FOO" job is started$/, { timeout: TIMEOUT }, () =>
```
d2lam: I think we can reuse this somehow: https://github.com/screwdriver-cd/screwdriver/blob/1b6727f610bcb49f1f94dac2ca1f792e33c5fa2c/features/step_definitions/secrets.js#L74-L103

Might need some changes to the feature file.

features/metadata.feature
wahapo: A good idea did not come to mind. I think if we attempt to somehow resolve, for example, create dedicated repository `functional-metadata-1` for testing for `meta set "foo" "Foo"`, and create `functional-metadata-2` for testing for `meta set "foo" 10` in the same way. Is there any other best idea?

features/step_definitions/metadata.js
wahapo: fixed. thanks!

features/step_definitions/metadata.js
wahapo: Modified to reuse pointed out, and also fixed that corresponding part of the feture file.

features/step_definitions/metadata.js
```diff
@@ -0,0 +1,95 @@
...
+        this.repoName = 'functional-metadata';
+        this.pipelineId = null;
+        this.eventId = null;
+        this.meta = null;
+        this.jwt = null;
+    });
+
+    this.Given(/^a metadata starts with an empty object$/);
+
+    this.Then(/^the "BAR" job is started$/, { timeout: TIMEOUT }, () =>
```
tk3fftk: `the "BAR" job is started` case and `the "BAZ" job is started` case are looks similar.
Maybe these can be changed like:
https://github.com/screwdriver-cd/screwdriver/blob/master/features/step_definitions/templates.js#L9

wahapo: fixed

features/metadata.feature
d2lam: I think @minz1027  meant the case `Store a record of the metadata when the event is done` seems duplicate. Maybe we can remove it/combine it with the other scenario.

wahapo: Oh, I understood. Fix to combine `Adding some data to metadata` and `Store a record of the metadata when the event is done`. Like this?

features/metadata.feature
```diff
@@ -51,37 +50,27 @@ Feature: Metadata
...
         Then in the build, the { "foo": <foobar> } is available from metadata
         And in the build, the { "bar": <barbaz> } is available from metadata
+        And the build succeeded
+        And the event is done
+        Then a record of the metadata is stored
 
         Examples:
             | foobar       | barbaz       |
-            | "Foo"        | "Bar"        |
+            | "foobar"     | "barbaz"     |
```
wahapo: Is there a problem with changing to "foobar" from "FOO". In [functional-metadata](https://github.com/screwdriver-cd-test/functional-metadata/commit/6e2f1dfee80e94c20de9fc6b294dfd2027909542) I described it as "foobar".

features/step_definitions/secrets.js
```diff
@@ -79,6 +79,8 @@ module.exports = function server() {
         }).then((response) => {
             this.jobId = response.body[0].id;
             this.secondJobId = response.body[1].id;
+            if (2 in response.body) this.thirdJobId = response.body[2].id;
+            this.lastJobId = response.body[response.body.length - 1].id;
```
wahapo: `lastJobId` is used for `the event is done`. The end of the event is synonymous with the end of the last job.

features/metadata.feature
d2lam: Yes! :)

features/metadata.feature
d2lam: Should be fine!

features/step_definitions/secrets.js
d2lam: When is `this.lastJobId` used? I don't see it.

features/step_definitions/secrets.js
```diff
@@ -79,6 +79,8 @@ module.exports = function server() {
         }).then((response) => {
             this.jobId = response.body[0].id;
             this.secondJobId = response.body[1].id;
+            if (2 in response.body) this.thirdJobId = response.body[2].id;
```
petey: Is response.body an array? That seems weird to me. 

You definitely shouldn't be using the `in` operator here, it doesn't always do what you expect it to do.
For example:
```
const a = ['a', 'b', undefined];
console.log(2 in a); // true
```

Instead, look for what you expect to find.
```
this.thirdJobId = typeof response.body[2] === 'object' ? response.body[2].id : null;
```

Also, always use brackets with if statements.

d2lam: Yeah, this is the LIST jobs endpoint so `response.body` is an array of jobs. 

I agree with Petey on not using `in` here.

features/step_definitions/secrets.js
wahapo: It the correction was missing. It is used by `the event is done`.

features/step_definitions/secrets.js
```diff
@@ -79,6 +79,16 @@ module.exports = function server() {
         }).then((response) => {
             this.jobId = response.body[0].id;
             this.secondJobId = response.body[1].id;
+            this.thirdJobId = typeof response.body[2] === 'object' ? response.body[2].id : null;
+            this.lastJobId = (function getLastJobId() {
```
petey: I really don't think you needed to change this, but this is a one-liner that should be equivalent.

```
response.body.reverse().find(b => typeof b === 'object').id;
```

features/step_definitions/secrets.js
wahapo: Modify not to use `in`.

features/step_definitions/secrets.js
wahapo: It's amazing! I adopted it.

d2lam:
Ah we should have squashed first before merging.

d2lam:
@wahapo  This is not working for some reason. I opened a PR to ignore it: https://github.com/screwdriver-cd/screwdriver/pull/502

wahapo:
Sorry, I tried modifying it. https://github.com/screwdriver-cd/screwdriver/pull/503

PR #503, opened 2017-03-21
feat: Modify metadata cucumber

wahapo:
relay: https://github.com/screwdriver-cd/screwdriver/pull/499

Perhaps it was because timeout was required for `a metadata starts with an empty object`. I tried to fix it, the part of the problem seems to work.
```
  @metadata
  Scenario: Adding some data to metadata
  ✔ Given a metadata starts with an empty object
```

In addition, the table of `Example` was automatically expanded, and it turned out that the step was undefined. In order to fix this, I think that we need to modify matching pattern and remove some from the `Examples`.
```
  - When the "main" job is started
  ? Then add the { "foo": "foobar" } to metadata in the "main" build container
  - And the build succeeded
  - And the "BAR" job is started
  ? Then in the build, the { "foo": "foobar" } is available from metadata
  ? And add the { "bar": "barbaz" } to metadata in the "BAR" build container
  - And the build succeeded
  - And the "BAZ" job is started
  ? Then in the build, the { "foo": "foobar" } is available from metadata
  ? And in the build, the { "bar": "barbaz" } is available from metadata
  - And the build succeeded
  - And the event is done
  - Then a record of the metadata is stored

  - When the "main" job is started
  ? Then add the { "foo": 10 } to metadata in the "main" build container
  - And the build succeeded
  - And the "BAR" job is started
  ? Then in the build, the { "foo": 10 } is available from metadata
  ? And add the { "bar": 20 } to metadata in the "BAR" build container
  - And the build succeeded
  - And the "BAZ" job is started
  ? Then in the build, the { "foo": 10 } is available from metadata
  ? And in the build, the { "bar": 20 } is available from metadata
  - And the build succeeded
  - And the event is done
  - Then a record of the metadata is stored
```

@tk3fftk said that he would try to set PATH of the meta command, to be able to use even without `/opt/sd/`. [This PR](https://github.com/screwdriver-cd/launcher/pull/111) is it. If this is not merged, the metadata test job will fail because the `meta` not found in [this part](https://github.com/screwdriver-cd-test/functional-metadata/blob/master/screwdriver.yaml).

### DIFF
--- a/features/step_definitions/metadata.js
+++ b/features/step_definitions/metadata.js
@@ -17,7 +17,7 @@ module.exports = function server() {
         this.jwt = null;
     });
 
-    this.Given(/^a metadata starts with an empty object$/);
+    this.Given(/^a metadata starts with an empty object$/, { timeout: TIMEOUT }, () => null);
 
     this.Then(/^the "(BAR|BAZ)" job is started$/, { timeout: TIMEOUT }, (jobName) => {
         let jobId = '';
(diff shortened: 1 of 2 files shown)


d2lam:
It's weird that the unit test failed

d2lam:
Actually it's not specific to this PR. I cloned your repo, and it seems to be failing on `master`. However, it passes on `screwdriver:master`. I'm very confused. Is `wahapo:master` different from `screwdriver:master`?

wahapo:
Became clear why the test failed, so I put it in https://github.com/screwdriver-cd/screwdriver/pull/504. It was influenced by the dependency library `boom v4.3.0`. Please review  https://github.com/screwdriver-cd/screwdriver/pull/504 for a fix for test failures.

features/step_definitions/metadata.js
```diff
@@ -17,7 +17,7 @@ module.exports = function server() {
         this.jwt = null;
     });
 
-    this.Given(/^a metadata starts with an empty object$/);
+    this.Given(/^a metadata starts with an empty object$/, { timeout: TIMEOUT }, () => null);
```
minz1027: Hmm.. why we need timeout here? This step does not look like it's doing anything.

PR #504, opened 2017-03-21
feat: Modify validator/template test

wahapo:
I Modified it because test has getting failed. Does this modification would be the right decision?
Test failed reason is,
```
  1) template validator plugin test POST /validator/template returns BAD REQUEST for template that cannot be parsed:
     AssertError: expected value to match
    expected = /YAMLException/
    actual = incomplete explicit mapping pair; a key node is missed at line 1, column 8:
    error: :
```

And modified a lint warning.
```
/Users/kkawahar/screwdriver/features/step_definitions/templates.js
  8:18  warning  Unexpected unnamed function  func-names

✖ 1 problem (0 errors, 1 warning)
```

### DIFF
--- a/features/step_definitions/templates.js
+++ b/features/step_definitions/templates.js
@@ -5,7 +5,7 @@ const request = require('../support/request');
 const path = require('path');
 const Assert = require('chai').assert;
 
-module.exports = function () {
+module.exports = function server() {
     this.Given(/^a (valid|invalid)\b job-level template$/, (templateType) => {
         let targetFile = '';
 
--- a/plugins/template-validator.js
+++ b/plugins/template-validator.js
@@ -26,7 +26,7 @@ exports.register = (server, options, next) => {
                 const templateString = request.payload.yaml;
 
                 return validator(templateString)
-                .then(reply, err => reply(boom.badRequest(err)));
+                .then(reply, err => reply(boom.badRequest(err.toString())));
             },
             validate: {
                 payload: templateSchema.input
--- a/test/plugins/validator/template.test.js
+++ b/test/plugins/validator/template.test.js
@@ -132,7 +132,7 @@ describe('template validator plugin test', () => {
 
                 const payload = JSON.parse(reply.payload);
 
-                assert.match(payload.message, /YAMLException/);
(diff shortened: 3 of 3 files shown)


d2lam:
It's weird that the test failed for https://github.com/screwdriver-cd/screwdriver/pull/503. There is no change to the test or any code that should affect the unit test. 

It passed everywhere else (on master, and even on this PR). We should investigate why that happens...

features/step_definitions/templates.js
```diff
@@ -5,7 +5,7 @@ const request = require('../support/request');
 const path = require('path');
 const Assert = require('chai').assert;
 
-module.exports = function () {
+module.exports = function server() {
```
d2lam: This can just be `module.exports = () => {`

Same as other files. I'm not sure why we had it as `function server() {` in the first place

wahapo:
The reason for failing the test was that boom logic was fixed. Previously, `toString()` was implicitly called [here](https://github.com/hapijs/boom/pull/139/files#diff-6d186b954a58d5bb740f73d84fe39073L88), but in boom v4.3.0 or later it is necessary to use a character string in message as the argument of `badRequest(message)`. By using `toString()`, `YAMLException` can appear in `message`.
Please tell me if it is better to lock with boom v4.2.0.

features/step_definitions/templates.js
wahapo: `module.exports = () => {` can not be used here. When used, I think that `this` will be undefined because `this` does not exist in this scope. [References is here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions)
Also, if omit the function name, then it will get a warning with eslint. It is no problem, but I think that it is better to deal with this. [References is here](http://eslint.org/docs/rules/func-names)

d2lam:
Nice find, @wahapo! I thought it was a dependency problem too. I was thinking it's `js-yaml`, but saw there was no change.

features/step_definitions/templates.js
d2lam: I see

PR #508, opened 2017-03-23
feat: Fix not to capture unused string

wahapo:
I found that it need to give arguments when capturing with regular expressions.
https://cd.screwdriver.cd/pipelines/1/build/2781

So, I modified to not capture. By modifying this way, it seems that there is no problem without specifying arguments.

### DIFF
--- a/features/step_definitions/metadata.js
+++ b/features/step_definitions/metadata.js
@@ -33,13 +33,12 @@ module.exports = function server() {
             throw new Error('jobName is neither BAR or BAZ');
         }
 
-        request({
+        return request({
             uri: `${this.instance}/${this.namespace}/jobs/${jobId}/builds`,
             method: 'GET',
             json: true
         }).then((response) => {
             this.buildId = response.body[0].id;
-            this.meta = response.body.meta;
 
             Assert.equal(response.statusCode, 200);
         });
@@ -49,15 +48,9 @@ module.exports = function server() {
 
     this.Then(/^add the { "bar": (?:.*) } to metadata in the "BAR" build container$/, () => null);
 
-    this.Then(/^in the build, the { "foo": "(.*)" } is available from metadata$/, (value) => {
-        Assert.ok('foo', this.meta);
-        Assert.equal(value, this.meta.foo);
-    });
+    this.Then(/^in the build, the { "foo": "(?:.*)" } is available from metadata$/, () => null);
 
-    this.Then(/^in the build, the { "bar": "(.*)" } is available from metadata$/, (value) => {
-        Assert.ok('bar', this.meta);
-        Assert.equal(value, this.meta.bar);
-    });
+    this.Then(/^in the build, the { "bar": "(?:.*)" } is available from metadata$/, () => null);
 
     this.Then(/^the build succeeded$/, { timeout: TIMEOUT }, () =>
         this.waitForBuild(this.buildId).then((resp) => {
@@ -73,6 +66,7 @@ module.exports = function server() {
             json: true
         }).then((response) => {
             this.buildId = response.body[0].id;
+            this.meta = response.body[0].meta;
 
             return this.waitForBuild(this.buildId).then((resp) => {
                 Assert.equal(resp.statusCode, 200);
--- a/features/step_definitions/metadata.js
+++ b/features/step_definitions/metadata.js
@@ -44,13 +44,12 @@ module.exports = function server() {
         });
     });
 
-    this.Then(/^add the { "foo": (?:.*) } to metadata in the "main" build container$/, () => null);
+    this.Then(/^add the { "(.*)": "(.*)" } to metadata/, (key, value) => {
+        this.expectedMeta = this.expectedMeta || {};
+        this.expectedMeta[key] = value;
+    });
 
-    this.Then(/^add the { "bar": (?:.*) } to metadata in the "BAR" build container$/, () => null);
-
-    this.Then(/^in the build, the { "foo": "(?:.*)" } is available from metadata$/, () => null);
-
-    this.Then(/^in the build, the { "bar": "(?:.*)" } is available from metadata$/, () => null);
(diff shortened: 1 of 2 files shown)


d2lam:
@wahapo  I will run this locally tomorrow to see if this fixes it.

d2lam:
@wahapo Did you try running it locally? I tried running and it got passed this step but failed at `And the build succeeded`

wahapo:
@d2lam I'm sorry I did not been able proper test. Since it is forbidden to run docker on my Mac, I will try building an alternative environment so that we can test it. Please give me a time.

features/step_definitions/metadata.js
```diff
@@ -33,31 +33,24 @@ module.exports = function server() {
             throw new Error('jobName is neither BAR or BAZ');
         }
 
-        request({
+        return request({
```
wahapo: When I tried it, it seems necessary to add `return` to get `this.buildId`.

features/step_definitions/metadata.js
```diff
@@ -33,31 +33,24 @@ module.exports = function server() {
...
     });
 
-    this.Then(/^add the { "foo": (.*) } to metadata in the "main" build container$/, () => null);
+    this.Then(/^add the { "foo": (?:.*) } to metadata in the "main" build container$/, () => null);
 
-    this.Then(/^add the { "bar": (.*) } to metadata in the "BAR" build container$/, () => null);
+    this.Then(/^add the { "bar": (?:.*) } to metadata in the "BAR" build container$/, () => null);
 
-    this.Then(/^in the build, the { "foo": "(.*)" } is available from metadata$/, (value) => {
-        Assert.ok('foo', this.meta);
```
wahapo: It seems that `meta` can not be obtained from the API while the status is RUNNING. So, we can't testing in this method that is running a build.

wahapo:
I think it will work if the correction of https://github.com/screwdriver-cd-test/functional-metadata/pull/1 works

features/step_definitions/metadata.js
```diff
@@ -33,31 +33,24 @@ module.exports = function server() {
...
         }).then((response) => {
             this.buildId = response.body[0].id;
-            this.meta = response.body.meta;
 
             Assert.equal(response.statusCode, 200);
         });
     });
 
-    this.Then(/^add the { "foo": (.*) } to metadata in the "main" build container$/, () => null);
+    this.Then(/^add the { "foo": (?:.*) } to metadata in the "main" build container$/, () => null);
```
petey: I don't think that the `?:` adds anything to this regular expression.

d2lam:
@wahapo  I commented on the other PR. It shouldn't need `/opt/sd/meta`, `meta` by itself should work.

features/step_definitions/metadata.js
wahapo: If `?:` is not used, we have to make coding like this. It is mandatory to give an argument `value`.
```
this.Then(/^add the { "foo": (.*) } to metadata in the "main" build container$/, (value) => {
  // Must do something with ${value}
});
```
There is no way to do a good test using "value", so I think that "value" can be discarded.

--
Or simply, such coding will work. Is this better?
```
this.Then(/^add the { "foo": "foobar" } to metadata in the "main" build container$/, () => null);
 ```

Comment (wahapo) [inline features/step_definitions/metadata.js]: @petey How about this? Do we need to fix anything?

**Comment by petey** [comment]
I will take another look.

File: features/step_definitions/metadata.js (hunk 3 above)
Comment (petey) [inline features/step_definitions/metadata.js]: I think what I would expect from a step definition like this would be that this would set up expected values you would check later.

For example:
```
this.Then(/^add the { "(.*)": "(.*)" } to metadata in the "main" build container$/, (key, value) => {
   this.expectedMeta = this.expectedMeta || {};
   this.expectedMeta[key] = value;
});
```

Then in `a record of the metadata is stored`:
```
Object.keys(this.expectedMeta).forEach((key) => {
  assert.equal(this.meta[key], this.expectedMeta[key]);
});
```

Comment (wahapo) [inline features/step_definitions/metadata.js]: Ohhh I see, I adopted it because it is a very good idea.

File: features/step_definitions/metadata.js [hunk 4]
```diff
@@ -81,9 +74,9 @@ module.exports = function server() {
     );
 
     this.Then(/^a record of the metadata is stored$/, { timeout: TIMEOUT }, () => {
-        Assert.ok('foo', this.meta);
-        Assert.ok('bar', this.meta);
-        Assert.equal('foobar', this.meta.foo);
-        Assert.equal('barbaz', this.meta.bar);
+        Object.keys(this.expectedMeta).forEach((key) => {
+            Assert.ok(key, this.meta);
```
petey: I do not think this is doing what you think it is doing. http://chaijs.com/api/assert/#method_isok

For one, chai.assert does not have a documented `ok` method. This is borrowed from `expect` so it still works (for now). The documented api for chai.assert is `Assert.isOk(value, message)` where it determines if `value` is truthy. In this loop, `key` is always truthy. All you are doing here is verifying that the keys of expectedMeta are defined.

What I suspect you are actually trying to to is `Assert.property(this.meta, key)` to determine if the actual meta object has a property with the key from the value of `key`. 

This really isn't necessary either, since the next line asserts that there is a value at that key in actual meta that matches an expected value with the same key. If the key didn't exist, the value would be `undefined`, and that will not match the expected value.

wahapo: Right, got it. Even without using`Assert.property`, I checked that I was expecting the test. Since this assert is not necessary, I deleted it.

wahapo:
@d2lam I think that there is no problem anymore. On your business day, please check if there is no other concern. Thanks.

d2lam:
Hey @wahapo. I have been looking, but the screwdriver pipeline was failing. Just got it to pass after work today. I can now merge this.

wahapo:
thanks a lot❤️

PR #509, opened 2017-03-24
fix: datastore scan requires params key

d2lam:
It looks through `config.params` in datastore scan:
https://github.com/screwdriver-cd/datastore-sequelize/blob/master/index.js#L319-L320

### DIFF
--- a/features/metadata.feature
+++ b/features/metadata.feature
@@ -1,4 +1,4 @@
-@metadata
+@ignore
 Feature: Metadata
 
     Users want to pass structured data between their builds.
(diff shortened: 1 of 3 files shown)


features/metadata.feature
```diff
@@ -1,4 +1,4 @@
-@metadata
+@ignore
```
tkyi: no metadata tests?

d2lam: it's blocking the pipeline: https://cd.screwdriver.cd/pipelines/1
We are still trying to fix it here: https://github.com/screwdriver-cd/screwdriver/pull/508

PR #520, opened 2017-04-12
fix: POST/templates takes in yaml string instead of config

d2lam:
`POST /templates` is currently taking in the screwdriver config, and was validated based on `schema.models.template.create`: https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/templates/create.js#L69

This PR changes the payload to be the same as [`POST /validator/template`](https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/template-validator.js). It takes in yaml string instead of the config. 

So payload looks like:
```
{"yaml":"{\"name\":\"template_namespace/nodejs_main\",\"version\":\"1.2\",\"description\":\"Template for building a NodeJS module\\nInstalls dependencies and runs tests\\n\",\"maintainer\":\"me@nowhere.com\",\"config\":{\"image\":\"node:6\",\"steps\":[{\"install\":\"npm install\"},{\"test\":\"npm test\"}],\"environment\":{\"KEYNAME\":\"value\"},\"secrets\":[\"NPM_TOKEN\"]}}"}
```

It also calls validator before creating the template, and returns 400 if template is invalid. 

### DIFF
--- a/plugins/templates/create.js
+++ b/plugins/templates/create.js
@@ -2,6 +2,8 @@
 
 const boom = require('boom');
 const schema = require('screwdriver-data-schema');
+const validator = require('screwdriver-template-validator');
+const templateSchema = schema.api.templateValidator;
 const hoek = require('hoek');
 const urlLib = require('url');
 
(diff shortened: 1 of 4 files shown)


plugins/templates/create.js
```diff
@@ -21,38 +23,42 @@ module.exports = () => ({
...
-                if (templates.length === 0) {
-                    return templateFactory.create(templateConfig);
-                }
+                return Promise.all([
+                    pipelineFactory.get(pipelineId),
+                    templateFactory.list({ params: { name: config.template.name } })
+                ]).then(([pipeline, templates]) => {
+                    const templateConfig = hoek.applyToDefaults(config.template, {
+                        pipelineId: pipeline.id,
+                        labels: config.template.labels || []
```
petey: Are we really making labels part of the template definition? That doesn't seem right. This used to be part of the original payload.

d2lam: I'm not sure what you mean @petey 
`labels` is still part of the original payload. `config` is the parsed version of yaml. This line is to just to say if user does not define it, then set it to `[]`.

d2lam:
@tkyi  I also thought it was kinda weird, but I guess it comes down to what is the purpose of the endpoint. 
The CREATE endpoint's purpose is to create a template, so if the template is invalid, it * create, so it should throw error. 
The VALIDATE endpoint's purpose is to validate. When the template is invalid, it can still validate it. So it should still reply, and display what errors are there in the template.

plugins/templates/create.js
petey: ok, we need to readdress labels in a later PR. The way they are implemented currently is not the way we should implement them in the future.

PR #530, opened 2017-04-21
feat(300): add step definitions for shared steps

tk3fftk:
There is no functional test for shared steps. This PR is implementing step definitions for shared steps.

The test passed locally.

Use this test-repo and need merge before  https://github.com/screwdriver-cd-test/functional-shared-steps/pull/1

### DIFF
--- a/features/step_definitions/sd-step.js
+++ b/features/step_definitions/sd-step.js
@@ -1,0 +1,123 @@
+'use strict';
+
+/* eslint-disable no-unused-vars */
+
+const Assert = require('chai').assert;
+const request = require('../support/request');
+const sdapi = require('../support/sdapi');
+const TIMEOUT = 240 * 1000;
+
+module.exports = function server() {
+    // eslint-disable-next-line new-cap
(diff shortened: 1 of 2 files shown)


features/step_definitions/sd-step.js
```diff
@@ -0,0 +1,123 @@
...
+
+    this.Given(/^an existing pipeline with these images and packages with version:$/,
+        { timeout: TIMEOUT }, table =>
+        this.getJwt(this.accessKey)
+        .then((response) => {
+            this.jwt = response.body.token;
+            this.expectedImage = table.rows()[0][0];
+            this.expectedPackage = table.rows()[0][1];
+
+            return request({
```
tkyi: Wonder if we could refactor this `POST` request at some point
Totally not a blocking comment, though

tk3fftk: I agree. This `POST` request is used by other step definitions.
And I think `checkoutUrl` should be configurable (SCM url and branch).

Issue #244, opened 2016-10-05
Restart PR from API targets wrong SHA


d2lam:
When I click on PR-239 from the UI:
<img width="1293" alt="screen shot 2016-10-05 at 2 50 41 pm" src="https://cloud.githubusercontent.com/assets/3401924/19133021/258e1d60-8b0b-11e6-8b22-7b744542c74e.png">

It takes me to: https://cd.screwdriver.cd/pipelines/27cec39c9baea6cf64c684479115300d22039b2a/build/7a298ec06f4bd5d5bd543ed9cc74b156eb8edca2

When I click on the target_url from commit status:
<img width="651" alt="screen shot 2016-10-05 at 2 57 31 pm" src="https://cloud.githubusercontent.com/assets/3401924/19133239/1f9db5ea-8b0c-11e6-8af4-1ff58f131e02.png">

It takes me to: https://cd.screwdriver.cd/pipelines/27cec39c9baea6cf64c684479115300d22039b2a/build/d489f4ad4204f6b8ca3438ffdb5a3508efcbf811

By looking at this: https://api.screwdriver.cd/v4/jobs/a7057c60e15ec92582f6a2c5b9aa739418844a5d/builds. UI is right, commit status is wrong.

stjohnjohnson:
This is because the job start doesn't load the SHA from the PR, it loads from master.

minz1027:
Implementation:

In API `plugins/builds/create.js`
- [ ] pipeline.sync will take an extra param `jobName`
https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/builds/create.js#L53
- [ ] getCommitSha should take in `prNum` too
https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/builds/create.js#L55

In the SCM plugins (base, github and bitbucket):
- [ ] Add a function `getPrRef(config)` to get the prRef by config
- [ ] Modify `getCommitSha` to take in `prNum` and return the correct sha

In the model:
- [ ] pipeline.sync will be a wrapper and decide it should sync the PR or the pipeline
- [ ] add a function `_syncPR()` to sync the PR job's permutation

d2lam:
This is done. Should be working now.

PR #546, opened 2017-05-08
feat(244): restart PR

d2lam:
If the job is a PR, calls `syncPR()` to get the configuration from `screwdriver.yaml` on that branch (done inside `syncPR`). 

Also passes in `prNum` to `getCommitSha`. This function is modified to return the correct sha of the PR instead of the sha from master.

Blocked by: https://github.com/screwdriver-cd/models/pull/159
Related: https://github.com/screwdriver-cd/screwdriver/issues/244


plugins/builds/create.js
```diff
@@ -50,9 +50,19 @@ module.exports = () => ({
...
+                        return pipeline.sync();
+                    })
                     .then(() => user.unsealToken())
-                    .then(token => scm.getCommitSha({ token, scmUri: pipeline.scmUri }))
+                    .then(token => scm.getCommitSha({
+                        token,
+                        scmUri: pipeline.scmUri,
+                        prNum: job.prNum
+                    }))
                     .then(sha => eventFactory.create({
```
minz1027: Need to pass the correct `sha` to payload so that new build create will have the correct sha.
https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/builds/create.js#L65

d2lam: Fixed

PR #547, opened 2017-05-08
docs: Add quickstart guide for setting up a local copy of the API

ian-fox:
Add a quickstart guide with instructions to set up a local copy of the API while avoiding the problems I encountered while doing so.


tkyi:
I think the general README could just be updated with this information. Or it would be valuable to clean up the README and add a link from the Local API section in the README to your doc.

d2lam:
I agree with @tkyi. Some duplications from here: https://github.com/screwdriver-cd/screwdriver#from-source

I think we should modify the current README instead of creating more docs

ian-fox:
Definitely a possibility. I also learned that there's [this](http://docs.screwdriver.cd/cluster-management/configure-api/) page of docs that I just somehow completely missed, probably because I passed over it before I realized what clusters really were. 

It has some stuff that I missed (like the mandatory JWT key) that I wouldn't have if I had read it, that's on me. Though having a link to it from the main readme may also be useful.

d2lam:
> It has some stuff that I missed (like the mandatory JWT key) that I wouldn't have if I had read it, that's on me. Though having a link to it from the main readme may also be useful.

@ian-fox If the page isn't easy or obvious to get to for new users then it is a problem. I agree we need to link it better or make it more discoverable in some ways.

README.md
```diff
@@ -8,6 +8,7 @@
 ### Prerequisites
 
 - Node v6.0.0 or higher
+- [Kubernetes][kubectl] or [Docker][docker]
```
ian-fox: This part I'm not super pleased with because it links to the kubectl page instead of the generic kubernetes website, but I feel like the kubectl page is useful, and just by looking at the kubernetes quickstart guide it isn't super easy to find. If anybody can think of a better way to link to that page please let me know!

README.md
```diff
@@ -45,6 +46,8 @@ $ python <(curl https://raw.githubusercontent.com/screwdriver-cd/screwdriver/mas
 
 Screwdriver already [defaults most configuration](config/default.yaml), but you can override defaults using a `local.yaml` or environment variables.
 
+For more information, see the [cluster management docs](https://github.com/screwdriver-cd/guide/blob/master/docs/cluster-management/configure-api.md#managing-the-api).
```
r3rastogi: I think we should clarify here that cluster management is necessary for most users. Otherwise, if you had read this when setting things up, you might have still glossed over the docs thinking that it was irrelevant. At minimum, we could say something like "To continue set up, follow the instructions for cluster management".

ian-fox: That is an excellent idea!

Issue #545, opened 2017-05-08
Build Timeout


petey:
As a screwdriver user,
I want my builds to timeout after some period of time,
So I'm not eating up system resources forever if something goes wrong.

We should add some configuration to allow jobs to be forcefully terminated as a failure if they go over established time limits

First Pass:
- [x] Jobs will be killed by the executor after 90 minutes of run time.
- [x] Jobs can be configured by users with a timeout of up to a maximum of (X) minutes, with a system default of 90 minutes if not defined.
- [ ] Validate annotation - Max: 10hrs (Builds should run fine if user timeout is less than the max)
- [ ] Default timeout from the cluster admin side

minz1027:
### Updates 02/26
https://github.com/screwdriver-cd/executor-queue/blob/master/index.js#L64
https://github.com/taskrabbit/node-resque/tree/v4

Since we have queue now, we can actually leverage on `resque`'s `enqueuIn` function, which enqueues the job in a future time from now. Whenever we enqueue a build `start` job, we can also enqueue a `stop` job for that build and delay it's execution to be 90 mins later.

User can override the default timeout in their job annotations. We can check on that and set the delay time properly.

FenrirUnbound:
A point was brought up that we should also be wary of setup and teardown steps timing out. On one hand, if a teardown step does not exit correctly, the build could left in a hanging state. An opposing concern that came up was: by managing all steps under a single timeout, we lose out on the opportunity to debug a build if it times out before executing the artifact upload procedure.

Should we consider separate timeouts for everything (e.g., timeout for setup, timeout for user-defined steps, timeout for teardown)?

jithin1987:
I believe timeout for setup and user defined steps should be same, no need to make it complicated for our implementation.

One one hand if user defined steps are failing then there might be value in having artifacts saved, but it might really be a bad test or deployment which is stuck causing the step to get aborted, then there won't be any legit artifacts to be saved.

So we can either just do one timeout overall or do a grace period of 10 - 15mins to finish uploading artifacts.

FenrirUnbound:
> I believe timeout for setup and user defined steps should be same, no need to make it complicated for our implementation.
>
> One one hand if user defined steps are failing then there might be value in having artifacts saved, but it might really be a bad test or deployment which is stuck causing the step to get aborted, then there won't be any legit artifacts to be saved.
>
> So we can either just do one timeout overall or do a grace period of 10 - 15mins to finish uploading artifacts.

[...]

PR #548, opened 2017-05-09
[DNM] feat(545): set timeout time when builds start running

petey:
Context:
--------
Builds should have a time limit on execution

Objective:
----------
A `timeoutTime` is added to a build when its status is changed to "RUNNING". Ideally, this would be affected by some user configuration.

Misc:
-----
* Related to #545
* Refactored the actions that occur after status is updated to separate methods to make the handler a little cleaner

TODO:
-----
* [ ] Get a value from some build/job configuration (settings)
* [ ] Act on the build timeout value (`setTimeout()`, cron, ...)


petey:
Let's step back and take another look at this after we have a build queue. It is possible that we will want to implement in the queue instead of the api.

Issue #532, opened 2017-04-24
[Technical] Implementing User API Token


FenrirUnbound:
#As a one-time experiment, separate the technical work from the user-facing feature work. The objective is: separate the levels of discussion, allowing dedicated spaces for technical collaboration and feature-level conversations.

Developer updates & technical conversations would be attached here. 

# Summary
**High-level description**: Users can generate an access token that can be used to interact with the v4 API, as an alternative to OAuth (think Github access token)
**Full-stack project** (front-end for user-friendly token generation; server side; DB to store issued tokens and track valid/invalid ones); good mix of front/back
**Story size**: Medium-sized
**Technologies used**: Node.JS
**Impact/value**: Better security

Main Feature (with rules/requirements) - #524 

---
## Preparation

- [x] Feature file (#538)

### API

- [x] Tokens endpoint
   - [x] Create tokens
   - [x] List tokens
      - [x] List tokens via pagination
   - [x] Update token name/description
   - [x] Delete tokens
   - [x] Update "last seen" token property

### Data Schema

- [x] Schema for Tokens endpoint (API)
- [x] Schema for tokens model

### Models

- [x] Tokens model

### UI

- [x] Personal Preferences page
 - [ ] Re-scope after API is completed

## Outcome

- [x] Passing acceptance tests described in feature file
   - [x] (Optional) Tech Debt: Use issued token in acceptance tests
- [x] Documentation
- [x] Blog post
- [ ] Optional: Video tutorial

## References
* Github Personal Access tokens (for reference)
   * https://github.com/settings/tokens
* OAuth testing
   * CasperJS
      * https://github.com/oauth-io/testing-oauth
   * Zombie.js
      * https://blog.bandwidth.com/testing-oauth-integration-flows-with-zombie-js/
      * https://github.com/a-type/zombie-oauth-example

### 2017-04-24

* Initial spec
* Added blog post as a requirement for definition of done (tkyi)
* Added feature file criteria in addition to OAuth testing references
   * Included ZombieJS references

### 2017-04-25

* Add Summary for context
* Tokens table details
* Add update endpoint for API
* Add schema for token table?
* Add Github personal tokens link under References

### 2017-04-26

* Included UI work

### 2017-05-10
* Removed requirements for datastore, and data-schema table. We don't need it. 

### 2017-05-30

* Included pagination for listing the user's tokens

tkyi:
Should we also add feature file to this list?

[...]

PR #556, opened 2017-05-19
feat: Add authentication with tokens

ian-fox:

* Token generating function now takes a boolean parameter for if it should put an expiry date on the token
* Token validation function checks that any tokens without an expiry date have not been revoked

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -47,6 +47,7 @@ function prettyPrintErrors(request, reply) {
  * @param  {Factory}     config.userFactory         User Factory instance
  * @param  {Factory}     config.buildFactory        Build Factory instance
  * @param  {Factory}     config.secretFactory       Secret Factory instance
+ * @param  {Factory}     config.tokenFactory        Token Factory instance
  * @param  {Factory}     config.eventFactory        Event Factory instance
  * @param  {Function}    callback                   Callback to invoke when server has started.
  * @return {http.Server}                            A listener: NodeJS http.Server object
@@ -85,6 +86,7 @@ module.exports = (config) => {
         userFactory: config.userFactory,
         buildFactory: config.buildFactory,
         secretFactory: config.secretFactory,
+        tokenFactory: config.tokenFactory,
         eventFactory: config.eventFactory,
         ecosystem: config.ecosystem
     };
--- a/plugins/auth/index.js
+++ b/plugins/auth/index.js
@@ -66,15 +66,22 @@ exports.register = (server, options, next) => {
     });
 
     /**
-     * Generates a jwt that is signed and has a 12h lifespan
+     * Generates a jwt that is signed
      * @method generateToken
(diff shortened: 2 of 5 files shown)


plugins/auth/index.js
```diff
@@ -105,8 +117,25 @@ exports.register = (server, options, next) => {
...
+                        .then(user => user.validateToken(token.uuid))
+                        .then(() => {
+                            cb(null, true, token);
+                        })
+                        .catch(() => {
+                            cb(null, false, {});
+                        });
+                    }
+
+                    // Token is a 12h access token
```
ian-fox: Notably, if the token is a 12h access token that has expired, it won't even make it to the validateFunc, so if it made it this far we don't need to check that it hasn't expired.

plugins/auth/index.js
```diff
@@ -66,15 +67,26 @@ exports.register = (server, options, next) => {
...
-     * @return {String}         Signed jwt that includes that profile
+     * @param  {Object}     profile     Object from generateProfile
+     * @param  {Boolean}    expires     True if the token should expire
+     * @return {String}                 Signed jwt that includes that profile
      */
-    server.expose('generateToken', profile => jwt.sign(profile, pluginOptions.jwtPrivateKey, {
-        algorithm: ALGORITHM,
-        expiresIn: EXPIRES_IN
-    }));
+    server.expose('generateToken', (profile, expires = true) => {
```
ian-fox: This is gross and there's probably a better way to do this, please if somebody has an idea let me know.

lib/server.js
```diff
@@ -105,7 +107,8 @@ module.exports = (config) => {
                 server.app.buildFactory.apiUri = server.info.uri;
                 server.app.buildFactory.tokenGen = (buildId, metadata) =>
                     server.plugins.auth.generateToken(
-                        server.plugins.auth.generateProfile(buildId, ['build'], metadata)
+                        server.plugins.auth.generateProfile(buildId, ['build'], metadata),
+                        true
```
minz1027: if you have the default value as `true` then no need for this.

ian-fox: Right you are!

plugins/auth/index.js
```diff
@@ -66,15 +66,23 @@ exports.register = (server, options, next) => {
     });
 
     /**
-     * Generates a jwt that is signed and has a 12h lifespan
+     * Generates a jwt that is signed
      * @method generateToken
-     * @param  {Object} profile Object from generateProfile
-     * @return {String}         Signed jwt that includes that profile
+     * @param  {Object}     profile     Object from generateProfile
+     * @param  {Boolean}    expires     True if the token should expire
```
minz1027: No need for this.

ian-fox:
Closed as authentication is being redone to trade a different access token for JWT's, rather than just using them directly

Issue #524, opened 2017-04-13
User API Tokens


petey:
> Users should be able to request a reusable token that can be used to identify them when making API requests. This is similar to tokens that github provides for the same purposes.

Currently the only support method of authentication is via OAuth flow. This works out great for humans to interact with the Screwdriver service. The downside with supporting only a signle authentication mechanism is it is incredibly difficult for developers to integrate tools or add on external features.

As an additional option, Screwdriver should be able to issue access tokens. This would allow users to interact with the Screwdriver API in a programmatic way. 

-----

Rules:

- Tokens should be specifically tied to the user that issued them
- Tokens should have a label/name associated with them
- Users can add a description to their issued tokens, so they know what they're using them for
- Users can revoke a token at any time.
- User can see when a token was last used
- No one can ever look up the raw value after initially generating the token

Discussion Items:

- Being able to generate unique tokens service-wide
- Secure methods of storing the tokens
   - Unique salt, similar to secrets
- Additional feature: token lifespans (purposely issuing a token with an expire time)

## Work

Outline & task list - #532

### 2017-04-26

* Removed `scope` requirement

### 2017-04-24 fenrirunbound

* Initial spec
   * First pass at the rules, work items
   * Topics that should be discussed or are possibly out-of-scope are documented in the *Discussion Items* section

FenrirUnbound:
The feature is labelled with `component/API`. Do we also want to include UI-related features for token management (e.g., generating new tokens, deleting)? Or are there plans to include that in a separate feature?

stjohnjohnson:
On a side note, we just realized that it would be amazing to have `pipeline` tokens as well.  That way it reduces the scope if the token is compromised.

petey:
Done.

PR #563, opened 2017-05-26
feat: Authenticate with API tokens

ian-fox:
* Users can trade their access token for a 12h JWT

Related to https://github.com/screwdriver-cd/screwdriver/issues/524

### DIFF
--- a/plugins/auth/index.js
+++ b/plugins/auth/index.js
@@ -113,15 +113,30 @@ exports.register = (server, options, next) => {
                 accessTokenName: 'access_key',
                 allowCookieToken: false,
                 allowQueryToken: true,
-                validateFunc: (token, cb) => {
-                    if (token !== pluginOptions.temporaryAccessKey) {
-                        return cb(null, false, {});
+                validateFunc: function _validateFunc(token, cb) {
+                    if (token === pluginOptions.temporaryAccessKey) {
+                        return cb(null, true, {
+                            username: pluginOptions.temporaryAccessUser,
+                            scope: ['user']
+                        });
                     }
 
-                    return cb(null, true, {
-                        username: pluginOptions.temporaryAccessUser,
-                        scope: ['user']
-                    });
+                    // Token is an API token
+                    // using function syntax makes 'this' the request
+                    // TODO: Should log that we're authenticating a user with a token
+                    const factory = this.server.app.userFactory;
+
+                    return factory.get({ accessToken: token })
+                        .then((user) => {
+                            if (!user) {
+                                return cb(null, false, {});
+                            }
+
+                            return cb(null, true, {
+                                username: user.username,
+                                scope: ['user']
+                            });
+                        });
                 }
             });
 
--- a/test/plugins/auth.test.js
+++ b/test/plugins/auth.test.js
@@ -434,6 +434,7 @@ describe('auth plugin test', () => {
         const id = '1234id5678';
         const username = 'batman';
         const token = 'qpekaljx';
+        const apiKey = 'aUserApiToken';
         const user = {
             id,
             username,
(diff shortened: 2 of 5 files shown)


test/plugins/auth.test.js
```diff
@@ -485,14 +486,34 @@ describe('auth plugin test', () => {
             })
         );
 
-        it('fails to issue a jwt given an invalid application auth token', () =>
+        it('returns user signed token given an API access token', () =>
             server.inject({
+                url: `/auth/token?access_key=${apiKey}`
```
tkyi: Is there a possibility that the `temporaryAccessKey` and `apiKey` could collide?

ian-fox: All of our access keys match `/[a-zA-Z0-9+/]{43}=/`, so if somebody picks an access key that matches that and it happens to be the same as one that we generate then there would be a collision.

ian-fox: It would be very unlikely though.

FenrirUnbound: > Is there a possibility that the temporaryAccessKey and apiKey could collide?

Yes. Once we fully support the app tokens, we should switch away from the `temporaryAccessKey`.

plugins/auth/index.js
```diff
@@ -113,15 +113,30 @@ exports.register = (server, options, next) => {
                 accessTokenName: 'access_key',
                 allowCookieToken: false,
                 allowQueryToken: true,
-                validateFunc: (token, cb) => {
-                    if (token !== pluginOptions.temporaryAccessKey) {
-                        return cb(null, false, {});
+                validateFunc: function _validateFunc(token, cb) {
```
FenrirUnbound: Why switch away from fat-arrow notation?

ian-fox: Arrow notation doesn't bind the request object to `this`

The only thing I need the request object for is to get access to the userFactory though, so if there's a better way to get at that then there's no need for `function` notation.

FenrirUnbound: What's different between `this.server` and `server` here? https://github.com/screwdriver-cd/screwdriver/blob/cf158a135aa11171683d9d4bc8cc58fc2e596f9b/plugins/auth/index.js#L112

plugins/auth/index.js
```diff
@@ -114,14 +114,29 @@ exports.register = (server, options, next) => {
...
                     }
 
-                    return cb(null, true, {
-                        username: pluginOptions.temporaryAccessUser,
-                        scope: ['user']
-                    });
+                    // Token is an API token
+                    // using function syntax makes 'this' the request
+                    // TODO: Should log that we're authenticating a user with a token
+                    const factory = server.app.userFactory;
```
minz1027: I remember when you try to access server.app not from the request, it is kinda weird. I tried it locally, it is actually `server._parent.app.userFactory`. Can you start the server locally and confirm?

ian-fox: You're totally right. I wonder why the automated test didn't catch that.

ian-fox: Accessing private properties seems kind of suspect though, would it be better to use `function` notation and get the factory from the request?

ian-fox: Eslint actually complains about it too, won't let you run npm test if you try to access `server._parent`. I think moving back to `function` notation and using the request is probably the correct choice.

plugins/auth/index.js
ian-fox: Apparently the `server` object doesn't actually have `server.app.userFactory`, to get that you need to get `server._parent.app.userFactory`. Using a private property seems suspect, and ESLint actually won't allow it. 

Between using `server._parent` and disabling the ESLint rule for the line or using `function` notation, I feel like `function` is cleaner.

test/plugins/auth.test.js
tkyi: Ohh okay. So we can merge this PR but make sure we do away with the `temporaryAccessKey` soon after.

PR #565, opened 2017-05-30
feat: Add token CREATE route

ian-fox:
* `POST`ing to /tokens will create a new token for the logged in user.

### DIFF
--- a/plugins/tokens/create.js
+++ b/plugins/tokens/create.js
@@ -1,0 +1,70 @@
+'use strict';
+
+const boom = require('boom');
+const schema = require('screwdriver-data-schema');
+const urlLib = require('url');
+
+module.exports = () => ({
+    method: 'POST',
+    path: '/tokens',
+    config: {
+        description: 'Create a new token',
+        notes: 'Create a specific token',
+        tags: ['api', 'tokens'],
+        auth: {
+            strategies: ['token', 'session'],
+            scope: ['user']
+        },
+        plugins: {
+            'hapi-swagger': {
+                security: [{ token: [] }]
+            }
+        },
+        handler: (request, reply) => {
+            const tokenFactory = request.server.app.tokenFactory;
+            const userFactory = request.server.app.userFactory;
+            const username = request.auth.credentials.username;
+
+            return userFactory.get({ username })
+            .then((user) => {
+                if (!user) {
+                    throw boom.notFound(`User ${username} does not exist`);
+                }
+
+                return user.tokens
+                    .then((tokens) => {
+                        // Make sure the name is unique
+                        const match = tokens && tokens.find(t => t.name === request.payload.name);
+
+                        if (match) {
+                            throw boom.conflict(`Token with name ${match.name} already exists`);
+                        }
+
+                        return tokenFactory.create({
+                            name: request.payload.name,
+                            description: request.payload.description,
+                            userId: user.id
+                        });
+                    });
+            })
+            .then((token) => {
+                const location = urlLib.format({
+                    host: request.headers.host,
+                    port: request.headers.port,
+                    protocol: request.server.info.protocol,
+                    pathname: `${request.path}/${token.id}`
+                });
+                const output = token.toJson();
+
+                delete output.hash;
+
+                return reply(output).header('Location', location).code(201);
+            })
+            // something broke, respond with error
+            .catch(err => reply(boom.wrap(err)));
+        },
+        validate: {
+            payload: schema.models.token.create
+        }
+    }
+});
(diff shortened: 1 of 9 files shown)


ian-fox:
For some reason the tests are all passing on my machine but failing on screwdriver, investigating now.

minz1027:
@ian-fox Ian, the last merge also failed the same unit test, probably the same reason.

ian-fox:
@minz1027 Seems to be something to do with dependencies, got the tests to start failing on my machine after I `rm`'d the node_modules folder and did a fresh `npm install`

test/plugins/tokens.test.js
```diff
@@ -0,0 +1,190 @@
...
+const getUserMock = (user) => {
+    const mock = hoek.clone(user);
+
+    mock.tokens = sinon.stub();
+
+    return mock;
+};
+
+const tokensGetterMock = tokens =>
+    new Promise(resolve => resolve(tokens));
```
minz1027: This can be tokensGetterMock = tokens => Promise.resolve(tokens);

test/plugins/tokens.test.js
```diff
@@ -0,0 +1,190 @@
...
+
+const getUserMock = (user) => {
+    const mock = hoek.clone(user);
+
+    mock.tokens = sinon.stub();
+
+    return mock;
+};
+
+const tokensGetterMock = tokens =>
```
FenrirUnbound: `const tokensGetterMock = tokens => Promise.resolve(tokens)`

test/plugins/tokens.test.js
```diff
@@ -0,0 +1,190 @@
...
+            port: 1234
+        });
+
+        server.auth.scheme('custom', () => ({
+            authenticate: (request, reply) => reply.continue({})
+        }));
+        server.auth.strategy('token', 'custom');
+        server.auth.strategy('session', 'custom');
+
+        server.register([{
```
FenrirUnbound: `server.register` can also return a promise.

test/plugins/tokens.test.js
```diff
@@ -0,0 +1,189 @@
...
+        userMock = getUserMock({
+            username,
+            id: userId
+        });
+        userMock.tokens = tokensGetterMock([]);
+        userFactoryMock.get.withArgs({ username }).resolves(userMock);
+    });
+
+    afterEach(() => {
+        server = null;
```
FenrirUnbound: Probably want to call `server.stop` before removing the only reference to it

test/plugins/tokens.test.js
ian-fox: So I should do the mock stuff first, and then `return` server.register?

plugins/tokens/create.js
```diff
@@ -0,0 +1,72 @@
...
+            }
+        },
+        handler: (request, reply) => {
+            const tokenFactory = request.server.app.tokenFactory;
+            const userFactory = request.server.app.userFactory;
+            const username = request.auth.credentials.username;
+            let user;
+
+            return userFactory.get({ username })
+            .then((theUser) => {
```
d2lam: Hmm why not just leave this as `.then((user) => ...`, and have the `.then(tokens)` inside this user scope? I think that would be cleaner.

plugins/tokens/create.js
```diff
@@ -0,0 +1,72 @@
...
+                user = theUser;
+
+                return user.tokens;
+            })
+            .then((tokens) => {
+                // Make sure the name is unique
+                const match = tokens && tokens.find(t => t.name === request.payload.name);
+
+                if (match) {
+                    throw boom.conflict('Token with that name already exists');
```
d2lam: Would be clearer to have the `${name}` here

test/plugins/tokens.test.js
```diff
@@ -0,0 +1,189 @@
...
+const tokensGetterMock = tokens => Promise.resolve(tokens);
+
+const generateProfile = sinon.stub();
+
+describe('token plugin test', () => {
+    const username = 'ifox';
+    const userId = testToken.userId;
+    let tokenFactoryMock;
+    let userFactoryMock;
+    let authMock;
```
minz1027: why we need the authMock here? This is only creating a token and save it in datastore. We shouldn't need to mock auth stuff.

test/plugins/tokens.test.js
FenrirUnbound: Yup, that approach is valid

test/plugins/tokens.test.js
ian-fox: Right, remnant from something else.

PR #567, opened 2017-05-30
fix: fix failing unit tests

minz1027:
Due to some dependency changes (still investigating), `throw` boom error without return reply() will cause 500. This PR is to fix the current failing tests.


ian-fox:
Looks good! :d2thanks:

PR #568, opened 2017-05-30
User tokens

ian-fox:
* `GET /tokens` will return a list of all tokens belonging to the current user

### DIFF
--- a/plugins/tokens/list.js
+++ b/plugins/tokens/list.js
@@ -1,0 +1,50 @@
+'use strict';
+
+const boom = require('boom');
+const joi = require('joi');
+const schema = require('screwdriver-data-schema');
+const getSchema = joi.array().items(schema.models.token.get);
+
+module.exports = () => ({
+    method: 'GET',
+    path: '/tokens',
+    config: {
+        description: 'Get tokens with pagination',
+        notes: 'Returns all token records belonging to the current user',
+        tags: ['api', 'tokens'],
+        auth: {
+            strategies: ['token', 'session'],
+            scope: ['user']
+        },
+        plugins: {
+            'hapi-swagger': {
+                security: [{ token: [] }]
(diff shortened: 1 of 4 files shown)


plugins/tokens/list.js
```diff
@@ -0,0 +1,53 @@
...
+
+            return userFactory.get({ username })
+            .then((user) => {
+                if (!user) {
+                    throw boom.notFound(`User ${username} does not exist`);
+                }
+
+                return user.tokens;
+            })
+            .then(tokens => reply(tokens.map((t) => {
```
FenrirUnbound: Preference to using meaningful variable names

test/plugins/tokens.test.js
```diff
@@ -52,7 +52,8 @@ describe('token plugin test', () => {
     beforeEach((done) => {
         tokenFactoryMock = {
             create: sinon.stub(),
-            get: sinon.stub()
+            get: sinon.stub(),
+            list: sinon.stub()
```
FenrirUnbound: This doesn't seem to be used

FenrirUnbound: In fact, I don't see where having the `tokenFactoryMock` is significant

plugins/tokens/list.js
```diff
@@ -0,0 +1,53 @@
...
+                delete output.userId;
+
+                return output;
+            })))
+            .catch(err => reply(boom.wrap(err)));
+        },
+        response: {
+            schema: getSchema
+        },
+        validate: {
```
FenrirUnbound: I don't think route handles pagination just yet

test/plugins/tokens.test.js
ian-fox: Ah, right, because `userFactory` is mocked

PR #569, opened 2017-05-31
fix(tests): Fix a pipelines test that was failing

ian-fox:
One of the tests in pipeline.test.js was failing, but it wasn't showing up in the summary at the end because the test was using a callback instead of a `then` block.

I *think* this is what the test should be testing for, by looking at the code for pipelines and talking to @d2lam, but @d2lam please confirm this!

### DIFF
--- a/test/plugins/pipelines.test.js
+++ b/test/plugins/pipelines.test.js
@@ -856,7 +856,7 @@ describe('pipeline plugin test', () => {
     describe('POST /pipelines', () => {
         let options;
         const unformattedCheckoutUrl = 'git@github.com:screwdriver-cd/data-MODEL.git';
-        const checkoutUrl = 'git@github.com:screwdriver-cd/data-model.git';
+        const formattedCheckoutUrl = 'git@github.com:screwdriver-cd/data-model.git#master';
         const scmUri = 'github.com:12345:master';
         const scmRepo = {
             id: 'github.com:123456:master'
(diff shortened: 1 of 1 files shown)


test/plugins/pipelines.test.js
```diff
@@ -856,7 +856,7 @@ describe('pipeline plugin test', () => {
     describe('POST /pipelines', () => {
         let options;
         const unformattedCheckoutUrl = 'git@github.com:screwdriver-cd/data-MODEL.git';
-        const checkoutUrl = 'git@github.com:screwdriver-cd/data-model.git';
+        const checkoutUrl = 'git@github.com:screwdriver-cd/data-model.git#master';
```
FenrirUnbound: Isn't the missing `#master` significant?

tkyi: Yeah I think it's supposed to append `#master` if it's missing.

minz1027: Yes, the `scm` will append the `master`: https://github.com/screwdriver-cd/scm-github/blob/master/index.js#L42. shouldn't need it here

test/plugins/pipelines.test.js
```diff
@@ -919,15 +919,11 @@ describe('pipeline plugin test', () => {
         });
 
         it('formats the checkout url correctly', () => {
-            const goodCheckoutUrl = 'git@github.com:screwdriver-cd/data-model.git#master';
+            userMock.getPermissions.withArgs(checkoutUrl).resolves({ admin: false });
```
minz1027: getPermissions will be called with `scmUri` instead of the `checkoutUrl`
https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/pipelines/create.js#L61

ian-fox: Oh shoot I could have sworn I made that change. Thanks!

test/plugins/pipelines.test.js
ian-fox: `checkoutUrl` was only used where the scmUri should have been, so I renamed `formattedCheckoutUrl` to `checkoutUrl` for line length. Probably less clear though, reverting back to `formattedCheckoutUrl`.

PR #573, opened 2017-05-31
feat(tokens): Add token UPDATE route

ian-fox:
* `POST` requests to `/tokens/{id}` will update the specified token
* Name can't be changed such that it conflicts with another token

### DIFF
--- a/plugins/tokens/update.js
+++ b/plugins/tokens/update.js
@@ -1,0 +1,79 @@
+'use strict';
+
+const boom = require('boom');
+const joi = require('joi');
+const schema = require('screwdriver-data-schema');
+const idSchema = joi.reach(schema.models.token.base, 'id');
+
+module.exports = () => ({
+    method: 'PATCH',
+    path: '/tokens/{id}',
+    config: {
+        description: 'Update a token',
+        notes: 'Update a specific tooken',
+        tags: ['api', 'secrets'],
+        auth: {
+            strategies: ['token', 'session'],
+            scope: ['user']
+        },
+        plugins: {
+            'hapi-swagger': {
+                security: [{ token: [] }]
+            }
+        },
+        handler: (request, reply) => {
+            const tokenFactory = request.server.app.tokenFactory;
+            const userFactory = request.server.app.userFactory;
+            const username = request.auth.credentials.username;
+
+            return Promise.all([
+                tokenFactory.get(request.params.id),
+                userFactory.get({ username })
+            ])
+                .then(([token, user]) => {
+                    if (!token) {
+                        throw boom.notFound('Token does not exist');
+                    }
+
+                    if (!user) {
+                        throw boom.notFound('User does not exist');
+                    }
+
+                    if (token.userId !== user.id) {
+                        throw boom.forbidden('User does not own token');
+                    }
+
+                    return user.tokens
+                        .then((tokens) => {
+                            // Make sure it won't cause a name conflict
+                            const match = tokens && tokens.find(
+                                t => t.name === request.payload.name);
+
+                            if (match && request.params.id !== match.id) {
+                                throw boom.conflict(`Token with name ${match.name} already exists`);
+                            }
+
+                            Object.keys(request.payload).forEach((key) => {
+                                token[key] = request.payload[key];
+                            });
+
+                            return token.update()
+                            .then(() => {
+                                const output = token.toJson();
+
+                                delete output.hash;
+
+                                return reply(output).code(200);
+                            });
+                        });
+                })
+                .catch(err => reply(boom.wrap(err)));
+        },
+        validate: {
+            params: {
+                id: idSchema
+            },
+            payload: schema.models.token.update
+        }
+    }
+});
--- a/test/plugins/tokens.test.js
+++ b/test/plugins/tokens.test.js
@@ -5,25 +5,24 @@ const sinon = require('sinon');
 const hapi = require('hapi');
 const mockery = require('mockery');
 const urlLib = require('url');
-const hoek = require('hoek');
 const testToken = require('./data/token.json');
 
 sinon.assert.expose(assert, { prefix: '' });
 require('sinon-as-promised');
 
 const getTokenMock = (token) => {
-    const mock = hoek.clone(token);
+    const mock = Object.assign({}, token);
 
     mock.update = sinon.stub();
     mock.regenerate = sinon.stub();
-    mock.toJson = sinon.stub().returns(token);
+    mock.toJson = sinon.stub().returns(Object.assign({}, token));
     mock.remove = sinon.stub();
 
     return mock;
 };
 
 const getUserMock = (user) => {
-    const mock = hoek.clone(user);
+    const mock = Object.assign({}, user);
 
     mock.tokens = sinon.stub();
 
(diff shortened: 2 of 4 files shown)


test/plugins/tokens.test.js
```diff
@@ -16,7 +16,7 @@ const getTokenMock = (token) => {
 
     mock.update = sinon.stub();
     mock.regenerate = sinon.stub();
-    mock.toJson = sinon.stub().returns(token);
+    mock.toJson = sinon.stub().returns(hoek.clone(token));
```
ian-fox: The fact that the `GET` endpoint calls `toJson` and then `delete`s some of the fields was causing problems with other tests. It should have been like this all along.

plugins/tokens/update.js
```diff
@@ -0,0 +1,80 @@
...
+                    // console.log(user.id);
+                    throw boom.forbidden('User does not own token');
+                }
+
+                return user.tokens
+                    .then((tokens) => {
+                        // Make sure it won't cause a name conflict
+                        const match = tokens && tokens.find(t => t.name === request.payload.name);
+
+                        if (match && request.payload.name && request.params.id !== match.id) {
```
FenrirUnbound: Why do you need to check for `request.payload.name`?

ian-fox: Oh actually that's a good point. I was thinking if they didn't include a new name there couldn't be a name conflict, but in that case there wouldn't be a `match` in the first place.

plugins/tokens/update.js
```diff
@@ -0,0 +1,80 @@
...
+                if (!token) {
+                    throw boom.notFound('Token does not exist');
+                }
+
+                if (!user) {
+                    throw boom.notFound('User does not exist');
+                }
+
+                if (token.userId !== user.id) {
+                    // console.log(token.userId);
```
tkyi: `console.log`

plugins/tokens/update.js
```diff
@@ -0,0 +1,80 @@
+'use strict';
+
+const boom = require('boom');
+const joi = require('joi');
+const schema = require('screwdriver-data-schema');
+const idSchema = joi.reach(schema.models.token.base, 'id');
+
+module.exports = () => ({
+    method: 'PUT',
```
FenrirUnbound: `PATCH` might be more appropriate

ian-fox: Today I learned that there are http methods other than `GET`, `POST`, `PUT`, and `DELETE`, and that `PUT` requires the full resource be in the request.

test/plugins/tokens.test.js
```diff
@@ -185,6 +189,8 @@ describe('token plugin test', () => {
                     scope: ['user']
                 }
             };
+
+            userMock.tokens = tokensGetterMock([tokenMock]);
```
FenrirUnbound: Not needed. This is redundant of https://github.com/screwdriver-cd/screwdriver/pull/573/files#diff-ee5dae69c9de402833a459d19c780d18R67

test/plugins/tokens.test.js
```diff
@@ -218,24 +225,119 @@ describe('token plugin test', () => {
...
                     scope: ['user']
+                },
+                payload: {
+                    name: testToken.name,
+                    description: 'a new description'
                 }
             };
 
             tokenFactoryMock.get.resolves(tokenMock);
+            userMock.tokens = tokensGetterMock([tokenMock]);
```
FenrirUnbound: Not needed. This is redundant of https://github.com/screwdriver-cd/screwdriver/pull/573/files#diff-ee5dae69c9de402833a459d19c780d18R67

test/plugins/tokens.test.js
```diff
@@ -218,24 +225,119 @@ describe('token plugin test', () => {
...
+            userFactoryMock.get.withArgs({ username }).resolves(userMock);
```
FenrirUnbound: Redundant of https://github.com/screwdriver-cd/screwdriver/pull/573/files#diff-ee5dae69c9de402833a459d19c780d18R74

test/plugins/tokens.test.js
```diff
@@ -218,24 +225,119 @@ describe('token plugin test', () => {
...
+            tokenMock.toJson.returns(expected);
+
+            return server.inject(options).then((reply) => {
+                assert.equal(reply.statusCode, 200);
+                assert.calledOnce(tokenMock.update);
+                assert.deepEqual(reply.result, expected);
+            });
+        });
+
+        it('doesn\'t error if the new name conflicts only with the old name', () => {
```
FenrirUnbound: From a testing perspective, this looks no different than this test https://github.com/screwdriver-cd/screwdriver/pull/573/files#diff-ee5dae69c9de402833a459d19c780d18R276

test/plugins/tokens.test.js
```diff
@@ -218,24 +225,119 @@ describe('token plugin test', () => {
...
                     username,
                     scope: ['user']
+                },
+                payload: {
+                    name: testToken.name,
+                    description: 'a new description'
                 }
             };
 
             tokenFactoryMock.get.resolves(tokenMock);
```
FenrirUnbound: This line is also redundant

test/plugins/tokens.test.js
```diff
@@ -218,24 +225,119 @@ describe('token plugin test', () => {
...
+                url: `/tokens/${tokenId}`,
+                credentials: {
+                    username,
+                    scope: ['user']
+                }
+            };
+
+            tokenFactoryMock.get.resolves(tokenMock);
+            userMock.tokens = tokensGetterMock([tokenMock]);
+            userFactoryMock.get.withArgs({ username }).resolves(userMock);
```
FenrirUnbound: All these are defined at the top-level `beforeEach` so there isn't a need to re-defining the same behavior here

test/plugins/tokens.test.js
FenrirUnbound: When did we introduce `hoek`? For cloning and combining object maps, ES6 has [`Object.assign`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign)

plugins/tokens/update.js
FenrirUnbound: It's a rabbit-hole of fun 😸

test/plugins/tokens.test.js
ian-fox: Ooh, did not know that existed. Will switch to that now.

test/plugins/tokens.test.js
ian-fox: Ah, there's where my confusion was. I thought that beforeEach ran before each of the `describe` blocks. Didn't realize it also ran before `it` blocks, so I thought I needed to have that there to restore functionality after doing things like having the `userFactoryMock` resolve `null` to test the 404 case.

plugins/tokens/update.js
```diff
@@ -0,0 +1,78 @@
...
+                if (!user) {
+                    throw boom.notFound('User does not exist');
+                }
+
+                if (token.userId !== user.id) {
+                    throw boom.forbidden('User does not own token');
+                }
+
+                return user.tokens
+                    .then((tokens) => {
```
FenrirUnbound: Let's keep the indentation consistent. Some of the `.then` statements are indented and some aren't.

ian-fox: Whoops, missed the one without an indent.

PR #574, opened 2017-06-01
feat(tokens): Add token refresh route

ian-fox:
* `PATCH` `/tokens/{id}/refresh` will refresh the token's value and return it while preserving the other fields of the token

### DIFF
--- a/test/plugins/tokens.test.js
+++ b/test/plugins/tokens.test.js
@@ -14,7 +14,7 @@ const getTokenMock = (token) => {
     const mock = Object.assign({}, token);
 
     mock.update = sinon.stub();
-    mock.regenerate = sinon.stub();
+    mock.refresh = sinon.stub();
     mock.toJson = sinon.stub().returns(Object.assign({}, token));
     mock.remove = sinon.stub();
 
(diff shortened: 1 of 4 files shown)


tkyi:
I kind of like `renew` or `refresh`, but that's just semantics.

ian-fox:
I don't have any attachment to the word `regenerate`. I'll ask at standup tomorrow.

test/plugins/tokens.test.js
```diff
@@ -302,6 +302,69 @@ describe('token plugin test', () => {
...
+        });
+
+        it('returns 200 if regenerated successfully', () => {
+            const expected = Object.assign({}, testToken, { value: 'newValue' });
+
+            tokenMock.toJson.returns(Object.assign({}, tokenMock.toJson(), expected));
+
+            return server.inject(options).then((reply) => {
+                assert.equal(reply.statusCode, 200);
+                assert.calledOnce(tokenMock.regenerate);
```
r3rastogi: Why are we using `regenerate` here but `refresh` elsewhere?

ian-fox: The function in the model is still called `regenerate`

I can put a PR in models to change that

ian-fox: https://github.com/screwdriver-cd/models/pull/168

PR #576, opened 2017-06-02
enhancement: remove .func_config.json

ian-fox:
Since `.func_config.json` was already in the index, it being in the .gitignore did nothing. This made it easy to accidentally publish tokens on github.

* removed `.func_config.json` from git index
* updated `readme.md` to have instructions on creating `.func_config.json` rather than modifying it, as well as adding that you can set these with environment variables as well


tkyi:
Can you stick to the semantic-release guidelines for commit messages, just for consistency.

Also, you can add Github Labels if you want to tag this PR as an enhancement.

PR #577, opened 2017-06-02
fix: do not throw err if func_config does not exist

minz1027:
It will throw an err during the `require('../../.func_config')` because we remove the `.func_config` file.
https://cd.screwdriver.cd/pipelines/1/builds/4770

### DIFF
--- a/features/support/before-hook.js
+++ b/features/support/before-hook.js
@@ -1,8 +1,7 @@
 'use strict';
 
-/* eslint-disable import/no-unresolved */
-const config = require('../../.func_config');
-/* eslint-enable import/no-unresolved */
+const path = require('path');
+const env = require('node-env-file');
 const requestretry = require('requestretry');
 const request = require('../support/request');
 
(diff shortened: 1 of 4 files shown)


features/support/before-hook.js
```diff
@@ -1,7 +1,8 @@
 'use strict';
 
+const fs = require('fs');
 /* eslint-disable import/no-unresolved */
-const config = require('../../.func_config');
+const config = fs.existsSync('../../.func_config') ? require('../../.func_config') : {};
```
FenrirUnbound: `fs.existsSync` is deprecated: https://nodejs.org/api/fs.html#fs_fs_exists_path_callback

FenrirUnbound:
There are more elegant methods to handle this. [Node Env File](https://www.npmjs.com/package/node-env-file) handles secrets much better and the interface is always through environment variables

features/support/before-hook.js
ian-fox: It looks like `fs.exists` is deprecated, but not `fs.existsSync`.

ian-fox: Though node env file does look cleaner you're right there.

minz1027:
@FenrirUnbound Good idea. Switch to use `node-env-file`. Can you re-review it?

PR #578, opened 2017-06-02
[WIP] feat(tokens): First functional test for tokens

ian-fox:
* Add first functional test for tokens
* Included another step to make sure the token gets updated when used to log in
* Fixed some routes which the functional test found problems in

### DIFF
--- a/features/api-token.feature
+++ b/features/api-token.feature
@@ -13,14 +13,12 @@ Feature: User API Token
         - Users can revoke a token at any time
         - No one can ever look up the raw value after initially generating the token
 
-    Background:
-        Given the user "calvin" exists
-
-    @ignore
+    @create
     Scenario: Generate New API Token
-        When a new API token is generated
+        When the user generates a new API token
         And the token is used to log in
-        Then a valid JWT is received that represents "calvin"
+        Then a valid JWT is received that represents the user
+        And the token's 'last used' property is updated
 
     @ignore
     Scenario: List API Tokens
--- a/features/step_definitions/api-token.js
+++ b/features/step_definitions/api-token.js
@@ -1,0 +1,105 @@
+'use strict';
+
+const Assert = require('chai').assert;
+const jwt = require('jsonwebtoken');
+const request = require('../support/request');
+const testTokenName = 'an access token';
+
+module.exports = function server() {
+    // eslint-disable-next-line new-cap
+    this.Before(() => {
+        if (!(this.accessKey)) {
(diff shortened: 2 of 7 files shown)


features/step_definitions/api-token.js
```diff
@@ -0,0 +1,85 @@
+'use strict';
+
+const Assert = require('chai').assert;
+const jwt = require('jsonwebtoken');
+const request = require('../support/request');
+const testTokenName = 'an access token';
+
+module.exports = function server() {
+    this.When(/^a new API token is generated$/, () =>
+        // Make sure there won't be a conflict: delete the token if it's already there
```
FenrirUnbound: Clean up operations should typically be defined in a `Before` block.

features/step_definitions/api-token.js
```diff
@@ -0,0 +1,85 @@
...
+                auth: {
+                    bearer: this.jwt
+                },
+                body: {
+                    name: testTokenName
+                },
+                json: true
+            })).then((response) => {
+                Assert.equal(response.statusCode, 201);
+                this.userAccessToken = response.body.value;
```
FenrirUnbound: Should also assert that the body has something in `value`. Otherwise, it'll fail in the next step and the cause of the error won't be deterministic

features/step_definitions/api-token.js
ian-fox: In cucumber it looks like `Before` is more of a `beforeEach` and runs before each scenario. The other scenarios are going to want the token to exist, whereas this one does not.

Maybe I should modify the feature file, to say something like "When calvin has no token named 'foo', and generates a token named 'foo'" so that the cleanup can be done in the bit with "has no token named foo"

ian-fox: Ooh, before hooks can be applied only to specific tags, that's probably cleaner. I'll do that.

features/api-token.feature
```diff
@@ -13,14 +13,12 @@ Feature: User API Token
         - Users can revoke a token at any time
         - No one can ever look up the raw value after initially generating the token
 
-    Background:
```
ian-fox: Moved authentication to a `before` hook, since other `before` hooks need to be authenticated.

ian-fox:
Not super pleased with the one scenario having a tag and having everything in before, rewriting to use a "given"

ian-fox:
Realized I was forgetting to `return` one of the promises, the test isn't actually passing yet. Closing.

PR #581, opened 2017-06-06
feat(tokens): First functional test for tokens

ian-fox:
* add steps to the api tokens test to ensure there won't be a conflict, and check that the lastUsed time is updated
* fix some bugs in the tokens API
     - token.remove wasn't being called correctly
     - no longer have to delete `hash` before returning token values since the new `token.toJson` does that in the models layer.
* upgrade `sinon` to version 2


ian-fox:
Oh crap, auth test is broken. Need to fix that.

ian-fox:
I don't *think* this should break the authorization test, but I can't be sure since some of them don't run on the local api, and the auth test fails when I run it against screwdriver-test because the changes in the middleware order aren't there yet.

Is there a way to see if that test will work other than clicking merge and seeing if the beta job fails?

features/step_definitions/api-token.js
```diff
@@ -0,0 +1,90 @@
+'use strict';
+
+const Assert = require('chai').assert;
+const jwt = require('jsonwebtoken');
+const request = require('../support/request');
+
+module.exports = function server() {
+    // eslint-disable-next-line new-cap
```
FenrirUnbound: Do you still need this?

ian-fox: Oh, huh, I could have sworn ESlint was complaining about it earlier. Apparently not!

ian-fox:
Doing slightly more fiddling with `GET /auth/token` upon realizing that it doesn't work with JWT's any more with this change, the rest of the code should be the same though.

ian-fox:
Talked to Petey, we're going to use the URL-safe base64 method, which means auth doesn't have to be touched after all. Flagging as WIP until I make that change (should be within the half hour)

plugins/auth/token.js
```diff
@@ -16,7 +16,7 @@ module.exports = () => ({
         notes: 'Generate a JWT for use throughout Screwdriver',
         tags: ['api', 'auth', 'token'],
         auth: {
-            strategies: ['token', 'session', 'auth_token'],
+            strategies: ['session', 'auth_token'],
```
ian-fox: Talked to @petey about this one, if `'token'` is listed among the strategies for this route then a JWT can be used to "refresh" itself by calling `GET /auth/token`.

ian-fox: Potentially should be in its own PR? Not really really related to the rest of this one.

features/step_definitions/api-token.js
```diff
@@ -0,0 +1,88 @@
...
+            }
+        }).then(() => this.getJwt(this.userAccessToken).then((response) => {
+            Assert.strictEqual(response.statusCode, 200);
+
+            this.newJwt = response.body.token;
+
+            return null;
+        })));
+
+    this.Then(/^a valid JWT is received that represents "calvin"$/, () => {
```
ian-fox: Not sure about the use of `"calvin"` here. If it's left as a regular expression cucumber complains about not having the right number of arguments in the callback. 

Ways I've seen this dealt with are using `/* eslint-disable no-unused-vars */` in `sd-step.js`, and a `switch` statement on the user with only one case for `'calvin'` in

ian-fox: Oh, `authorization` also has one where the regexp has `"main"` in it, I guess this is probably fine then.

features/api-token.feature
```diff
@@ -13,14 +13,16 @@ Feature: User API Token
...
 
+
     Background:
-        Given the user "calvin" exists
+        Given "calvin" is logged in
 
-    @ignore
     Scenario: Generate New API Token
-        When a new API token is generated
+        And does not own a token named "tiger"
```
petey: The `Background` directive is new to me, but it feels really weird in any case that Scenarios aren't using a `Given-When-Then` style.

I'd change this to `Given "calvin" does not own a token named "tiger"`

features/step_definitions/api-token.js
```diff
@@ -0,0 +1,84 @@
+'use strict';
+
+const Assert = require('chai').assert;
+const jwt = require('jsonwebtoken');
+const request = require('../support/request');
+
+module.exports = function server() {
+    this.Given(/^does not own a token named "([^"]*)"$/, tokenName =>
+        // Ensure there won't be a conflict: delete the token if it's already there
```
petey: I suggest refactoring this out somewhere and _also_ do this in an `after` hook for these tests. The expectation should be that the system is in the same "expected" state before and after tests are run.

plugins/auth/token.js
petey: I'm still pretty sure we shouldn't be able to refresh a JWT by passing the same JWT, unless this goes through the workflow used to generate the original JWT, e.g. check the auth provider [github] to ensure access hasn't been revoked.

test/plugins/tokens.test.js
```diff
@@ -15,7 +18,13 @@ const getTokenMock = (token) => {
 
     mock.update = sinon.stub();
     mock.refresh = sinon.stub();
-    mock.toJson = sinon.stub().returns(Object.assign({}, token));
+    mock.toJson = sinon.stub().returns((() => {
```
petey: Maybe this will work?
```
mock.toJson = sinon.stub().callsFake(() => {
...
});
```

plugins/auth/token.js
ian-fox: I would have thought it only went through github if using the `oath` strategy.

My naive attempt of just making a `GET` request with an Authorization header borrowed from another request didn't work though, so I may be wrong.

ian-fox:
Update June 8:
* Added cleanup in `after` hook
* Upgraded `sinon`
* General refactoring to have code be cleaner.

ian-fox:
Oh, one moment, broke something else

features/step_definitions/api-token.js
```diff
@@ -0,0 +1,102 @@
...
+                bearer: this.jwt
+            }
+        }).then(deleteResponse => Assert.strictEqual(deleteResponse.statusCode, 204));
+    });
+}
+
+module.exports = function server() {
+    this.After(() => {
+        if (this.testToken) {
+            return cleanup.apply(this, [this.testToken.name]);
```
petey: This would be cleaner with 
```
return cleanup.call(this, this.testToken.name);
```

features/step_definitions/api-token.js
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        return null;
+    });
+
+    this.Given(/^"calvin" does not own a token named "([^"]*)"$/, tokenName =>
+        // Ensure there won't be a conflict: delete the token if it's already there
+        cleanup.apply(this, [tokenName]));
```
petey: Also use `call()` here.

features/step_definitions/api-token.js
```diff
@@ -0,0 +1,102 @@
...
+            method: 'DELETE',
+            auth: {
+                bearer: this.jwt
+            }
+        }).then(deleteResponse => Assert.strictEqual(deleteResponse.statusCode, 204));
+    });
+}
+
+module.exports = function server() {
+    this.After(() => {
```
petey: Hooks typically go in the support dir.

https://github.com/screwdriver-cd/screwdriver/tree/master/features/support

https://github.com/cucumber/cucumber-js/blob/1.x/docs/support_files/hooks.md

ian-fox: Good call, created an `after-hook` file.

features/step_definitions/api-token.js
ian-fox: Ahh, good to know about! Actually switched to using a `config` object like the other functions in `support` though, since the function got moved to `support` to be accessible to the `after` hook.

features/support/after-hook.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+const sdapi = require('./sdapi');
+
+/**
+ * After hooks
+ * @return
+ */
+function afterHooks() {
+    this.After('@apitoken', () => {
```
FenrirUnbound: Sorry, I may have missed the initial conversation.

If you need to clean-up, you'll have to do it in the `Before`. The `After` is not guaranteed to execute if a scenario fails. This is why [some of the other features have](https://github.com/screwdriver-cd/screwdriver/blob/7948251dea91758236d95a282a3a008fa6b930eb/features/step_definitions/authorization.js#L10) a `Before` defined in them and associated via tag.

ian-fox: The one that cares about a potential conflict also calls the cleanup in `before`, my understanding is that this is just to leave the server in a cleaner state provided everything goes well.

ian-fox: Specifically [here](https://github.com/screwdriver-cd/screwdriver/pull/581/files#diff-8c0434ae40099e4cb24df13dea09ccdfR9)

ian-fox: It also looks like `cucumber` it will run after hooks even if a test fails.

PR #591, opened 2017-06-09
fix(545): Pin screwdriver-datastore-sequelize version

minz1027:
The previous rollback PR didn't work because 2.1.0 has the Breaking Change already but we failed to bump the version. Pin the last working version now till we figure out the sequelize problem. 

Last rollback PR: https://github.com/screwdriver-cd/screwdriver/pull/590
Releases for `screwdriver-datastore-sequelize`: https://github.com/screwdriver-cd/datastore-sequelize/releases

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "screwdriver-build-bookend": "^2.0.1",
     "screwdriver-config-parser": "^3.3.0",
     "screwdriver-data-schema": "^16.8.5",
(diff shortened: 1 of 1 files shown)


minz1027:
Problem solved by manually publish `datastore-sequelize` v2.1.1 without the Breaking Change introduced in v2.1.0.
https://github.com/screwdriver-cd/datastore-sequelize/pull/19

PR #596, opened 2017-06-09
User Tokens: Add test for GET route

ian-fox:
* Add the acceptance test for the token `GET` route
* Remove the `eslint-ignore`s because that rule got amended to ignore `Before` and `After` in https://github.com/screwdriver-cd/eslint-config-screwdriver/pull/23

### DIFF
--- a/features/api-token.feature
+++ b/features/api-token.feature
@@ -23,11 +23,10 @@ Feature: User API Token
         Then a valid JWT is received that represents "calvin"
         And the "tiger" token's 'last used' property is updated
 
-    @ignore
     Scenario: List API Tokens
-        And owns an existing API token
-        When "calvin" lists all their tokens
-        Then their API token is in the list
+        Given "calvin" owns an existing API token named "tiger"
+        When they list all their tokens
+        Then their "tiger" token is in the list
         And their token is safely described
 
     @ignore
--- a/features/step_definitions/api-token.js
+++ b/features/step_definitions/api-token.js
@@ -68,4 +68,51 @@ module.exports = function server() {
 
             Assert.notEqual(lastUsed, '');
         }));
+
+    this.Given(/^"calvin" owns an existing API token named "([^"]*)"$/, tokenName =>
+        request({
+            uri: `${this.instance}/${this.namespace}/tokens`,
+            method: 'POST',
+            auth: {
+                bearer: this.jwt
+            },
+            body: {
+                name: tokenName
+            },
+            json: true
+        }).then((response) => {
+            Assert.oneOf(response.statusCode, [409, 201]);
+        }));
+
+    this.When(/^they list all their tokens$/, () =>
+        request({
+            uri: `${this.instance}/${this.namespace}/tokens`,
+            method: 'GET',
+            auth: {
+                bearer: this.jwt
+            }
+        }).then((response) => {
+            Assert.strictEqual(response.statusCode, 200);
+
+            this.tokenList = JSON.parse(response.body);
+        }));
+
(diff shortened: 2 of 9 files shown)


features/step_definitions/api-token.js
```diff
@@ -68,4 +68,51 @@ module.exports = function server() {
...
+            },
+            body: {
+                name: tokenName
+            },
+            json: true
+        }).then((response) => {
+            Assert.oneOf(response.statusCode, [409, 201]);
+        }));
+
+    this.When(/^"calvin" lists all their tokens$/, () =>
```
r3rastogi: We should do a regex for the user's name, similar to how you did for the tokenName above

ian-fox: The only issue with that is that then the callback needs to have a parameter, otherwise cucumber will error.

There isn't really any logic that depends on the name though, so that leads to situations like [this](https://github.com/screwdriver-cd/screwdriver/blob/master/features/step_definitions/authorization.js#L35) where there's a `switch` statement that doesn't really do anything.

I felt like using a literal (as used in places like the [events](https://github.com/screwdriver-cd/screwdriver/blob/master/features/step_definitions/events.js#L55) tests) was probably cleaner.

ian-fox:
My bad on the PR title, forgot I had two commits so it wouldn't default to the commit message.

features/step_definitions/api-token.js
r3rastogi: Hmm ok. Maybe just make it generic, maybe something similar to this:

```
Given a user
And they own an existing API token named "tiger"
When they list their tokens
```

features/api-token.feature
```diff
@@ -23,11 +23,10 @@ Feature: User API Token
         Then a valid JWT is received that represents "calvin"
         And the "tiger" token's 'last used' property is updated
 
-    @ignore
     Scenario: List API Tokens
-        And owns an existing API token
-        When "calvin" lists all their tokens
-        Then their API token is in the list
+        Given the user owns an existing API token named "tiger"
```
r3rastogi: I'd strongly prefer if you changed "calvin" --> "the user" everywhere in this file, or just leave it as "calvin" everywhere. Consistency is key.

ian-fox: Makes sense. Putting it back to "calvin" then, as it's actually needed for the background step.

features/step_definitions/api-token.js
```diff
@@ -68,4 +68,51 @@ module.exports = function server() {
...
+
+            this.tokenList = JSON.parse(response.body);
+        }));
+
+    this.Then(/^their "([^"]*)" token is in the list$/, (tokenName) => {
+        const match = this.tokenList.find(token => token.name === tokenName);
+
+        Assert.isOk(match);
+
+        this.testToken = match;
```
tkyi: Do you need some sort of `return null;` statement for the step to finish?

features/step_definitions/api-token.js
```diff
@@ -68,4 +68,51 @@ module.exports = function server() {
...
+
+    this.Then(/^their token is safely described$/, () => {
+        const expectedKeys = ['id', 'name', 'lastUsed'];
+        const forbiddenKeys = ['hash', 'value'];
+
+        expectedKeys.forEach(property =>
+            Assert.property(this.testToken, property));
+
+        forbiddenKeys.forEach(property =>
+            Assert.notProperty(this.testToken, property));
```
tkyi: Same question, do you need some sort of `return` statement?

features/step_definitions/api-token.js
ian-fox: In javascript if the `return` is omitted, it will automatically return `undefined`

PR #601, opened 2017-06-12
feat(tokens): Token PUT acceptance test

ian-fox:
* Added a criterion to make sure `PUT`ting the token doesn't count as "using" it.
* Switch token update/refresh routes to use `PUT` instead of `PATCH`

### DIFF
--- a/features/step_definitions/api-token.js
+++ b/features/step_definitions/api-token.js
@@ -6,6 +6,10 @@ const request = require('../support/request');
 const sdapi = require('../support/sdapi');
 
 module.exports = function server() {
+    this.Before('@apitoken', () => {
+        this.testToken = null;
+    });
+
     this.Given(/^"calvin" does not own a token named "([^"]*)"$/, token =>
         // Ensure there won't be a conflict: delete the token if it's already there
         sdapi.cleanupToken({
@@ -82,6 +86,25 @@ module.exports = function server() {
             json: true
         }).then((response) => {
             Assert.oneOf(response.statusCode, [409, 201]);
+
+            if (response.statusCode === 201) {
+                this.testToken = response.body;
+
+                return null;
+            }
+
+            return request({
+                uri: `${this.instance}/${this.namespace}/tokens`,
+                method: 'GET',
+                auth: {
+                    bearer: this.jwt
+                }
+            }).then((listResponse) => {
+                Assert.strictEqual(listResponse.statusCode, 200);
+
+                this.testToken = JSON.parse(listResponse.body)
+                                     .find(token => token.name === tokenName);
+            });
         }));
 
     this.When(/^they list all their tokens$/, () =>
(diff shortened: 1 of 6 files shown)


ian-fox:
It looks like ember's `model.save()` [will use `PATCH` by default](https://guides.emberjs.com/v2.13.0/models/creating-updating-and-deleting-records/#toc_persisting-records), and I think it should be possible to [customize the REST adapter](https://guides.emberjs.com/v2.13.0/models/customizing-adapters/) if there's anywhere else that needs it.

That said, looking at [the controller for secrets](https://github.com/screwdriver-cd/ui/blob/master/app/pipeline/secrets/controller.js) and our current [REST adapter](https://github.com/screwdriver-cd/ui/blob/master/app/application/adapter.js) it looks like it should be using `PATCH` already, so I may be missing something somewhere.

features/step_definitions/api-token.js
```diff
@@ -115,4 +115,45 @@ module.exports = function server() {
...
+
+    this.Then(/^the token's 'last used' property will not be updated$/, () =>
+        request({
+            uri: `${this.instance}/${this.namespace}/tokens`,
+            method: 'GET',
+            auth: {
+                bearer: this.jwt
+            }
+        }).then((response) => {
+            const lastUsed = JSON.parse(response.body)
```
r3rastogi: Shouldn't you be getting and saving the lastUsed property before we make the the PATCH request, instead of reading it from the body?

ian-fox: Oh yeah, that'd be a lot simpler than making a whole extra `GET` request.

plugins/tokens/update.js
```diff
@@ -61,8 +61,6 @@ module.exports = () => ({
                             .then(() => {
                                 const output = token.toJson();
 
-                                delete output.hash;
```
ian-fox: no longer needed as this is handled by [Token.toJson](https://github.com/screwdriver-cd/models/blob/master/lib/token.js#L48)

ian-fox:
Ooh, I bet it isn't the UI that calls those methods. It's probably some other part of screwdriver, like an executor or something.

ian-fox:
I found the solution: the docs I was looking at were for the JSON adapter, whereas in the REST adapter it does use [PUT](https://github.com/emberjs/data/blob/v2.13.0/addon/adapters/rest.js#L1301).

features/support/before-hook.js
```diff
@@ -30,6 +31,18 @@ function promiseToWait(timeToWait) {
 }
 
 /**
+ * Generate a random string
+ * @method randomBytes
+ * @param {Number}      size    Number of bytes to generate
+ * @return {Promise}
+ */
+function randomBytes(size) {
```
r3rastogi: Would it be better to just give the new token a new name (`tiger2` or something like that) instead of calling an extra function/pulling in the `crypto` package?

features/step_definitions/api-token.js
```diff
@@ -115,4 +115,34 @@ module.exports = function server() {
...
+                    this.updatedToken = response.body;
+                });
+            }));
+
+    this.Then(/^their token will have that new label$/, () => {
+        Assert.strictEqual(this.updatedToken.description, this.newDescription);
+    });
+
+    this.Then(/^the token's 'last used' property will not be updated$/, () => {
+        Assert.strictEqual(this.updatedToken.lastUsed, this.testToken.lastUsed);
```
r3rastogi: Where are you saving the original `lastUsed` value?

ian-fox:
I did some further reading just to be sure, and it seems that neither PATCH nor PUT is quite right. PUT requires the full resource in the payload, PATCH requires a list of operations like   { "op": "replace", "path": "/a/b/c", "value": 42 }. Different people seem to use one or the other to accomplish "here are some keys and values, update them accordingly."

features/support/before-hook.js
ian-fox: In terms of dependencies, `crypto` is part of node. It doesn't have to be in the `package.json` or anything, it's just there.

If all of the tests are run in sequence then setting the description to be a literal value should work, however the setup `Given "calvin" owns an existing API token named "tiger"` doesn't actually set the description to anything, it just makes sure the token exists. So if the `update` test is called repeatedly without a `create` or `delete` between it, it'd keep setting the description to the same value.

That could be fixed either by making an `update` request in the setup step if the token is found to already be there, or making a `get` request as part of this step and then calling update with a value that's different than the one already there (like alternating between "tiger" and "not tiger" or something). I guess it depends what you think would be cleaner.

ian-fox: The `crypto` package is part of node itself, so it doesn't represent any new dependency. However, the fix for your second comment about the original `lastUsed` value will lead to a cleaner solution here I think.

features/step_definitions/api-token.js
ian-fox: Right now it's actually being set in the `list` and `create` tests. I'll modify this so that it'll still run correctly if only the `update` test gets run.

ian-fox: It's now saved as part of the `owns a token named "tiger"` step

PR #602, opened 2017-06-13
test(tokens): Token DELETE acceptance test

ian-fox:
Adding the acceptance test for revoking a token.

### DIFF
--- a/features/api-token.feature
+++ b/features/api-token.feature
@@ -25,19 +25,18 @@ Feature: User API Token
 
     Scenario: List API Tokens
         Given "calvin" owns an existing API token named "tiger"
-        When they list all their tokens
-        Then their "tiger" token is in the list
-        And their token is safely described
+        When he lists all his tokens
+        Then his "tiger" token is in the list
+        And his token is safely described
 
     Scenario: Edit API Token Labels
         Given "calvin" owns an existing API token named "tiger"
-        When they change the label associated with the token
-        Then their token will have that new label
+        When he changes the label associated with the token
+        Then his token will have that new label
         And the token's 'last used' property will not be updated
 
-    @ignore
     Scenario: Revoke API Token
-        And owns an existing API token
-        When "calvin" revokes the token
+        Given "calvin" owns an existing API token named "tiger"
+        When he revokes the token
         And the token is used to log in
         Then the login attempt fails
--- a/features/step_definitions/api-token.js
+++ b/features/step_definitions/api-token.js
@@ -8,6 +8,7 @@ const sdapi = require('../support/sdapi');
 module.exports = function server() {
     this.Before('@apitoken', () => {
         this.testToken = null;
+        this.loginResponse = null;
     });
 
     this.Given(/^"calvin" does not own a token named "([^"]*)"$/, token =>
@@ -46,14 +47,17 @@ module.exports = function server() {
             auth: {
                 bearer: this.jwt
             }
+        // Logging in is accomplished through this.getJwt
         }).then(() => this.getJwt(this.testToken.value).then((response) => {
-            Assert.strictEqual(response.statusCode, 200);
+            Assert.oneOf(response.statusCode, [200, 401]);
 
-            this.newJwt = response.body.token;
+            this.loginResponse = response;
         })));
 
     this.Then(/^a valid JWT is received that represents "calvin"$/, () => {
-        const decodedToken = jwt.decode(this.newJwt);
+        Assert.strictEqual(this.loginResponse.statusCode, 200);
+
+        const decodedToken = jwt.decode(this.loginResponse.body.token);
 
         Assert.strictEqual(decodedToken.username, this.username);
     });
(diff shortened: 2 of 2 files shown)


features/api-token.feature
```diff
@@ -35,9 +35,8 @@ Feature: User API Token
         Then their token will have that new label
         And the token's 'last used' property will not be updated
 
-    @ignore
     Scenario: Revoke API Token
-        And owns an existing API token
+        Given "calvin" owns an existing API token named "tiger"
         When "calvin" revokes the token
```
tkyi: ~~"calvin"~~ they

features/step_definitions/api-token.js
```diff
@@ -47,13 +48,15 @@ module.exports = function server() {
                 bearer: this.jwt
             }
         }).then(() => this.getJwt(this.testToken.value).then((response) => {
-            Assert.strictEqual(response.statusCode, 200);
+            Assert.oneOf(response.statusCode, [200, 401]);
```
r3rastogi: Why are both response codes valid?

features/step_definitions/api-token.js
```diff
@@ -167,4 +170,17 @@ module.exports = function server() {
...
+        request({
+            uri: `${this.instance}/${this.namespace}/tokens/${this.testToken.id}`,
+            method: 'DELETE',
+            auth: {
+                bearer: this.jwt
+            }
+        }).then(response => Assert.strictEqual(response.statusCode, 204)));
+
+    this.Then(/^the login attempt fails$/, () => {
+        Assert.strictEqual(this.loginResponse.statusCode, 401);
```
r3rastogi: Don't you need to POST a login attempt?

r3rastogi: Nevermind, didn't realize that step was written previously.

features/step_definitions/api-token.js
tkyi: I think depending on which test is run, those would be the two expected response codes. A `500` would not be good/expected. Not sure if there's a better way to do this, but we also do this in the authorization test:
https://github.com/screwdriver-cd/screwdriver/blob/master/features/step_definitions/authorization.js#L58

features/step_definitions/api-token.js
r3rastogi: Ok related to the above - https://github.com/screwdriver-cd/screwdriver/blob/35249f0112292ab502e8fbcabc57ce057c18d4c2/features/step_definitions/api-token.js#L45 aren't you logging out here instead of in?

features/step_definitions/api-token.js
ian-fox: Basically what Tiff said.

This step gets used by both the `create` test, and the `delete` test. In the `create` test we're logging in with a valid token, so we expect the response to be 200 (which is checked in the [step for that case](https://github.com/screwdriver-cd/screwdriver/blob/user-tokens/features/step_definitions/api-token.js#L56) too).

In the `delete` test, we're calling login with an invalid token, so we expect the response to be `401`. Again in that case it's further checked in the [next step](https://github.com/screwdriver-cd/screwdriver/blob/user-tokens/features/step_definitions/api-token.js#L183) to make sure we got the right one of the two.

features/step_definitions/api-token.js
ian-fox: Initially I had it just logging in, but I realized that if you do that it gets authenticated with the session from the background/setup step. By logging out before logging in with the token it ensures it's actually the token that's being used to log in.

r3rastogi: But where are you calling log in in that step then? I'm only seeing the POST to logout

ian-fox: Oh, that's hidden by `getJwt`. Maybe I should add a comment to that effect?

features/api-token.feature
```diff
@@ -35,9 +35,8 @@ Feature: User API Token
         Then their token will have that new label
         And the token's 'last used' property will not be updated
 
-    @ignore
     Scenario: Revoke API Token
-        And owns an existing API token
-        When "calvin" revokes the token
+        Given "calvin" owns an existing API token named "tiger"
+        When they revoke the token
```
d2lam: Wonder why it's changed from `calvin` to `they`? Shouldn't only calvin has the ability to revoke his own token?

ian-fox: In this case `calvin` and `they` are the same person.

tkyi: I know originally @r3rastogi commented that it was strange that we put `"calvin"` on all the lines when we weren't actually regex matching/using that value anywhere.

I talked to @petey and he said feature files are meant to be human-readable, and since he thinks `they` is easier to read than having `"calvin"`, I think we've decided to go ahead with `they`.

On a side note, maybe we could use `he` since that is more specific and makes it more obvious only `"calvin"` can revoke the token. This might just be bike shedding, so I'll leave it up to @ian-fox to use his best judgement here.

ian-fox: `He` would probably be clearer yeah. I'll swap that.

ian-fox: Left it as `they` in `authorization.feature` since there's a mix in there, but updated the `api-token.feature` to use he since there's no need to be overly general there.

d2lam: Yeah I think it's less confusing to use **"he"**, since it's implied "calvin" = "he".

PR #606, opened 2017-06-14
test(tokens): Token refresh acceptance test

ian-fox:
* fix some style mistakes
* calling the `refresh` returns things other than 500 now
* added acceptance test for `refresh` endpoint

### DIFF
--- a/features/step_definitions/api-token.js
+++ b/features/step_definitions/api-token.js
@@ -7,8 +7,9 @@ const sdapi = require('../support/sdapi');
 
 module.exports = function server() {
     this.Before('@apitoken', () => {
-        this.testToken = null;
         this.loginResponse = null;
+        this.testToken = null;
+        this.updatedToken = null;
     });
 
     this.Given(/^"calvin" does not own a token named "([^"]*)"$/, token =>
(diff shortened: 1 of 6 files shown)


features/step_definitions/api-token.js
```diff
@@ -68,11 +58,12 @@ module.exports = function server() {
...
                 bearer: this.jwt
-            }
+            },
+            json: true
         }).then((response) => {
-            const lastUsed = JSON.parse(response.body)
-                .find(token => token.name === tokenName)
-                .lastUsed;
+            const lastUsed = response.body
+                                     .find(token => token.name === tokenName)
```
petey: Why is this indented so far? Seemed ok before.

ian-fox: My instinct was to line the periods up, but it is pretty far. Perhaps one tab would look less weird.

ian-fox: The change was because I realized if I sent `json: true` I didn't need `JSON.parse`

plugins/tokens/update.js
```diff
@@ -58,11 +58,11 @@ module.exports = () => ({
                             });
 
                             return token.update()
-                            .then(() => {
-                                const output = token.toJson();
+                                .then(() => {
```
petey: You could probably just have the following to do the same thing:
```
.then(() => reply(token.toJson()));
```

ian-fox: Ah, I must have forgotten to refactor that when I got rid of the `delete token.hash`

PR #617, opened 2017-06-20
chore(tests): Upgrade cucumber to version 2

ian-fox:
## Context

* We've been pinned on cucumber 1.x, and since then a new major version has come out

## Objective

* Fix the tests to work with the new cucumber 2.x API

## Notes

* Can't use arrow functions for hooks/step definitions any more because of issues with `this` binding
* Renamed `before-hook.js` to `world.js` because it isn't really a hook.

## References

[Cucumber 2.3.1 docs](https://github.com/cucumber/cucumber-js/tree/master/docs/support_files)
[Why we can't use arrows](https://github.com/cucumber/cucumber-js/blob/master/docs/faq.md)


features/step_definitions/secrets.js
```diff
@@ -2,56 +2,70 @@
...
+                return request({
+                    uri: `${this.instance}/${this.namespace}/pipelines`,
+                    method: 'POST',
+                    auth: {
+                        bearer: this.jwt
+                    },
+                    body: {
+                        checkoutUrl: `git@github.com:${this.repoOrg}/${this.repoName}.git#master`
+                    },
+                    json: true
```
ian-fox: This stuff got moved here because [line 159](https://github.com/screwdriver-cd/screwdriver/pull/617/files#diff-2b66eab6945aba4b87fba63045f30482R159) uses the `secondJobId` that we find with this `GET` request.

It didn't need to be here in the past because it got set in another step and stuff didn't get cleaned up, but now cucumber rebuilds the world (`this`) object on each scenario, so `this.secondJobId` was undefined in that step.

ian-fox:
Something broke, hold on

features/support/world.js
```diff
@@ -0,0 +1,136 @@
...
+}
+
+/**
+ * Ensure a pipeline exists, and get its jobs
+ * @method ensurePipelineExists
+ * @param   {Object}    config
+ * @param   {String}    config.repoName     Name of the pipeline
+ * @return {Promise}
+ */
+function ensurePipelineExists(config) {
```
ian-fox: This is the only thing that's really new in this file: I found that this was defined in two different hooks.

May be cleaner to just modify the regex in one of them to catch both cases rather than defining it here, curious to hear @petey's opinion in particular on that.

features/step_definitions/secrets.js
```diff
@@ -68,51 +41,38 @@ module.exports = function server() {
...
-            Assert.equal(response.statusCode, 200);
+        }).then((resp) => {
+            Assert.equal(resp.statusCode, 201);
 
-            this.jobId = response.body[0].id;
-            this.secondJobId = response.body[1].id;
-            this.thirdJobId = typeof response.body[2] === 'object' ? response.body[2].id : null;
-            this.lastJobId = response.body.reverse().find(b => typeof b === 'object').id || null;
-        })
-        .then(() =>
```
ian-fox: Moved this bit to be part of the function that got moved to `world`, reason being that the `secondJobId` was set here, but used in a [step](https://github.com/screwdriver-cd/screwdriver/pull/617/files#diff-2b66eab6945aba4b87fba63045f30482R74) in a different scenario that ran afterwards.

In cucumber 2, the `world` object is rebuilt for every scenario, so it was undefined when it came time to actually use it.

features/support/world.js
ian-fox: It's used in both `secrets` and `events`, so if the regex does get changed to match both maybe there should be some sort of file like `shared-steps`

ian-fox:
WIP: fixing merge conflict

features/step_definitions/sd-step.js
```diff
@@ -20,77 +22,81 @@ module.exports = function server() {
...
-            request({
-                uri: `${this.instance}/${this.namespace}/builds`,
-                method: 'POST',
-                body: {
-                    jobId: this.jobId
-                },
-                auth: {
-                    bearer: this.jwt
-                },
+    When(/^the (main|tilde|hat|specify) job is started$/,
```
ian-fox: I'm not sure why this file is displaying weirdly in the diff view. This is the only significant change here, because with the `.*` expression this step was also applying to one from secrets, which cucumber doesn't know how to deal with.

ian-fox:
image

PR #618, opened 2017-06-20
feat(472): alternative default docker registry

### LINKED ISSUE #472: Alternative Default Docker Registry
> As a user, I want to change the default docker registry to my private registry, so that there are less human errors when selecting private images. 

Right now, Screwdriver just uses the default Docker implementation of `node:4 == hub.docker.com/r/_/node:4`.  For companies that run their own private registry, having to type out the private registry over and over again can lead to hard to detect human errors.

Screwdriver should support a way to specify a global default registry that translates `node:4` into the private registry.

Rules:
 - Configuration is made at the API layer.

FenrirUnbound:
## Context

Screwdriver uses the default Docker implementation of `node:4 => registry.hub.docker.com/r/library/node:4`. For companies that run their own private registry, it may be more beneficial for them to override that default behavior.

## Objective

This change allows for a global configuration where an image with a name like `node:4` translates into `private.registry.com/_/node:4`.

## References

* Resolves #472 
* Dependent on screwdriver-cd/models#175

### DIFF
--- a/config/custom-environment-variables.yaml
+++ b/config/custom-environment-variables.yaml
@@ -151,3 +151,5 @@ ecosystem:
     store: ECOSYSTEM_STORE
     # Badge service (needs to add a status and color)
     badges: ECOSYSTEM_BADGES
+    # Default registry to pull build containers from
+    dockerRegistry: ECOSYSTEM_DOCKER_REGISTRY
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -163,3 +163,5 @@ ecosystem:
     store: https://store.screwdriver.cd
     # Badge service (needs to add a status and color)
     badges: https://img.shields.io/badge/build-{{status}}-{{color}}.svg
+    # Default registry to pull build containers from. Uses Docker Hub if nothing/empty string is provided
+    dockerRegistry: ""
--- a/bin/server
+++ b/bin/server
@@ -78,6 +78,7 @@ const userFactory = Models.UserFactory.getInstance({
 });
 const buildFactory = Models.BuildFactory.getInstance({
     datastore,
+    dockerRegistry: ecosystem.dockerRegistry,
     scm,
     executor,
     bookend,


config/custom-environment-variables.yaml
```diff
@@ -151,3 +151,5 @@ ecosystem:
     store: ECOSYSTEM_STORE
     # Badge service (needs to add a status and color)
     badges: ECOSYSTEM_BADGES
+    # Default registry to pull build containers from
+    # dockerRegistry: ECOSYSTEM_DOCKER_REGISTRY
```
d2lam: Don't you need to uncomment this line in order for user to configure via environment variable?

config/default.yaml
```diff
@@ -162,3 +162,5 @@ ecosystem:
     store: https://store.screwdriver.cd
     # Badge service (needs to add a status and color)
     badges: https://img.shields.io/badge/build-{{status}}-{{color}}.svg
+    # Default registry to pull build containers from
+    dockerRegistry: ""
```
d2lam: I think we should comment the default is docker hub.

config/default.yaml
```diff
@@ -163,3 +163,5 @@ ecosystem:
     store: https://store.screwdriver.cd
     # Badge service (needs to add a status and color)
     badges: https://img.shields.io/badge/build-{{status}}-{{color}}.svg
+    # Default registry to pull build containers from. Uses Docker Hub if nothing/empty string is provided
+    dockerRegistry: ""
```
ian-fox: Where is the logic that defaults to dockerhub? Would it make more sense to just have dockerhub in the default.yaml?

FenrirUnbound: I think the main issue captures it well (https://github.com/screwdriver-cd/screwdriver/issues/472), but I'll see if I can clarify it.

If you have an image `screwdrivercd/screwdriver:latest`, Docker needs to know which registry to pull the image from. When no registry is specified (like in this example), it will default to using `registry.hub.docker.com`. So when you say `screwdrivercd/screwdriver:latest`, it's implied that the registry is Docker Hub.

If you want to pull the image from a different registry, you would add the prefix to it, like `private.registry.com/screwdrivercd/screwdriver:latest`. Then, when the build starts, it will pull the image from `private.registry.com` instead of Docker Hub.

The feature this introduces is that if this setting is not configured, it will use the default Docker behavior (e.g., pull images from the implied registry). If you want all builds to pull images from a non-Docker Hub registry by default, you would configure this to your specific registry.

ian-fox: Ah, I see now. Thanks!

Issue #597, opened 2017-06-09
Kubernetes Pod template for supporting VMs


FenrirUnbound:
For supporting the effort in having VMs available to builds in Kubernetes, we need to re-visit our Pod template structure. The current pod template for `executor-k8s` can be [found here](https://github.com/screwdriver-cd/executor-k8s/blob/master/config/pod.yaml.tim). 

For VMs, we plan on using [Hypercontainer](https://hypercontainer.io/) for allocating VMs in a Kubernetes cluster.

Based on the current PoC, this is what the template would look similar to:

```
spec:
  privileged: true
  container:
    command: hyperctl -a --rm --name build123 <other arguments> -p pod.yaml
  resources:
    - 10GB
    # 2CPUs in micro-CPU
    - 2000m  
```

Hyperctl references a `pod.yaml` for defining what is running within the VM environment. Based on what we have today, this is a draft of what it may look like:

```
spec:
  containers:
    - name: build
      image: <user_defined_container>
      command: <launcher command>
    - name: docker
      command: dockerd -Htcp://<host_ip>:<port>
    # might not need this one
    - name: logservice
      image: ...
      command: ...
```

## Reference

* [K8s pod templates](https://kubernetes.io/docs/concepts/workloads/pods/pod-overview/#pod-templates)
* [Hypercontainer](https://hypercontainer.io/)
   * [Hyperctl run command](https://docs.hypercontainer.io/reference/run.html)
   * [Hyperctl pod spec](https://docs.hypercontainer.io/reference/podfile.html)
* Switching Log Service implementation -- #598

PR #625, opened 2017-06-27
feat(597): add config for executor-k8s-vm

minz1027:
# Context

To allow build to run inside a VM in Kubernetes, we need to add the config for executor-k8s-vm.

## Objective

To allow build to run inside a VM in Kubernetes, we need to add the config for executor-k8s-vm.

## References
Block by https://github.com/screwdriver-cd/executor-k8s-vm/pull/3

Related to https://github.com/screwdriver-cd/screwdriver/issues/597


config/default.yaml
```diff
@@ -95,8 +95,8 @@ executor:
             # The host or IP of the kubernetes cluster
             host: kubernetes.default
             # The jwt token used for authenticating kubernetes requests
-            # Loaded from /var/run/secrets/kubernetes.io/serviceaccount/token by default
-            # token:
+            # Loaded from /var/run/secrets/kubernetes.io/serviceaccount/token by default if not specified
+            token: 'fakeToken'
```
minz1027: Add this so that it won't break when you run `npm start` locally. `executor-router` will require all executor plugins no matter what default you pick. So when I test it locally using `docker` as my default, it complains that ` /var/run/secrets/kubernetes.io/serviceaccount/token` does not exist which is annoying and does not make any sense.

FenrirUnbound: The issue we're going to have is when we run this in the hosted cluster. We want the cluster to read from `/var/run/secrets/kubernetes.io/serviceaccount/token`.

Doesn't this change mean we have to now add another configuration change to override this fake token?

minz1027: @FenrirUnbound That's a good point! Changed the k8s executor to not complain if no token passed in and path not exists.
https://github.com/screwdriver-cd/executor-k8s/pull/61

Issue #365, opened 2016-11-21
Ability to use multiple scms


catto:
It would be nice if we could use multiple scms such as GitHub.com and GitHub Enterprise.
Do you have any plan to achieve it?

d2lam:
@catto  This should be already working. To use GHE, you need to add `SCM_GITHUB_GHE_HOST` environment variable. Does this documentation help: http://screwdriver.cd/getting-started/plugins/#source-control

stjohnjohnson:
@d2lam I think @catto is looking for the ability to support multiple SCMs at one time.

catto:
Ah, yes, I meant to say about the ability to use the scms at one time like below
```yaml
scms:
    - github.com:
      plugin: github
      oauthClientId: githubclientid
      oauthClientSecret: githubclientsecret
      secret: githubsecret
    - ghe:
      plugin: github
      oauthClientId: gheclientid
      oauthClientSecret: gheclientsecret
      gheHost: github.mysite.org
      secret: ghesecret
```

stjohnjohnson:
Right now our recommendation is going to be to run multiple clusters.  However, we'd be open to hear any ideas you have around how to make this kind of change (as it would touch all parts of SD).

catto:
@stjohnjohnson Thank you, I got it. I'll think about it and let you know if I find the way how to achieve it.

catto:
I had think about this feature, and I came up with one idea.
It uses an authentication context and switches the context completely between multiple scms.

The abstract is as follows.
* Cluster admin configures multiple scms to their cluster, like [this](https://github.com/screwdriver-cd/screwdriver/issues/365#issuecomment-262134135).
* Screwdriver shows multiple login buttons for each scm in the login page.
* When an authentication is completed, Screwdriver issues a session cookie for a user with an auth context.
    * This can be "github.com:catto" or "my-ghe:catto"
    * It is better to show a user name with an auth context in the header.
* If a user wants to use other scm, the user need to re-login to the desired scm.

This method would require major changes to the user model, but it seems that other ones don't need to do so much.

stjohnjohnson:
I like this @catto. How do we want to detect which scm_url works for which scm module?

catto:
@stjohnjohnson How about this Façade-like design?
* Implement and expose a function such as `canHandle` which can determine if it can handle the specified scm_url or hostname, in each scm module. Or, implement a detector directly into `scm-base`.
[...]

PR #632, opened 2017-07-05
feat(365) Add functional test for multiple scms

s-yoshika:
## Context

For the support multiple SCMs, Screwdriver has to determine the user by not only `username` but also `scmContext`.

## Objective

This PR add a functional test which check the user profile by `username` and `scmContext`.

## References

issue: https://github.com/screwdriver-cd/screwdriver/issues/365
data-schema change: https://github.com/screwdriver-cd/data-schema/pull/152

### DIFF
--- a/features/step_definitions/authorization.js
+++ b/features/step_definitions/authorization.js
@@ -34,6 +34,10 @@ defineSupportCode(({ Before, Given, Then }) => {
             case 'calvin':
                 Assert.strictEqual(decodedToken.username, this.username);
                 break;
+            case 'github:calvin':
+                Assert.strictEqual(decodedToken.username, this.username);
+                Assert.strictEqual(decodedToken.scmContext, this.scmContext);
+                break;
             default:
                 return Promise.resolve('pending');
             }
--- a/features/support/world.js
+++ b/features/support/world.js
@@ -100,6 +100,7 @@ function CustomWorld({ attach, parameters }) {
     this.instance = `${this.protocol}://${process.env.SD_API}`;
     this.testOrg = process.env.TEST_ORG;
     this.username = process.env.TEST_USERNAME;
+    this.scmContext = process.env.TEST_SCM_CONTEXT;
     this.namespace = 'v4';
     this.promiseToWait = time => promiseToWait(time);
     this.getJwt = accessKey =>
--- a/features/authorization.feature
+++ b/features/authorization.feature
@@ -58,3 +58,10 @@ Feature: Authorization
         And they update the checkoutUrl
         Then the pipeline checkoutUrl is updated
         And the pipeline has the same id as before
+
+    @ignore
+    Scenario: SCM Context
+        And "github:calvin" is logged in
+        Then they can see the pipeline
+        And they can start the "main" job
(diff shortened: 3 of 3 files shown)


features/support/world.js
```diff
@@ -100,6 +100,7 @@ function CustomWorld({ attach, parameters }) {
     this.instance = `${this.protocol}://${process.env.SD_API}`;
     this.testOrg = process.env.TEST_ORG;
     this.username = process.env.TEST_USERNAME;
+    this.scmContext = process.env.TEST_SCM_CONTEXT;
```
tk3fftk: We should add `TEST_SCM_CONTEXT`.
https://github.com/screwdriver-cd/screwdriver/blob/5e014e0c61855edf585ac009f254f08695224140/screwdriver.yaml#L55
https://github.com/screwdriver-cd/screwdriver/blob/5e014e0c61855edf585ac009f254f08695224140/screwdriver.yaml#L83

s-yoshika: Yes, If screwdriver.cd will be enable this feature in the future, we should set the setting.

features/step_definitions/authorization.js
```diff
@@ -34,6 +34,10 @@ defineSupportCode(({ Before, Given, Then }) => {
             case 'calvin':
                 Assert.strictEqual(decodedToken.username, this.username);
                 break;
+            case 'github:calvin':
+                Assert.strictEqual(decodedToken.username, this.username);
+                Assert.strictEqual(decodedToken.scmContext, this.scmContext);
```
minz1027: So we will add the `scmContext` to be part of the jwt token profile later?
https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/auth/index.js#L57

s-yoshika: > So we will add the scmContext to be part of the jwt token profile later?

Yes, if we support multiple SCMs feature, we can't determine the user only by username, so I think jwt token profile should also contain `scmContext`.

> And how can we tell what scmContext the user is on when they first log in?

I think we can tell `scmContext` by API path.
Currently, the path of the login API is `/auth/login/{web?}`.
https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/auth/login.js#L15

For multiple SCMs, we have to change the API path like as `/auth/login/{scmContext?}/{web?}`.
`{scmContext}` parameter specifies the scm by which user want to log-in.

minz1027: Nice! Thx for the explanation. Can we document these in the issue?

features/support/world.js
minz1027: Need to add it to the readme too. https://github.com/screwdriver-cd/screwdriver/blob/master/README.md#functional-tests

features/step_definitions/authorization.js
s-yoshika: Sorry, I'll update the issue :bow:

s-yoshika:
I documented the details for the support multiple SCMs.
https://github.com/screwdriver-cd/screwdriver/issues/365#issuecomment-313277401

s-yoshika:
rebased origin master

PR #634, opened 2017-07-05
Pin scm-github version to not have habitat

minz1027:
Habitat server is down. 

Roll scm-github version back  to not have habitat 
https://github.com/screwdriver-cd/scm-github/commit/625077368519cfe57769afe4f925d831f44ceab7


FenrirUnbound:
Build will always fail because Habitat is down.

PR #638, opened 2017-07-07
feat(tokens): Remove temporary access user

ian-fox:
## Context

Previously, for functional tests, we used a "temporary access key" and "temporary access user" to bypass OAuth. Now that tokens are supported, that is no longer necessary.

## Objective

Tokens were already functional, this PR removes the logic around looking for the temporary access key.

In switching to using an actual token, you will need a `sd-buildbot` user in your local datastore to run local functional tests. This can be accomplished with the new `create-test-user` script to generate the user from a git token.

## References

https://github.com/screwdriver-cd/screwdriver/issues/532


d2lam:
Also https://github.com/screwdriver-cd/screwdriver/blob/master/config/default.yaml#L23-L26

minz1027:
Probably should rename the ACCESS_KEY to API_TOKEN or ACCESS_TOKEN:
https://github.com/screwdriver-cd/screwdriver/blob/master/screwdriver.yaml#L58
The name is confusing

tkyi:
And we'll have to update the pipeline secret in Screwdriver with the new value

ian-fox:
Working on updating it to `API_TOKEN`

ian-fox:
While trying to run the functional tests on this I realized that you need to have the `sd-buildbot` user in your local database with the change to using a regular token instead of a special `temporaryAccessUser` one.

I'm working on a script that will add the sd-buildbot user to your local datastore if you have the git token for it so that we don't need to bother St.John to 2FA each time somebody sets screwdriver up for development.

ian-fox:
Added the script.

Seeing a bug while running the local functional tests about "The header content contains invalid characters," once I figure that out I'll remove the DNM.

ian-fox:
Looks like the bug came from my local config, not from the changes!

create-test-user.js
```diff
@@ -0,0 +1,71 @@
...
+    datastore,
+    scm,
+    password: authConfig.encryptionPassword
+});
+const tokenFactory = Models.TokenFactory.getInstance({
+    datastore
+});
+
+const username = process.argv[2];
+const gitToken = process.argv[3];
```
minz1027: magic number 2 and 3?
Should probably document how to use this script.

create-test-user.js
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env node
```
d2lam: Is this meant to be run manually? Where is the README for this?

create-test-user.js
ian-fox: Oh whoops I used to have a usage block, must have missed that when I moved it from a standalone to part of the SD repo!

features/scripts/readme.md
```diff
@@ -0,0 +1,19 @@
+## Scripts
+
+Scripts used for setting up tests
```
d2lam: `functional tests`

features/scripts/readme.md
```diff
@@ -0,0 +1,19 @@
...
+
+As a consequence of this, running the functional tests now requires the test user to be in your local database, and an API token for that user in your `func_config` file.
+
+The script `create-test-user` will create the required user in your database. Run this script with the username of the test user and their SCM access token, and the script will create the user and print an API token for use in tests to `stdout`.
+
+```bash
+$ npm run create-test-user -- test-user git-access-token
+```
+
+Put the token generated by the script in your `.func_config` file (or set it as an environemnt variable) as `API_TOKEN`.
```
d2lam: typo environment

features/scripts/readme.md
```diff
@@ -0,0 +1,19 @@
+## Scripts
+
+Scripts used for setting up tests
+
+### Removing the Temporary Access User
```
d2lam: Do we need to say this? It might be confusing for users when we go to major version `1`. I feel like it's not necessary to mention previous behaviors.

Issue #523, opened 2017-04-13
Favorites display on homepage


petey:
Users should be able to add a set of pipelines to a sharable collection, and display this collection on the main landing page for screwdriver.

petey:
We are still working out the logistics of the UI for this feature.

petey:
The current designs for this is to create collections based on pipeline keywords (#571). 

Collections:
- [ ] crud API for collections {id, userid, name, description, keywords, isDefault} 
- [ ] collections flyout UI panel to create/edit/remove collections

Dashboard:
- [ ] Display pipelines from current default collection
   - [ ] pipeline name
   - [ ] last event status
   - [ ] open/failing PRs
- [ ] Sortable pipelines, by health, last event time, custom?

TODO:
- [ ] Sharing collections with other users
- [ ] Collections of pipeline ids vs. pipeline keywords

![edit collections](https://user-images.githubusercontent.com/78533/27356398-2694102e-55c3-11e7-877f-ec96d074bbf0.png)

petey:
Initial tests with the current API responses requires about 155 requests to fetch data for 46 pipeline. This needs to be remedied by storing more information in events (last build status), or creating a separate API, maybe under collections, that aggregates the necessary data in one call.

tkyi:
# User Research Update

## Summary

After conducting user research/observations about dashboards, @joelseq and I have come to some conclusions.

### Observations

- keywords - not as useful as anticipated
  - they're cumbersome - require making code changes (in the screwdriver.yaml) in order to update them
  - there's too much overhead - they're difficult to keep track of and require the entire team to do
- amount of detail
  - there's usually too much detail displayed on a dashboard for an everyday developer
    - last build or build status of last run is sufficient information for a dashboard
    - if they want more information, they would deep dive on the pipeline page
- use cases
  - dashboards shouldn't really need to be checked everyday to find out if pipelines are healthy or not
    - developers are more concerned with the health of their deployed apps vs pipelines
    - developers should only be concerned with pipeline/build health if something fails; if something fails, it should notify the developer through Slack or email
  - dashboards would be good places for pipeline overviews/subway maps; they would be even more useful/powerful if they had actionable buttons in place
- searching
  - it would be nice if search results returned results sorted by relevancy - so pipelines the developer has contributed to in general/recently
- grouping
  - would be nice to be able to better group favorited pipelines, otherwise developers would have to scroll through way too many Pipelines to find the one they want

### Conclusion

[...]

PR #641, opened 2017-07-11
feat(523): Add CREATE route for collections

joelseq:
## Context

For the collections feature, we need to have a CREATE route for the creation of new collections

## Objective

This PR adds a plugin for collections with a CREATE route. With this change, users should now be able to create collections through the API.

## References

issue: #523


test/plugins/collections.test.js
```diff
@@ -0,0 +1,170 @@
...
+
+    after(() => {
+        mockery.disable();
+    });
+
+    it('registers the plugin', () => {
+        assert.isOk(server.registrations.collections);
+    });
+
+    describe('POST /collections', () => {
```
ian-fox: Username/collection name is a unique way of identifying collections right?

Probably should have a test making sure you can't make more than one with the same name/user.

joelseq: I discussed that with @minz1027 earlier and we came to the conclusion that the since the Model is mocked, the only way to test that functionality out is in the data schema and data store end and not the API directly.

test/plugins/collections.test.js
```diff
@@ -0,0 +1,170 @@
...
+    return mock;
+};
+
+const getUserMock = (user) => {
+    const mock = hoek.clone(user);
+
+    return mock;
+};
+
+// const collectionsGetterMock = collections => Promise.resolve(collections);
```
petey: Get rid of commented code.

test/plugins/collections.test.js
petey: Sounds like you need a functional test.

PR #642, opened 2017-07-12
fix(tokens): don't show tokens in server logs

ian-fox:
## Context

Currently if a user makes a GET request to `/v4/auth/token?api_token=(api token here)` while they have a session, the server logs will display the arguments that the path got called with, exposing the token.

## Objective

This PR adds a transform stream to the `hapi/good` list of streams that filters out anything that looks like one of our API tokens from the logs.

## References

https://github.com/hapijs/good/blob/master/API.md#options

I did a quick search, I don't think we had an issue for this.

### DIFF
--- a/test/plugins/tokens.test.js
+++ b/test/plugins/tokens.test.js
@@ -1,12 +1,16 @@
 'use strict';
 
-const assert = require('chai').assert;
-const sinon = require('sinon');
+const { assert } = require('chai');
 const hapi = require('hapi');
 const mockery = require('mockery');
+const { PassThrough } = require('stream');
+const sinon = require('sinon');
+const suppressAPITokens = require('../../plugins/tokens/filter');
 const urlLib = require('url');
+
 const testToken = require('./data/token.json');
-const testTokenWithValue = Object.assign({}, testToken, { value: '1234' });
+const testValue = '1234123412341234123412341234123412341234123';
+const testTokenWithValue = Object.assign({}, testToken, { value: testValue });
 
 delete testTokenWithValue.hash;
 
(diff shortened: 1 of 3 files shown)


ian-fox:
Forgot to add unit test, adding that now.

test/plugins/tokens.test.js
```diff
@@ -141,7 +144,7 @@ describe('token plugin test', () => {
         });
 
         it('returns 201 and correct token data', () => {
-            tokenMock = getTokenMock(Object.assign({}, testToken, { value: '1234' }));
+            tokenMock = getTokenMock(Object.assign({}, testToken, { value: testValue }));
```
joelseq: Is it possible to just use `testTokenWithValue` here or do `getTokenMock(Object.assign({}, testTokenWithValue)` ?

ian-fox: Good call

test/plugins/tokens.test.js
```diff
@@ -436,4 +440,19 @@ describe('token plugin test', () => {
...
+    describe('Logging suppresses API tokens', () => {
+        it('does not print API tokens in logs', () => {
+            const source = new PassThrough({ objectMode: true });
+            const result = new PassThrough({ objectMode: true });
+
+            source.write(`This is a string with a token in it! ${testTokenWithValue.value}`);
+
+            source.pipe(suppressAPITokens).pipe(result);
+
+            result.on('data', (chunk) => {
```
petey: Typically streams are asynchronous, and this looks like an event listener. What are you doing to ensure that the test function isn't completing before the assertion happens? With mocha, this is typically accomplished by having a `done` callback in the test definition, which is called in the test when it should actually be completed.

ian-fox: Good call! That was exactly what was happening, I had a mistake in my regex that I didn't even notice.

Issue #616, opened 2017-06-19
Templates - Second Pass


petey:
*Background* See https://github.com/screwdriver-cd/screwdriver/issues/470

- [x] Wrapping around the steps inside template. For example: pretest, posttest. (https://github.com/screwdriver-cd/config-parser/pull/39)
Inside template:
```yaml
   steps: 
       - test: echo test
       - publish: echo publish
```

Inside user's screwdriver.yaml:
```yaml
main:
   template: mytemplate@1
   steps: 
       - pretest: echo hello
       - posttest: echo goodbye
```

Translates into: 
```yaml
   steps:
       - pretest: echo hello
       - test: echo test
       - posttest: echo goodbye
       - publish: echo publish
```

- [x] Override step: (https://github.com/screwdriver-cd/config-parser/pull/40)
```yaml
main:
   template: mytemplate@1
   steps: 
       - publish: echo not publishing
```

Translate into:
```yaml
   steps:
       - test: echo test
       - publish: echo not publishing
```

- [x] [Version tags](https://github.com/screwdriver-cd/screwdriver/issues/616#issuecomment-314568245) (__WIP__): crud api for tagging template versions, e.g. npm's `dist_tags` allow fetch on template version by tag. `mytemplate@latest` (https://github.com/screwdriver-cd/data-schema/pull/157, https://github.com/screwdriver-cd/data-schema/pull/158, https://github.com/screwdriver-cd/models/pull/179, https://github.com/screwdriver-cd/config-parser/pull/41, https://github.com/screwdriver-cd/screwdriver/pull/646, https://github.com/screwdriver-cd/screwdriver/pull/647)
- [ ] Restrict API calls for template writes to admins, and template owners
- [ ] Version Lifespan: After a set period of time [configurable: default 6 months], a given version of a template should display a warning message in the build that the template is no longer being updated/maintained. (https://github.com/screwdriver-cd/screwdriver/issues/1490)
- [x] [Template namespaces](https://github.com/screwdriver-cd/screwdriver/issues/926) (e.g.: chef/publish, screwdriver/checkout, etc) 
- [ ] Template keywords: Listed in annotations of template.yaml, e.g. `language:node.js`, `publishing`, `aws`. Primarily to allow discovery of templates by searching by keyword. Keywords should apply to all versions of a template.
- Documentation in guide:
   - [x] create, validate, and publish (https://github.com/screwdriver-cd/guide/pull/90)
[...]

PR #646, opened 2017-07-14
feat(616): Can get a template using template name and tag

tkyi:
## Context

Users should be able to get a template by using the endpoint `/templates/{name}/{version}` or `/templates/{name}/{tag}`.

## Objective

This PR will modify the original `/templates/{name}/{version}` endpoint to also accept `{tag}` instead of `{version}`.

_Not sure if there's a better name for `versionOrTag`.._

## References

- Original issue: https://github.com/screwdriver-cd/screwdriver/issues/616
- Blocked by: https://github.com/screwdriver-cd/models/pull/179
- Related to: https://github.com/screwdriver-cd/screwdriver/pull/647

### DIFF
--- a/plugins/templates/get.js
+++ b/plugins/templates/get.js
@@ -5,27 +5,59 @@ const joi = require('joi');
 const schema = require('screwdriver-data-schema');
 const getSchema = schema.models.template.get;
 const baseSchema = schema.models.template.base;
+const versionRegex = schema.config.regex.VERSION;
+const versionSchema = joi.reach(baseSchema, 'version');
 
 module.exports = () => ({
     method: 'GET',
-    path: '/templates/{name}/{version}',
+    path: '/templates/{name}/{versionOrTag}',
     config: {
-        description: 'Get a single template',
+        description: 'Get a single template given template name and version or tag',
         notes: 'Returns a template record',
         tags: ['api', 'templates'],
         handler: (request, reply) => {
-            const factory = request.server.app.templateFactory;
+            const versionOrTag = request.params.versionOrTag;
+            const templateFactory = request.server.app.templateFactory;
 
-            return factory.getTemplate({
-                name: request.params.name,
-                version: request.params.version
-            }).then((template) => {
-                if (!template) {
-                    throw boom.notFound('Template does not exist');
+            // check if version or tag
+            const isVersion = versionOrTag.match(versionRegex);
+
+            return new Promise((resolve, reject) => {
+                // if tag, get template tag version
+                if (!isVersion) {
+                    const templateTagFactory = request.server.app.templateTagFactory;
+
+                    return templateTagFactory.get({
+                        name: request.params.name,
+                        tag: request.params.versionOrTag
+                    })
+                    .then((templateTag) => {
+                        if (!templateTag) {
+                            return reject(boom.notFound(`Template ${request.params.name} ` +
+                                `does not exist with tag ${request.params.versionOrTag}`));
+                        }
+
+                        return resolve(templateTag.version);
+                    });
                 }
 
-                return reply(template);
+                // otherwise just return the version
+                return resolve(versionOrTag);
             })
+            .then(version =>
+                // get the template
+                templateFactory.getTemplate({
+                    name: request.params.name,
+                    version
+                }).then((template) => {
+                    if (!template) {
+                        throw boom.notFound(`Template ${request.params.name} ` +
+                            `does not exist with version ${version}`);
+                    }
+
+                    return reply(template);
+                })
+            )
             .catch(err => reply(boom.wrap(err)));
         },
         response: {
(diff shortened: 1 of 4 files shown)


test/plugins/templates.test.js
```diff
@@ -178,6 +208,18 @@ describe('template plugin test', () => {
             });
         });
 
+        it('returns 404 when template tag does not exist', () => {
```
r3rastogi: We should probably add a test for returning a 404 if the version does not exist

plugins/templates/get.js
```diff
@@ -5,27 +5,59 @@ const joi = require('joi');
...
+                    const templateTagFactory = request.server.app.templateTagFactory;
+
+                    return templateTagFactory.get({
+                        name: request.params.name,
+                        tag: request.params.versionOrTag
+                    })
+                    .then((templateTag) => {
+                        if (!templateTag) {
+                            // eslint-disable-next-line max-len
+                            return reject(boom.notFound(`Template ${request.params.name} does not exist with tag ${request.params.versionOrTag}`));
```
d2lam: Why not make the line shorter so we don't need the `disable max-len`

test/plugins/templates.test.js
tkyi: When the version does not exist, the template will not exist. So that use case should cover that.

plugins/templates/get.js
tkyi: I forgot how without adding a newline 😢

tkyi: updated

plugins/templates/get.js
```diff
@@ -5,27 +5,59 @@ const joi = require('joi');
...
-                name: request.params.name,
-                version: request.params.version
-            }).then((template) => {
-                if (!template) {
-                    throw boom.notFound('Template does not exist');
+            const versionOrTag = request.params.versionOrTag;
+            const templateFactory = request.server.app.templateFactory;
+
+            // check if version or tag
+            const isVersion = versionOrTag.match(versionRegex);
```
FenrirUnbound: Oh, this is interesting. Does this mean that a `tag` can never be of a similar format to a version?

plugins/templates/get.js
```diff
@@ -5,27 +5,59 @@ const joi = require('joi');
...
+                                `does not exist with tag ${request.params.versionOrTag}`));
+                        }
+
+                        return resolve(templateTag.version);
+                    });
                 }
 
-                return reply(template);
+                // otherwise just return the version
+                return resolve(versionOrTag);
```
FenrirUnbound: We don't need to use a `new Promise` for this. We can instead use `return Promise.resolve()` for this operation.

plugins/templates/get.js
tkyi: Hmm.. good point. Shouldn't be..

FenrirUnbound: > Do we have any formal requirements for what a tag should/shouldn't be?

I'm unsure. 

Because of the way that the regex is formatted, the user cannot use a tag that is only composed of numbers. So this means that a user cannot use a simplified date stamp (e.g., `20170731`) or time (e.g., a unix timestamp like `1500491037`).

tkyi: Hmm should the regex be revised to only allow versions with 1 decimal point or two (allowing the single large number)?

plugins/templates/get.js
tkyi: Wasn't sure how to write it so I don't have to write lines 47-61 twice.

FenrirUnbound: Could do it a couple of ways. One way would be:

```
const isVersion = versionOrTag.match(versionRegex);

if (isVersion) {
  return Promise.resolve(versionOrTag);
}

const templateTagFactory = ...;
return templateTagFactory.get({...});
// rest of logic here...
```

Another way would be:

```
const isVersion = versionOrTag.match(versionRegex);

if (!isVersion) {
  const templateTagFactory = ...;
  return templateTagFactory.get({...});
  // rest of logic here...
}

return Promise.resolve(versionOrTag);
```

plugins/templates/get.js
FenrirUnbound: > we should change it so tags have to start with an alpha character.

I like that resolution. It's a lot more distinct that it's a tag 😸

plugins/templates/get.js
FenrirUnbound: Talked about it offline. I offered the alternative way to write it, and I'm ok with how it is currently written

d2lam:
I think it's ok to merge this now. But we should refactor this logic to models here: https://github.com/screwdriver-cd/models/blob/master/lib/templateFactory.js#L91

The `config-parser` is calling that function to get the template, so we need to modify it (since we won't be using `labels` anymore). Instead, it needs to take in a version or tag. And since it's basically the same logic. It makes sense to push the logic there.

PR #647, opened 2017-07-14
feat: Add create and remove template tag routes

d2lam:
## Context
This PR allows tagging a template, similar to how docker tagging works.
- 1 tag can only map to 1 version
- 1 version can have multiple tags

Example:

| id | name | tag | version | 
|---|---|---|---|
| 1 | mytemplate | stable | 1.2.0 | 
| 2 | mytemplate | latest | 1.2.0 | 

## Objective
Add 2 routes for template tagging: CREATE and DELETE. 
UPDATE uses the same route as CREATE: send a `PUT templates/{templateTag}/tags/{tagName}` with the payload as: `{version}` instead of the usual `/id`, since it doesn't make sense that users need to look up the `id` to update.

DELETE: `DELETE templates/{templateTag}/tags/{tagName}`

**Notes:**  This only allows `build` scope to tag template due to security reasons. It will check if the build belongs to the pipeline that publishes this template. If not, then the build is not allowed to tag the template. 

## References
https://github.com/screwdriver-cd/screwdriver/issues/616

### DIFF
--- a/plugins/templates/README.md
+++ b/plugins/templates/README.md
@@ -27,18 +27,18 @@ server.register({
 
 ### Routes
 
-#### Get all templates
-`page` and `count` optional
+#### Template
+##### Get all templates
 
-`GET /templates?page={pageNumber}&count={countNumber}`
+`GET /templates`
 
-#### Get a single template
+##### Get a single template
 
 You can get a single template by providing the template name and the specific version or the tag.
 
 `GET /templates/{name}/{tag}` or `GET /templates/{name}/{version}`
 
-**Arguments**
+###### Arguments
 
 'name', 'tag' or 'version'
 
@@ -46,18 +46,16 @@ You can get a single template by providing the template name and the specific ve
 * `tag` - Tag of the template (e.g. `stable`, `latest`, etc)
 * `version` - Version of the template
 
-#### Create a template
-Creating a template will store the template data (`config`, `name`, `version`, `description`, `maintainer`, `labels`) into the datastore.
+##### Create a template
+Creating a template will store the template data (`config`, `name`, `version`, `description`, `maintainer`) into the datastore.
 
-If the exact template and version already exist, the only thing that can be changed is `labels`.
-
-If the template already exists but not the version, the new version will be stored provided that the build has correct permissions.
+`version` will be auto-bumped. For example, if `mytemplate@1.0.0` already exists and the version passed in is `1.0.0`, the newly created template will be version `1.0.1`. 
 
 *Note: This endpoint is only accessible in `build` scope and the permission is tied to the pipeline that first creates the template.*
 
 `POST /templates`
 
-**Arguments**
+###### Arguments
 
 'name', 'version', 'description', 'maintainer', labels
 
@@ -84,25 +82,23 @@ Example payload:
 }
 ```
 
-#### Create/Update a tag for a template version
+#### Template Tag
+Template tag allows fetching on template version by tag. For example, tag `mytemplate@1.1.0` as `stable`. 
 
-Tagging a template version allows fetching on template version by tag. For example, tag `mytemplate@1.1.0` as `stable`. 
+##### Create/Update a tag
+
+If the template tag already exists, it will update the tag with the new version. If the template tag doesn't exist yet, this endpoint will create the tag.
 
 *Note: This endpoint is only accessible in `build` scope and the permission is tied to the pipeline that creates the template.*
 
-`PUT /templates/tags` with the following payload
+`PUT /templates/{templateName}/tags/{tagName}` with the following payload
 
-* `name` - Name of the template (ex: `mytemplate`)
-* `version` - Version of the template (ex: `1.1.0`)
-* `tag` - Name of the tag (ex: `stable`)
+* `version` - Exact version of the template (ex: `1.1.0`)
 
-#### Delete a template tag
+##### Delete a tag
 
 Delete the template tag. This does not delete the template itself. 
 
 *Note: This endpoint is only accessible in `build` scope and the permission is tied to the pipeline that creates the template.*
 
-`DELETE /templates/tags` with the following payload
-
-* `name` - Name of the template (ex: `mytemplate`)
-* `tag` - Name of the tag (ex: `stable`)
+`DELETE /templates/{templateName}/tags/{tagName}`
--- a/plugins/templates/README.md
+++ b/plugins/templates/README.md
@@ -32,7 +32,7 @@ server.register({
 
 `GET /templates?page={pageNumber}&count={countNumber}`
 
-#### Get single template
+#### Get a single template
 
 You can get a single template by providing the template name and the specific version or the tag.
 
@@ -47,13 +47,13 @@ You can get a single template by providing the template name and the specific ve
 * `version` - Version of the template
 
 #### Create a template
-Create a template will store the template data (`config`, `name`, `version`, `description`, `maintainer`, `labels`) into the datastore.
+Creating a template will store the template data (`config`, `name`, `version`, `description`, `maintainer`, `labels`) into the datastore.
 
 If the exact template and version already exist, the only thing that can be changed is `labels`.
 
 If the template already exists but not the version, the new version will be stored provided that the build has correct permissions.
 
-This endpoint is only accessible in `build` scope.
+*Note: This endpoint is only accessible in `build` scope and the permission is tied to the pipeline that first creates the template.*
 
 `POST /templates`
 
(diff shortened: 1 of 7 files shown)


plugins/templates/removeTag.js
```diff
@@ -0,0 +1,65 @@
...
+
+                return Promise.all([
+                    pipelineFactory.get(pipelineId),
+                    templateFactory.get({
+                        name: config.name,
+                        version: templateTag.version
+                    })
+                ])
+                .then(([pipeline, template]) => {
+                    // Check for permission
```
FenrirUnbound: There should be a check for the `template` existing

plugins/templates/createTag.js
```diff
@@ -0,0 +1,79 @@
+'use strict';
+
+const boom = require('boom');
+const schema = require('screwdriver-data-schema');
+const urlLib = require('url');
+
+/* Currently, only build scope is allowed to tag template due to security reasons.
```
tkyi: 👍 good comment

plugins/templates/createTag.js
```diff
@@ -0,0 +1,79 @@
...
+                // If template exists, then create the tag
+                return templateTagFactory.create(config).then((tag) => {
+                    const location = urlLib.format({
+                        host: request.headers.host,
+                        port: request.headers.port,
+                        protocol: request.server.info.protocol,
+                        pathname: `${request.path}/${tag.id}`
+                    });
+
+                    return reply(tag.toJson()).header('Location', location).code(201);
```
tkyi: Hm should this return the `template` or the `templateTag` or some combination of the two?

plugins/templates/removeTag.js
d2lam: I don't think it's necessary because of our logic. `templatetag` can't be created if the `template` itself doesn't exist. There is already [a check if `templateTag` exists]( https://github.com/screwdriver-cd/screwdriver/blob/02434e10cb28eb81214a33821b937d7a70f33c22/plugins/templates/removeTag.js#L37).  If `templateTag` exists, then `template` should exist.

plugins/templates/createTag.js
d2lam: I think it makes more sense to return just the tag. If it returns the `template` itself, it might be confusing to the users: `"Did I just create this template?"`

tkyi: Hm that makes sense.

plugins/templates/removeTag.js
FenrirUnbound: 1. Create template
2. Create tag
3. Delete template
4. Lookup tag
5. 500 error

FenrirUnbound: Deleting a template does not clean up the tags. And since you can operate on tags independently from templates, that delete operation isn't there yet.

d2lam: Deleting a template should delete the tags that are associated with it. It doesn't make sense to keep a tag that references something that doesn't exist. Just like how deleting a pipelines should delete all its jobs & builds.
Also, we don't have delete template at the moment.

plugins/templates/createTag.js
d2lam: Generally I think it should return what GOT CREATED. That's why I don't want to include the template information since the template is not created, the tag is created. For a `GET` it makes sense to return the template. 
What are others' opinions on this?

plugins/templates/removeTag.js
d2lam: I can add it for extra safety, but it will never reach the code with our current design.

plugins/templates/createTag.js
minz1027: I think template name, version and tag together should be good enough.

tkyi: No preference here

plugins/templates/removeTag.js
```diff
@@ -0,0 +1,65 @@
...
+            'hapi-swagger': {
+                security: [{ token: [] }]
+            }
+        },
+        handler: (request, reply) => {
+            const pipelineFactory = request.server.app.pipelineFactory;
+            const templateFactory = request.server.app.templateFactory;
+            const templateTagFactory = request.server.app.templateTagFactory;
+            const pipelineId = request.auth.credentials.pipelineId;
+            const config = request.payload;
```
tkyi: Is it weird to pass in a payload for a `DELETE`?

plugins/templates/createTag.js
```diff
@@ -0,0 +1,79 @@
...
+            const templateFactory = request.server.app.templateFactory;
+            const templateTagFactory = request.server.app.templateTagFactory;
+            const pipelineId = request.auth.credentials.pipelineId;
+            const config = request.payload;
+
+            return Promise.all([
+                pipelineFactory.get(pipelineId),
+                templateFactory.get({
+                    name: config.name,
+                    version: config.version
```
minz1027: There can be problem with the template get. It's not exact match. If user put `1.2` in their config, it can just get the first one with prefix `1.2`. Is it what we want? If that's what we want, we should document that.
 https://github.com/screwdriver-cd/models/blob/master/lib/templateFactory.js#L91

plugins/templates/removeTag.js
d2lam: I think it is unusual, but not forbidden according to [RFC 7231]( https://tools.ietf.org/html/rfc7231#section-4.3.5). 
A couple posts related to this:

- https://stackoverflow.com/questions/299628/is-an-entity-body-allowed-for-an-http-delete-request
- https://github.com/zalando-stups/lizzy/issues/151


I would say it's weird and we should avoid it if we can. But I think in this case, it makes sense to send a payload with it. An alternative would be delete `templates/tags/{name}/{tag}`?  I'm not sure if that's weird either. What do you think?

plugins/templates/README.md
```diff
@@ -73,3 +73,26 @@ Example payload:
   }
 }
 ```
+
+#### Create a tag for a template version
```
minz1027: How about update / retagging?

plugins/templates/removeTag.js
tkyi: I think to it would be safer for us to use the `/templates/tags/{name}/{tag}` endpoint instead of getting the data from the payload. Just in case other browsers drop the payload. What do you think?

tkyi:
Also do we do any checks to make sure the build that deletes the tag is the one that created the tag? Or can anyone create/delete a tag right now?

plugins/templates/removeTag.js
petey: The RESTful endpoint would be `DELETE /templates/{name}/tags/{tagName}`. You should be able to fetch the pipeline data from the template data and the JWT when comparing for authorization [build scope].

Alternatively, this could be `DELETE /templates/{name}/{tagName or version}` if it was less restful, but this adds a layer of confusion.

plugins/templates/createTag.js
FenrirUnbound: Don't mind any way you end up choosing, as long as the UI can consume it easily. If you want my opinion, I would have expected the resource that was created to be returned.

plugins/templates/removeTag.js
d2lam: Let's go with `DELETE /templates/{name}/tags{tagName}`

tkyi: @petey Should the `GET` endpoint also be `GET /templates/{name}/tags/{tagName}` then?

d2lam:
@tkyi  @petey  Updated with the new endpoints in this commit: https://github.com/screwdriver-cd/screwdriver/pull/647/commits/7d342943b10b58312a411902f842bf0a51d2ae56

- `PUT templates/{templateName}/tags/{tagName}` with payload as `{version}`
- `DELETE templates/{templateName}/tags/{tagName}`

plugins/templates/README.md
```diff
@@ -27,12 +27,12 @@ server.register({
 
 ### Routes
 
-#### Get all templates
-`page` and `count` optional
+#### Template
+##### Get all templates
```
tkyi: Hm this heading style gets interesting because it looks smaller than the *Arguments*

plugins/templates/createTag.js
```diff
@@ -0,0 +1,84 @@
...
+            const templateFactory = request.server.app.templateFactory;
+            const templateTagFactory = request.server.app.templateTagFactory;
+            const pipelineId = request.auth.credentials.pipelineId;
+            const name = request.params.templateName;
+            const tag = request.params.tagName;
+            const version = request.payload.version;
+
+            return Promise.all([
+                pipelineFactory.get(pipelineId),
+                templateFactory.get({ name, version }),
```
tkyi: Should this be `getTemplate`?

d2lam: Should be `get`, since we want to get the exact version. And `getTemplate` will be refactored into string only later.

plugins/templates/README.md
```diff
@@ -27,37 +27,37 @@ server.register({
...
 `GET /templates/{name}/{tag}` or `GET /templates/{name}/{version}`
 
-**Arguments**
+###### Arguments
 
 'name', 'tag' or 'version'
 
 * `name` - Name of the template
 * `tag` - Tag of the template (e.g. `stable`, `latest`, etc)
 * `version` - Version of the template
```
tkyi: Maybe we could call out that we'll autobump versions for them when they create templates.

PR #651, opened 2017-07-20
feat(523): Add GET and LIST routes to collections

joelseq:
## Context

For collections, we need to be able to LIST all of a user's collections as well as be able to GET a particular collection from the id.

## Objective

This PR adds the GET and LIST routes for collections

## References

Issue: #523 
PR regarding schema discussion: [screwdriver-cd/data-schema#162](https://github.com/screwdriver-cd/data-schema/pull/162)

### DIFF
--- a/plugins/collections/README.md
+++ b/plugins/collections/README.md
@@ -27,6 +27,14 @@ server.register({
 
 ### Routes
 
+#### Get a list of collections belonging to the requesting user
+
+`GET /collections`
+
+#### Get a single collection
+
+`GET /collections/{id}`
+
 #### Create a collection
 
 `POST /collections`
--- a/plugins/collections/get.js
+++ b/plugins/collections/get.js
@@ -1,0 +1,62 @@
+'use strict';
+
+const boom = require('boom');
+const joi = require('joi');
+const schema = require('screwdriver-data-schema');
+const getSchema = schema.models.collection.get;
+const idSchema = joi.reach(schema.models.collection.base, 'id');
+
+module.exports = () => ({
+    method: 'GET',
+    path: '/collections/{id}',
+    config: {
+        description: 'Get a single collection',
+        notes: 'Returns a collection record',
+        tags: ['api', 'collections'],
+        handler: (request, reply) => {
+            const { collectionFactory, pipelineFactory } = request.server.app;
+
+            return collectionFactory.get(request.params.id)
+                .then((collection) => {
+                    if (!collection) {
+                        throw boom.notFound('Collection does not exist');
+                    }
+
+                    // Store promises from pipelineFactory fetch operations
+                    const collectionPipelines = [];
+
+                    collection.pipelineIds.forEach((id) => {
+                        collectionPipelines.push(pipelineFactory.get(id));
+                    });
+
+                    return Promise.all(collectionPipelines)
+                        .then((pipelines) => {
+                            const result = collection.toJson();
+
+                            // Iterate over all the fetched pipelines, skip if null
+                            // else add it to the result object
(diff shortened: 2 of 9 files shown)


plugins/collections/get.js
```diff
@@ -0,0 +1,62 @@
...
+const boom = require('boom');
+const joi = require('joi');
+const schema = require('screwdriver-data-schema');
+const getSchema = schema.models.collection.get;
+const idSchema = joi.reach(schema.models.collection.base, 'id');
+
+module.exports = () => ({
+    method: 'GET',
+    path: '/collections/{id}',
+    config: {
```
ian-fox: Swagger block?

plugins/collections/README.md
```diff
@@ -27,6 +27,10 @@ server.register({
 
 ### Routes
 
+#### Get a list of collections belonging to the requesting user
+
+`GET /collections`
```
ian-fox: Params?

test/plugins/collections.test.js
```diff
@@ -20,6 +23,31 @@ const getCollectionMock = (collection) => {
...
+    let result = null;
+
+    testPipelines.forEach((pipeline) => {
+        if (pipeline.id === id) {
+            result = hoek.clone(pipeline);
+
+            result.update = sinon.stub();
+            result.toJson = sinon.stub().returns(pipeline);
+            result.remove = sinon.stub();
+        }
```
ian-fox: could we return `Promise.resolve` here instead of setting the variable in the loop and then returning it afterwards?

joelseq: The reason I kept the `return` outside was to have 1 common `return` statement to account for both the null case as well as in the case that the pipeline is found. I'll see if there's a way I can better refactor this. Let me know if you know of a better way

plugins/collections/get.js
```diff
@@ -0,0 +1,62 @@
...
+                        collectionPipelines.push(pipelineFactory.get(id));
+                    });
+
+                    return Promise.all(collectionPipelines)
+                        .then((pipelines) => {
+                            const result = collection.toJson();
+
+                            // Iterate over all the fetched pipelines, skip if null
+                            // else add it to the result object
+                            result.pipelines = pipelines.reduce((acc, curr) => {
```
joelseq: It's short for accumulator but I'll probably write out the entire thing instead. Are there other variable names that you recommend that would be better to use here? I was just using the names in the [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce)

test/plugins/collections.test.js
d2lam: I think either way is ok.

plugins/collections/get.js
d2lam: Full name is fine.

plugins/collections/list.js
```diff
@@ -0,0 +1,48 @@
...
+                security: [{ token: [] }]
+            }
+        },
+        handler: (request, reply) => {
+            const { userFactory, collectionFactory } = request.server.app;
+            const { username } = request.auth.credentials;
+
+            return userFactory.get({ username })
+                .then((user) => {
+                    if (!user) {
```
d2lam: I wonder if this case is even possible.

joelseq: I'm not too sure but I was referencing how [LIST](https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/tokens/list.js#L30) was implemented for tokens and it had that check there so I wrote it as well. Not sure if it will directly error out if the user is not found.

PR #657, opened 2017-07-21
fix(616): refactor GET template, move logic to models [3]

d2lam:
## Context

Moving logic to `models`. Since the same logic needs to be used by `config-parser`. Related: https://github.com/screwdriver-cd/config-parser/pull/41

Change is blocked by the actual model change. 

## Objective
Since `models` **will** have logic to check version and tag, it makes sense to just use it directly, instead of handling the logic here. 

## References
https://github.com/screwdriver-cd/screwdriver/issues/616

### DIFF
--- a/plugins/templates/get.js
+++ b/plugins/templates/get.js
@@ -4,9 +4,9 @@ const boom = require('boom');
 const joi = require('joi');
 const schema = require('screwdriver-data-schema');
 const getSchema = schema.models.template.get;
-const baseSchema = schema.models.template.base;
-const versionRegex = schema.config.regex.VERSION;
-const versionSchema = joi.reach(baseSchema, 'version');
+const nameSchema = joi.reach(schema.models.template.base, 'name');
+const versionSchema = joi.reach(schema.models.template.base, 'version');
+const tagSchema = joi.reach(schema.models.templateTag.base, 'tag');
 
 module.exports = () => ({
     method: 'GET',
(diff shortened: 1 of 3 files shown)


test/plugins/templates.test.js
```diff
@@ -153,7 +153,7 @@ describe('template plugin test', () => {
         beforeEach(() => {
             options = {
                 method: 'GET',
-                url: '/templates/screwdriver%2Fbuild/1.7.3'
```
tkyi: If a user wants to input their template name `chef/publish` in the url, do they need to use `%2F`?

plugins/templates/get.js
```diff
@@ -16,59 +16,28 @@ module.exports = () => ({
...
-                templateFactory.getTemplate({
-                    name: request.params.name,
-                    version
-                }).then((template) => {
+            return templateFactory.getTemplate(`${name}@${versionOrTag}`)
+                .then((template) => {
                     if (!template) {
-                        throw boom.notFound(`Template ${request.params.name} ` +
-                            `does not exist with version ${version}`);
+                        throw boom.notFound(`Template ${name}@${versionOrTag} does not exist`);
```
tkyi: Much cleaner 👍 🎉

test/plugins/templates.test.js
d2lam: Yes. But I want to leave it out for now since we haven't really addressed namespace yet.

d2lam: Yes. But I want to leave it out for now since we haven't really addressed namespace yet.

tkyi: Sounds reasonable. Is that tracked in an issue somewhere?

d2lam: It should be part of the templates issue.

tkyi: Added a checkbox to the templates issue

PR #659, opened 2017-07-24
design: New Auth Service Proposal

stjohnjohnson:
## Context

This is a rewrite of how we do authentication and authorization in the Screwdriver world.  The purpose of this is to harden our security as well as enable us to support private repositories.  Much of what is being proposed here is based on the [JWT Spec](https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-32) and Docker Hub's [implementation](https://docs.docker.com/registry/spec/auth/jwt/).

## Objective

Get feedback on new proposal for Auth service

## References


design/auth.md
```diff
@@ -0,0 +1,270 @@
...
+
+When accessing the Screwdriver API, API should validate that the user has sent a Bearer token valid for that API resource.  If not, it should reject with `401 Unauthorized` with a RFC7235 valid `WWW-Authenticate` header indicating how to get the appropriate Bearer token.
+
+### Search
+
+Pipeline List API will _only_ return "publicly accessible" pipelines.  That means "private" repositories will not show up in search.  Additionally, if Screwdriver is in "Required Login" mode, then that API will still return "publicly accessible" repositories, but require a valid user Bearer token.
+
+### Resources
+
+The resources (or scope) that we care about are `pipelines`, their `jobs`, and their `builds`.  There are two types of permission, `read` and `write`.
```
petey: Missing here is a discussion on template ownership and permissions. Forward thinking, we may also want to do something with shared steps as we do with templates.

design/auth.md
```diff
@@ -0,0 +1,270 @@
...
+    "project:20": "read",
+    "job:100": "write",
+    "job:101": "write",
+    "job:102": "write",
+    "job:103": "write"
+  }
+}
+```
+
+### Build Refresh Token
```
petey: We may want to talk about when refreshing a token is allowed. For example, if a build has a status that indicates it has been completed, any refresh request for that build should be denied.

tkyi:
Nitpicky comment but shouldn't everything be `pipeline`, not `project`?

design/auth.md
```diff
@@ -0,0 +1,270 @@
...
+
+### Login Types
+
+Tokens will be given out for logged-in users, anonymous users, and builds.  Subject will be based on the type:
+
+ - Logged-in User - `user:<YOUR-NAME-HERE>`
+ - Anonymous User - `anon:<RANDOM-ID-HERE>`
+ - Build - `build:<BUILD-ID-HERE>`
+
+## Flow Changes
```
r3rastogi: image
This is rendering as so, currently

r3rastogi: And clicking on any of the URLs generated results in a 404

design/auth.md
```diff
@@ -0,0 +1,270 @@
...
+
+##### 5 - No Access
+
+If `sue` requests access to the `21` pipeline, the `21` project is private, `sue` is not a collaborator:
+
+```json
+404 Resource Not Found
+```
+
+_Note, the reason we return 404 is to mask private repositories._
```
r3rastogi: Should add in a note that this is how Github (and maybe others) operate? Might be an unnecessary detail, but I thought it could be good to show that this is standard industry practice

r3rastogi: Also, "Note: The reason ..."

design/auth.md
```diff
@@ -0,0 +1,270 @@
...
+```json
+{
+  ....
+  "scope": [
+    "build:3001": "write"
+  ]
+}
+```
+
+_Note, in this example `projects` and `jobs` are not listed because it's assumed they have read permission based on the `build` listed._
```
r3rastogi: Again, "Note: In this example ..."

design/auth.md
```diff
@@ -0,0 +1,270 @@
...
+
+## Signing Changes
+
+Signing keys should be rotated periodically.  For us to do that, we should have an identity for each public/private key as well as an expiration.  That gives the clients the information to expire keys appropriately and be able to pull new keys when they are made available.
+
+```json
+{
+  "WbXr5q1JbSx8JDzTBm6u": {
```
r3rastogi: Should this have a meaningful name?

design/auth.md
stjohnjohnson: Wouldn't that just be tied to the Screwdriver pipeline?

design/auth.md
stjohnjohnson: fixed

PR #660, opened 2017-07-24
feat(523): Add UPDATE route for collections

joelseq:
## Objective

This PR adds an UPDATE route for collections so that you can modify the name, description, or pipelineIds for the pipelines in the collection

## References

Issue: #523

### DIFF
--- a/plugins/collections/README.md
+++ b/plugins/collections/README.md
@@ -53,3 +53,23 @@ Example payload:
     "pipelineIds": [12, 35, 47, 89]
 }
 ```
+
+#### Update a collection
+You can update the name, description, or pipelineIds of a collection.
+
+`PUT /collections/{id}`
+
+**Arguments**
+
+* `name` - An optional new name for the collection. Names must be unique.
+* `description` - An optional new description for the collection.
+* `pipelineIds` - An optional new array of ids of pipelines for the collection.
+
+Example payload:
+```json
+{
+    "name": "foo",
+    "description": "bar",
+    "pipelineIds": [1, 2, 5]
+}
+```
--- a/plugins/collections/get.js
+++ b/plugins/collections/get.js
@@ -31,7 +31,7 @@ module.exports = () => ({
 
                     return Promise.all(collectionPipelines)
                         .then((pipelines) => {
-                            const result = collection.toJson();
+                            const result = Object.assign({}, collection.toJson());
 
                             // Iterate over all the fetched pipelines, skip if null
                             // else add it to the result object
--- a/plugins/collections/update.js
+++ b/plugins/collections/update.js
@@ -1,0 +1,58 @@
+'use strict';
+
+const boom = require('boom');
+const joi = require('joi');
+const schema = require('screwdriver-data-schema');
+const idSchema = joi.reach(schema.models.collection.base, 'id');
+
+module.exports = () => ({
+    method: 'PUT',
+    path: '/collections/{id}',
+    config: {
+        description: 'Update a collection',
+        notes: 'Update a specific collection',
+        tags: ['api', 'collection'],
+        auth: {
+            strategies: ['token', 'session'],
+            scope: ['user']
+        },
+        plugins: {
+            'hapi-swagger': {
+                security: [{ token: [] }]
+            }
+        },
+        handler: (request, reply) => {
+            const { id } = request.params;
+            const { collectionFactory, userFactory } = request.server.app;
+            const { username } = request.auth.credentials;
+
+            // get the collection and user
+            return Promise.all([
+                collectionFactory.get({ id }),
+                userFactory.get({ username })
+            ])
+            .then(([oldCollection, user]) => {
+                if (!oldCollection) {
+                    throw boom.notFound(`Collection ${id} does not exist`);
+                }
+
+                // Check if user owns collection
+                if (oldCollection.userId !== user.id) {
+                    throw boom.unauthorized(`User ${username} does not own this collection`);
+                }
+
+                Object.assign(oldCollection, request.payload);
+
+                return oldCollection.update()
+                .then(updatedCollection => reply(updatedCollection.toJson()).code(200));
+            })
+            .catch(err => reply(boom.wrap(err)));
+        },
+        validate: {
+            params: {
+                id: idSchema
+            },
+            payload: schema.models.collection.update
+        }
+    }
+});
--- a/test/plugins/collections.test.js
+++ b/test/plugins/collections.test.js
@@ -10,6 +10,7 @@ const testCollection = require('./data/collection.json');
 const testCollectionResponse = require('./data/collection.response.json');
 const testCollections = require('./data/collections.json');
 const testPipelines = require('./data/pipelines.json');
+const updatedCollection = require('./data/updatedCollection.json');
 
 sinon.assert.expose(assert, { prefix: '' });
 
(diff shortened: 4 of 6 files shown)


plugins/collections/README.md
```diff
@@ -53,3 +53,23 @@ Example payload:
...
 ```
+
+#### Update a collection
+You can update the name, description, or pipelineIds of a collection.
+
+`PUT /collections/{id}`
+
+**Arguments**
+
+* `name` - Name of the collection. Names must be unique.
```
ian-fox: Is this one optional too?

plugins/collections/update.js
```diff
@@ -0,0 +1,70 @@
...
+
+                const oldCollectionUserId = oldCollection.toJson().userId;
+
+                // Check if user owns collection
+                if (oldCollectionUserId !== user.id) {
+                    throw boom.unauthorized(`User ${username} does not own this collection`);
+                }
+
+                // Check for fields to update
+                if (name) {
```
ian-fox: These three could probably be simplified to `Object.assign(oldCollection, request.payload);`

plugins/collections/update.js
```diff
@@ -0,0 +1,60 @@
...
+            return Promise.all([
+                collectionFactory.get({ id }),
+                userFactory.get({ username })
+            ])
+            .then(([oldCollection, user]) => {
+                if (!oldCollection) {
+                    throw boom.notFound(`Collection ${id} does not exist`);
+                }
+
+                const oldCollectionUserId = oldCollection.toJson().userId;
```
minz1027: why not `oldCollection.userId`?

test/plugins/collections.test.js
```diff
@@ -289,4 +290,98 @@ describe('collection plugin test', () => {
...
+            };
+
+            updatedCollectionMock = hoek.clone(collectionMock);
+            updatedCollectionMock.name = 'updated name';
+            updatedCollectionMock.description = 'updated description';
+            updatedCollectionMock.pipelineIds = [123, 124];
+
+            collectionFactoryMock.get.withArgs({ id }).resolves(collectionMock);
+            collectionMock.update.resolves(updatedCollectionMock);
+            collectionMock.toJson.returns(updatedCollection);
```
minz1027: `updatedCollection` is from a file under `data` and it's only used at line 326. Can we get rid of the file and do it this way instead? https://github.com/screwdriver-cd/screwdriver/blob/master/test/plugins/tokens.test.js#L286

test/plugins/collections.test.js
```diff
@@ -289,4 +290,98 @@ describe('collection plugin test', () => {
...
+                    description: 'updated description',
+                    pipelineIds: [123, 124]
+                },
+                credentials: {
+                    username,
+                    scope: ['user']
+                }
+            };
+
+            updatedCollectionMock = hoek.clone(collectionMock);
```
minz1027: The value inside the updatedCollectionMock does not matter actually. Only what returns from toJson matters

plugins/collections/update.js
joelseq: I tried doing that but I kept getting `undefined` when I tried to directly access it that way.

plugins/auth/login.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-/* eslint no-param-reassign: ["error", { "props": false }]*/
+/* eslint no-param-reassign: ["error", { "props": false }] */
```
FenrirUnbound: Is it really being re-assigned in this file?

plugins/collections/get.js
```diff
@@ -31,7 +32,7 @@ module.exports = () => ({
 
                     return Promise.all(collectionPipelines)
                         .then((pipelines) => {
-                            const result = collection.toJson();
+                            const result = hoek.clone(collection.toJson());
```
FenrirUnbound: Non-blocking comment.

Could also avoid using `hoek` by using `Object.assign()`

plugins/collections/update.js
minz1027: weird.. It works for me locally, let's try it out on your box to see what's the problem

plugins/collections/README.md
```diff
@@ -53,3 +53,23 @@ Example payload:
...
+#### Update a collection
+You can update the name, description, or pipelineIds of a collection.
+
+`PUT /collections/{id}`
+
+**Arguments**
+
+* `name` - An optional new name for the collection. Names must be unique.
+* `description` - An optional new description for the collection.
+* `pipelineIds` - An optional new array of ids of pipelines to be added to the collection.
```
FenrirUnbound: If all of these values are optional, does it do nothing if I `PUT` with an empty payload?

plugins/collections/update.js
FenrirUnbound: > I tried doing that but I kept getting `undefined` when I tried to directly access it that way.

Are you saying that you're experiencing this with your unit tests? That might mean that your mocks are configured incorrectly.

plugins/collections/README.md
```diff
@@ -53,3 +53,23 @@ Example payload:
...
 ```
+
+#### Update a collection
+You can update the name, description, or pipelineIds of a collection.
+
+`PUT /collections/{id}`
+
+**Arguments**
+
+* `name` - An optional new name for the collection. Names must be unique.
```
FenrirUnbound: > Names must be unique.

What's enforcing that?

test/plugins/collections.test.js
```diff
@@ -289,4 +290,98 @@ describe('collection plugin test', () => {
...
+        );
+
+        it('returns 200 and correct collection data when no fields modified', () => {
+            options.payload = {};
+            collectionMock.update.resolves(collectionMock);
+            collectionMock.toJson.returns(testCollection);
+
+            return server.inject(options).then((reply) => {
+                assert.deepEqual(reply.result, testCollection);
+                assert.calledOnce(collectionMock.update);
```
FenrirUnbound: You're just checking if `collectionMock.update` was called. Is there no check on whether the model's fields were not modified?

plugins/auth/login.js
joelseq: I was getting a linter error when I tried running the tests and the error was that there was no space in that comment

plugins/collections/README.md
joelseq: It's enforced in the the [data-schema](https://github.com/screwdriver-cd/data-schema/blob/master/models/collection.js#L105)

plugins/collections/README.md
joelseq: Yes it is supposed to do nothing

plugins/auth/login.js
FenrirUnbound: Do you get a linting error if you remove this line?

plugins/collections/get.js
```diff
@@ -31,7 +31,7 @@ module.exports = () => ({
 
                     return Promise.all(collectionPipelines)
                         .then((pipelines) => {
-                            const result = collection.toJson();
+                            const result = Object.assign({}, collection.toJson());
```
FenrirUnbound: Why do we care to clone the `collection` data?

plugins/collections/update.js
joelseq: @FenrirUnbound never mind figured it out. It was because my GET route was previously mutating the testCollection model object in memory before but it works now after I deep cloned it in the GET route.

plugins/collections/get.js
joelseq: So I found out that I was having a bug in my tests because if I don't clone the `collection` data, then the `collection` model object gets mutated in memory. Cloning it ensures that there are no side effects from this function.

plugins/auth/login.js
joelseq: @FenrirUnbound just tried it after removing that comment and nope I don't get a linting error. I'll make the change

plugins/collections/README.md
tkyi: Does this operation add pipelineIds to the collection or replace them?

joelseq: @tkyi the way that I have it right now, it can do both. If the user is adding a pipeline to the collection, then in the payload the id of the pipeline to add will also be included in the pipelineIds array

ian-fox: So if I make a request to `PUT /collections/1` with `{ pipelineIds: 3 }`, and the collection at `1` has pipelines `[1, 2]` what happens?

joelseq: @ian-fox assuming you meant `{ pipelineIds: [3] }`, then collection 1 will have pipelines `[3]` and the other 2 get removed. Would it be better to defer adding / removing pipelines to a separate route and just not allow pipelineIds to be modified in the UPDATE route? My thought process was that since mostly only the UI is going to be interfacing with the API, then if a user wants to add a pipeline with id `3` to a collection and it already has pipelines `[1, 2]`, then the UI would make a request to `PUT /collections/1` with `{ pipelineIds: [1, 2, 3] }`.

ian-fox: I think it's fine to have that in this route, that's how `PUT` is supposed to be used. My concern is more with the comments/docs about `pipelines to be added to the collection`, that makes it sound like the ones in the request will be added rather than replacing the current pipelines entirely.

joelseq: @ian-fox that's a good point! I'll change the description soon

r3rastogi: So is there/will there be any API route that just adds pipelines to a collection? Or would the workflow have to be for someone to check which pipelines are already in the collection and then send them along with the new ones they want to add to this route?

joelseq: @r3rastogi so right now I haven't really planned to add a route specifically for just adding or removing pipelines but I could add that in the future if there is a need for it

jithin1987: We could expose a POST endpoint later on to do additive updates to pipelineIds. When we have a user interaction where user is adding single/multiple pipelines to an existing collection, then an additive POST end point makes sense rather than use this PUT endpoint to send all data each time.

PR #669, opened 2017-07-27
refactor(collections): Add checks for CREATE and UPDATE routes

joelseq:
## Context

Currently, the CREATE and UPDATE routes still pass in pipelineIds in the payload even though the ids might be invalid.

## Objective

This PR adds in checks to the CREATE and UDPATE routes for collections to ensure that only valid ids of pipelines will be passed in to the config to create / update the collection.

## References

Issue: #523

### DIFF
--- a/plugins/collections/create.js
+++ b/plugins/collections/create.js
@@ -30,12 +30,23 @@ module.exports = () => ({
                         throw boom.notFound(`User ${username} does not exist`);
                     }
 
-                    return collectionFactory.create({
-                        name: request.payload.name,
-                        description: request.payload.description,
-                        userId: user.id,
-                        pipelineIds: request.payload.pipelineIds
-                    });
+                    const config = Object.assign({}, request.payload, { userId: user.id });
+
+                    // Check that the pipelines exist for the pipelineIds specified.
+                    if (request.payload.pipelineIds) {
+                        const { pipelineFactory } = request.server.app;
+
+                        return Promise.all(request.payload.pipelineIds.map(pipelineFactory.get))
+                        .then((pipelines) => {
+                            // If the pipeline exists, then add it to pipelineIds
+                            config.pipelineIds = pipelines.filter(pipeline =>
+                                pipeline).map(pipeline => pipeline.id);
+
+                            return collectionFactory.create(config);
+                        });
+                    }
+
+                    return collectionFactory.create(config);
                 })
                 .then((collection) => {
                     const location = urlLib.format({
--- a/test/plugins/collections.test.js
+++ b/test/plugins/collections.test.js
@@ -15,7 +15,7 @@ const updatedCollection = require('./data/updatedCollection.json');
 sinon.assert.expose(assert, { prefix: '' });
 
 const getCollectionMock = (collection) => {
-    const mock = hoek.clone(collection);
+    const mock = Object.assign({}, collection);
 
     mock.update = sinon.stub();
     mock.toJson = sinon.stub().returns(collection);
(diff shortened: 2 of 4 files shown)


plugins/collections/create.js
```diff
@@ -30,12 +30,26 @@ module.exports = () => ({
...
 
-                    return collectionFactory.create({
-                        name: request.payload.name,
-                        description: request.payload.description,
-                        userId: user.id,
-                        pipelineIds: request.payload.pipelineIds
-                    });
+                    const config = Object.assign({}, request.payload);
+
+                    config.userId = user.id;
```
FenrirUnbound: You could also keep this in the `Object.assign` operation above

plugins/collections/create.js
```diff
@@ -30,12 +30,26 @@ module.exports = () => ({
...
+                    // Check that the pipelines exist for the pipelineIds specified.
+                    if (request.payload.pipelineIds) {
+                        const { pipelineFactory } = request.server.app;
+
+                        return Promise.all(request.payload.pipelineIds.map(pipelineId =>
```
FenrirUnbound: Curious. Could this also be done with `request.payload.pipelineIds.map(pipelineFactory.get)`?

joelseq: Yep you're right, my bad!

test/plugins/collections.test.js
```diff
@@ -167,21 +165,58 @@ describe('collection plugin test', () => {
...
+                const expectedLocation = {
+                    host: reply.request.headers.host,
+                    port: reply.request.headers.port,
+                    protocol: reply.request.server.info.protocol,
+                    pathname: `${options.url}/${testCollection.id}`
                 };
+                const expected = Object.assign({}, testCollection);
 
+                delete expected.id;
+                delete expected.pipelineIds;
```
joelseq: It doesn't need it to be passed in to `create` I'm already handling it on the models end and I recently got a [PR](https://github.com/screwdriver-cd/models/pull/184) merged testing it out

test/plugins/collections.test.js
```diff
@@ -267,6 +302,22 @@ describe('collection plugin test', () => {
             });
         });
 
+        it('returns a collection only with pipelines that exist', () => {
```
FenrirUnbound: The `GET` operation filters as well? Does it save the changes to the collection?

test/plugins/collections.test.js
```diff
@@ -325,6 +376,15 @@ describe('collection plugin test', () => {
             })
         );
 
+        it('returns 200 when no pipelineIds in payload', () => {
+            delete options.payload.pipelineIds;
+
+            return server.inject(options).then((reply) => {
```
FenrirUnbound: Don't you also want to assert that the collection still has the pre-existing list of pipelineIDs?

test/plugins/collections.test.js
FenrirUnbound: It doesn't have to be passed in, but you're also not asserting that the payload being returned has it.

test/plugins/collections.test.js
joelseq: This is to handle the case where a pipeline used to exist but got deleted. It should not cause an issue while a `GET` operation is performed.

test/plugins/collections.test.js
joelseq: I tried doing it but then was pretty confused as to how I could test that functionality since there are no params passed in to update and instead the dirty keys determine what gets updated?

test/plugins/collections.test.js
FenrirUnbound: So if a collection references a pipeline that no longer exists, it will always filter it out on every call to this method. Which means that when a pipeline is removed, all collections that reference it will never update their set of IDs to remove that pipeline.

PR #649, opened 2017-07-19
feat(587): Skip builds when commit message includes [skip ci]

catto:
## Context

I'd like to implement feature reported at #587 

## Objective

Build will skip when the last commit message includes "[skip ci]" or "[ci skip]".
Changes needed for this feature has already been implemented at https://github.com/screwdriver-cd/scm-github/issues/64 and https://github.com/screwdriver-cd/scm-bitbucket/issues/40.
`/webhook` will reply 204 when the build is skipped.

## References

* Original issue: #587

PR #673, opened 2017-07-28
fix: Revert Skip builds when commit message includes [skip ci]

d2lam:
Reverts screwdriver-cd/screwdriver#649
It's breaking functional tests


catto:
:sob:

ian-fox:
Are we sure this was what was breaking the tests? They're still failing.

joelseq:
No this wasn't what was breaking the tests, it was caused by the cleanup not properly happening for the secrets functional test.

PR #683, opened 2017-08-03
fix: lint changes for eslint 4.x

petey:
Context
=======
ESLint 4.x included a rewrite of the `indent` rule that is much more strict. http://eslint.org/docs/user-guide/migrating-to-4.0.0#indent-rewrite

Objective:
----------
Fix all lint errors due to indent rule changes introduced by eslint 4.x

Misc:
------
`eslint --fix .` is nice.
Apologies ahead of time to Yahoo! Japan, as some of my previous indentation guidance is not consistent with this change.


plugins/builds/create.js
```diff
@@ -39,59 +39,59 @@ module.exports = () => ({
...
             // scmUri is buried in the pipeline, so we get that from the job
-            .then(([job, user]) => job.pipeline.then(pipeline =>
-                user.getPermissions(pipeline.scmUri)
-                // check if user has push access
-                .then((permissions) => {
-                    if (!permissions.push) {
-                        throw boom.unauthorized(`User ${username} `
+                .then(([job, user]) => job.pipeline.then(pipeline =>
+                    user.getPermissions(pipeline.scmUri)
+                    // check if user has push access
```
joelseq: should this be the same level of indentation as the `.then` below it?

plugins/builds/create.js
```diff
@@ -39,59 +39,59 @@ module.exports = () => ({
...
-                    token,
-                    scmUri: pipeline.scmUri,
-                    prNum: job.prNum
-                }))
-                .then((sha) => {
-                    let type = 'pipeline';
-                    let workflow = pipeline.workflow;
+                            }
+                        })
+                    // user has good permissions, sync and create a build
```
joelseq: should this comment be the same level of indentation as the `.then` below it?

plugins/builds/create.js
```diff
@@ -39,59 +39,59 @@ module.exports = () => ({
...
+                        .then((event) => {
+                            payload.eventId = event.id;
 
-                    return buildFactory.create(payload);
-                })
-            )).then((build) => {
+                            return buildFactory.create(payload);
+                        })
+                )).then((build) => {
                 // everything succeeded, inform the user
```
joelseq: same question regarding level of indentation for comment as ones above

ian-fox: This is code that has been removed.

plugins/builds/create.js
petey: Thanks, fixed.

PR #691, opened 2017-08-10
[523]fix: define feature scenarios for collections

FenrirUnbound:
## Context

The team started working on the [homepage favorites](https://github.com/screwdriver-cd/screwdriver/issues/523) prior to creating a cucumber-js feature file. This feature file helps define scenarios that a user is expected to interact with the Screwdriver API and assert that the results are sane.

## Objective

Create a feature file for collections, asserting that the user can use the API in order to create, list, update, and delete collections.

## References

* Issue #523 

EDIT1: fixed link

### DIFF
--- a/features/collections.feature
+++ b/features/collections.feature
@@ -1,0 +1,59 @@
+@collections
+Feature: User Collections
+
+    Users want to be able to view a set of pipelines that they most care about. These
+    pipelines can be grouped into collections that these users own and maintain.
+
+    Collections can be modified to include and exclude pipelines over their lifetimes.
+    These collections can also be shared between users, such as teammates sharing
+    collections they're responsible for.
+
+    Rules:
+        - Collections are unique by name and owner
+
+    Background:
+        Given an existing repository with these users and permissions:
+            | name          | permission  |
+            | calvin        | admin       |
+#            | miss wormwood | no access   |
+        And an existing pipeline with that repository
+
+    @ignore
+    Scenario: Create New Collection
+        And "calvin" is logged in
+        When they create a new collection "myCollection" with that pipeline
+        Then they can see that collection
+        And the collection contains that pipeline
+
+    @ignore
+    Scenario: Update Existing Collection
+        And "calvin" is logged in
+        When they create a new collection "myCollection"
+        Then they can see that collection
+        And the collection is empty
+        When they update the collection "myCollection" with that pipeline
+        Then they can see that collection
+        And the collection contains that pipeline
+
+    @ignore
+    Scenario: Listing A User's Collection
+        And "calvin" is logged in
+        And they have a collection "myCollection"
+        And they have a collection "anotherCollection"
+        When they fetch all their collections
+        Then they can see those collections
+
+    @ignore
+    Scenario: Deleting A Collection
+        And "calvin" is logged in
+        And they have a collection "myCollection"
+        When they delete that collection
+        Then that collection no longer exists
+        And that pipeline still exists
+
+    @ignore
+    Scenario: Collections Are Unique
+        And "calvin" is logged in
+        And they have a collection "myCollection"
+        When they create a new collection "myCollection"
+        Then they receive an error regarding unique collections


features/collections.feature
```diff
@@ -0,0 +1,51 @@
...
+
+    Users want to be able to view a set of pipelines that they most care about. These
+    pipelines can be grouped into collections that these users own and maintain.
+
+    Collections can be modified to include and exclude pipelines over their lifetimes.
+    These collections can also be shared between users, such as teammates sharing
+    collections they're responsible for.
+
+    Rules:
+        - Collections are unique to an owning user and its name
```
r3rastogi: Grammatically, the "its" here is referring to the user. Maybe reword to "Collections are unique by name and owner"

r3rastogi:
Why no "logged out" cases?

features/collections.feature
```diff
@@ -0,0 +1,51 @@
...
+        Then they can see that collection
+        And the collection contains that pipeline
+
+    @ignore
+    Scenario: Update Existing Collection
+        And "calvin" is logged in
+        When they create a new "myCollection"
+        Then they can see that collection
+        And the collection is empty
+        When they update the "myCollection" collection with the pipeline
```
r3rastogi: Give a specific pipeline number here

features/collections.feature
```diff
@@ -0,0 +1,51 @@
...
+    Scenario: Create New Collection
+        And "calvin" is logged in
+        When they create a new "myCollection" collection with that pipeline
+        Then they can see that collection
+        And the collection contains that pipeline
+
+    @ignore
+    Scenario: Update Existing Collection
+        And "calvin" is logged in
+        When they create a new "myCollection"
```
tkyi: can this either be "anotherCollection"
or just have it be assumed that they own the collection already
` And they have a collection "myCollection"`

features/collections.feature
```diff
@@ -0,0 +1,51 @@
...
+        Given an existing repository with these users and permissions:
+            | name          | permission  |
+            | calvin        | admin       |
+#            | miss wormwood | no access   |
+        And an existing pipeline with that repository
+
+    @ignore
+    Scenario: Create New Collection
+        And "calvin" is logged in
+        When they create a new "myCollection" collection with that pipeline
```
tkyi: a pipeline

tkyi: Oh okay I see that it refers to the background pipeline

features/collections.feature
FenrirUnbound: That's so we can re-use step definitions

tkyi: You can probably say "When they create a new collection __________"

features/collections.feature
tkyi: Can you make it consistent by putting "collection" before "nameOfCollection" here, like in lines 41 and 42.

features/collections.feature
tkyi: Still need to add the word "collection" here

features/collections.feature
```diff
@@ -0,0 +1,51 @@
...
+
+    @ignore
+    Scenario: Update Existing Collection
+        And "calvin" is logged in
+        When they create a new "myCollection"
+        Then they can see that collection
+        And the collection is empty
+        When they update the "myCollection" collection with the pipeline
+        Then they can see that collection
+        And the collection contains that pipeline
```
d2lam: Seems like just 1 pipeline in this case? We should also test for multiple pipelines case.

features/collections.feature
```diff
@@ -0,0 +1,51 @@
...
+        And they have a collection "anotherCollection"
+        When they fetch all their collections
+        Then they can see those collections
+
+    @ignore
+    Scenario: Deleting A Collection
+        And "calvin" is logged in
+        And they have a collection "myCollection"
+        When they delete that collection
+        Then that collection no longer exists
```
d2lam: We should add a line here to make sure the collection is deleted but the pipelines remain.

features/collections.feature
```diff
@@ -0,0 +1,51 @@
...
+
+    Users want to be able to view a set of pipelines that they most care about. These
+    pipelines can be grouped into collections that these users own and maintain.
+
+    Collections can be modified to include and exclude pipelines over their lifetimes.
+    These collections can also be shared between users, such as teammates sharing
+    collections they're responsible for.
+
+    Rules:
+        - Collections are unique by name and owner
```
d2lam: Should add a test for the same user creating a collection with the same name should give an error

features/collections.feature
FenrirUnbound: It's meant to cover the case where a user initializes a collection with a pipeline. Do you want it to assert that it can be initialized with more than one pipeline?

features/collections.feature
FenrirUnbound: > We should add a line here to make sure the collection is deleted but the pipelines remain.

Collections have nothing to do with maintaining the lifespan of pipelines. Why do we want to check that the pipelines are unaffected?

FenrirUnbound:
> Why no "logged out" or "no access" cases?

This is intended for certifying core functionality, so these cases will be addressed in a future update.

features/collections.feature
d2lam: @FenrirUnbound  Yes

features/collections.feature
d2lam: I think the expected behaviors of deleting collections is the pipelines are not deleted, so it's safer to assert that here. If you don't want it, it's ok.

FenrirUnbound: > I think the expected behaviors of deleting collections is the pipelines are not deleted, so it's safer to assert that here. 

Added the assertion 👍

PR #692, opened 2017-08-11
fix(365): Fix format for auth/contexts API

catto:
## Context

Multiple SCM feature API has been implemented to API(#365).
I'm trying to implement it to UI also, but I found that current format for `auth/contexts` API doesn't suit for dealing at clients.

## Objective

I'd like to change the format for `auth/contests` API so that we can implement an authorization switcher to UI easily.

before
```
[
  { "github:foo.com": "thisisdisplayname" },
  ...
]
```

after
```
[
  {
    context: "github:github.com",
    displayName: "thisisdisplayname"
  },
  ...
]
```

## References

- The issue for this feature: #365


r3rastogi:
This would be a breaking change, technically. Please bump the major version

catto:
@r3rastogi Technically, yes. But it isn't used by any component. Should I bump the major version even if it was merged yesterday?

tkyi:
@catto Actually we're still on major version 0 for the API. So I guess we should just bump minor.

catto:
Oh I see. I'm going to do that!

d2lam:
@r3rastogi  @catto  we don't bump major for this repo.

PR #702, opened 2017-08-17
fix: read scms config as json

minz1027:
We need scms config as json, so that when user wants to have multiple scms, they are not tied to the `github` one we currently have.

They can have multiple scms defined in SCM_SETTINGS.

I have updated beta and prod to have SCM_SETTINGS secret.

### DIFF
--- a/config/custom-environment-variables.yaml
+++ b/config/custom-environment-variables.yaml
@@ -106,25 +106,27 @@ executor:
                 database: QUEUE_REDIS_DATABASE
 
 scms:
-    github:
-        plugin: github
-        config:
-            # The client id used for OAuth with github. Look up GitHub OAuth for details
-            # https://developer.github.com/v3/oauth/
-            oauthClientId: SECRET_OAUTH_CLIENT_ID
-            # The client secret used for OAuth with github
-            oauthClientSecret: SECRET_OAUTH_CLIENT_SECRET
-            # You can also configure for use with GitHub enterprise
-            gheHost: SCM_GITHUB_GHE_HOST
-            # The username and email used for checkout with github
-            username: SCM_USERNAME
-            email: SCM_EMAIL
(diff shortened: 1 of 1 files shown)


config/custom-environment-variables.yaml
```diff
@@ -106,25 +106,27 @@ executor:
...
-            # The username and email used for checkout with github
-            username: SCM_USERNAME
-            email: SCM_EMAIL
-            # Secret to add to GitHub webhooks so that we can validate them
-            secret: WEBHOOK_GITHUB_SECRET
-            # Whether it supports private repo: boolean value.
-            # If true, it will ask for read and write access to public and private repos
-            # https://developer.github.com/v3/oauth/#scopes
-            privateRepo: SCM_PRIVATE_REPO_SUPPORT
+    __name: SCM_SETTINGS
```
ian-fox: Why the underscore prefix?

config/custom-environment-variables.yaml
```diff
@@ -106,25 +106,27 @@ executor:
...
+    __format: json
+    # github:
```
r3rastogi: Why is everything commented out?

ian-fox: We don't want to pass in the same config twice, so if it lives in the SCM settings now instead of as individual variables in the YAML. If the question is why it's commented out and not deleted though then I'm not sure :/

FenrirUnbound: It's an example of how to define these values if you need to override it via environment variable.

config/custom-environment-variables.yaml
FenrirUnbound: https://github.com/lorenwest/node-config/wiki/Environment-Variables#custom-environment-variables

FenrirUnbound:
For clarification, I didn't merge it because I wasn't sure that the other reviewers had a chance to look at the responses.

PR #708, opened 2017-08-17
fix(tests): move createTestUser into a function

ian-fox:
## Context

Cucumber was running `createTestUser` as part of the setup for functional tests.

## Objective
Move the script into a function so that cucumber doesn't run it.

### DIFF
--- a/features/scripts/create-test-user.js
+++ b/features/scripts/create-test-user.js
@@ -1,4 +1,5 @@
 #!/usr/bin/env node
+
 /**
  * Make a user in the database and generate an access token for it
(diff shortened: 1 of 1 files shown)


features/scripts/create-test-user.js
```diff
@@ -8,78 +9,87 @@
 'use strict';
 
 /* eslint-disable import/no-dynamic-require */
+/* eslint-disable global-require */
```
d2lam: Can put all these on the same line
```
/* eslint-disable import/no-dynamic-require, global-require, no-console */
```

PR #712, opened 2017-08-21
fix(tests): update create-test-user to work with multiple SCMs

ian-fox:
## Context

Cucumber-js was running `create-test-user.js` because it was in the `features` folder.

## Objective

Move `create-test-user` to be in a function and export it so that when cucumber traverses the `features` folder it doesn't get run.


package.json
```diff
@@ -8,7 +8,7 @@
     "test": "jenkins-mocha --recursive --timeout 3000",
     "start": "./bin/server",
     "functional": "cucumber-js --format=pretty --tags 'not @ignore'",
-    "create-test-user": "node ./features/scripts/create-test-user.js",
+    "create-test-user": "node -e 'require(\"./features/scripts/create-test-user.js\")()'",
```
minz1027: didn't know you can do this. 👍

PR #714, opened 2017-08-22
feat(523): Add last build health to pipelines in collection get request

joelseq:
## Context

For the user dashboard, the UI needs information about the health of the last build.

## Objective

This PR adds the last build object to the pipeline object in the response for the GET request for a collection.

## References

Issue: #523


package.json
```diff
@@ -86,14 +86,14 @@
...
     "screwdriver-config-parser": "^3.10.0",
-    "screwdriver-data-schema": "^17.0.0",
+    "screwdriver-data-schema": "^18.0.0",
     "screwdriver-datastore-sequelize": "^4.0.0",
     "screwdriver-executor-docker": "^2.2.2",
     "screwdriver-executor-k8s": "^10.3.3",
     "screwdriver-executor-k8s-vm": "^1.0.0",
     "screwdriver-executor-queue": "^1.1.0",
     "screwdriver-executor-router": "^1.0.0",
-    "screwdriver-models": "24.0.0",
```
ian-fox: If we're moving away from 24 we can unpin the version (add the caret back)

plugins/collections/get.js
```diff
@@ -6,6 +6,36 @@ const schema = require('screwdriver-data-schema');
 const getSchema = schema.models.collection.get;
 const idSchema = joi.reach(schema.models.collection.base, 'id');
 
+/**
+ * Helper function to get last build health of pipeline
+ *
+ * @param {Pipeline} pipeline - pipeline model to get last build health of
```
ian-fox: jsdoc doesn't match function

plugins/collections/get.js
```diff
@@ -6,6 +6,38 @@ const schema = require('screwdriver-data-schema');
 const getSchema = schema.models.collection.get;
 const idSchema = joi.reach(schema.models.collection.base, 'id');
 
+/**
+ * Helper function to get last builds of pipeline
+ *
+ * @param {Pipeline} pipeline     - pipeline model to get last builds of
```
tkyi: looks like this param is out of date

ian-fox:
Of note: this PR now pulls in the token changes as well since https://github.com/screwdriver-cd/screwdriver/pull/717 became blocked on this following the `data-schema` version bump.

plugins/collections/get.js
```diff
@@ -6,6 +6,42 @@ const schema = require('screwdriver-data-schema');
...
+            return eventFactory.get(pipeline.lastEventId)
+                .then((event) => {
+                    if (!event) {
+                        return result;
+                    }
+
+                    return event.getBuilds()
+                        .then((builds) => {
+                            if (builds.length) {
+                                result.lastBuilds = builds.map(b => b.toJson());
```
minz1027: If I remember correctly we are going to show the status of the last single build, right?
Then why not just have the logic here to return the health for that build instead of returning all the builds for that event. Are you going to handle that in UI?

joelseq: I was discussing it with @d2lam yesterday and we decided on showing all the builds for the last event rather than just a single build status to give more context. The reason I'm passing in the entire build objects here is to allow for sorting on most recent builds and for other stuff that might be needed in the future on the UI.

plugins/collections/get.js
```diff
@@ -6,6 +6,42 @@ const schema = require('screwdriver-data-schema');
 const getSchema = schema.models.collection.get;
 const idSchema = joi.reach(schema.models.collection.base, 'id');
 
+/**
+ * Helper function to get last builds of pipelines
+ *
+ * @param {Pipeline} pipelines     - pipeline models to get last builds for
```
tkyi: Shouldn't this be Array?

joelseq: @tkyi yes that is correct, I'll fix that right now!

plugins/collections/get.js
```diff
@@ -6,6 +6,42 @@ const schema = require('screwdriver-data-schema');
...
+            return eventFactory.get(pipeline.lastEventId)
+                .then((event) => {
+                    if (!event) {
+                        return result;
+                    }
+
+                    return event.getBuilds()
+                        .then((builds) => {
+                            if (builds.length) {
+                                result.lastBuilds = builds.map(b => b.toJson()).reverse();
```
joelseq: The order of the lastBuilds is being reversed here because `getBuilds()` returns the builds in the order of the most recent build being first. Reversing the order will allow us to map the builds to the workflow accurately on the UI  for displaying on the dashboard.

plugins/collections/get.js
```diff
@@ -6,6 +6,42 @@ const schema = require('screwdriver-data-schema');
...
+            const result = Object.assign({}, pipeline.toJson());
+
+            if (!pipeline.lastEventId) {
+                return result;
+            }
+
+            return eventFactory.get(pipeline.lastEventId)
+                .then((event) => {
+                    if (!event) {
+                        return result;
```
tkyi: Should this throw an error?

plugins/collections/get.js
```diff
@@ -6,6 +6,42 @@ const schema = require('screwdriver-data-schema');
...
+                    return event.getBuilds()
+                        .then((builds) => {
+                            if (builds.length) {
+                                result.lastBuilds = builds.map(b => b.toJson()).reverse();
+                            }
+
+                            return result;
+                        });
+                })
+                .catch(() => result);
```
tkyi: Should this throw an error?

joelseq: I'm not sure if it is very likely for it to fail but in the case that it does fail it could throw an error here. My thought process was that if an error is caused here by the `eventFactory` lookup, it should just return whatever data it has for the pipeline and not error out the whole thing and lead to the user receiving no data at all on the UI as this is the `GET` for the entire collection.

plugins/collections/get.js
joelseq: If this throws an error here it will be caught by the catch block below which currently returns `result`. I wrote my thought process behind returning the result and not throwing an error instead in a [reply](https://github.com/screwdriver-cd/screwdriver/pull/714#discussion_r134926676) to your comment

plugins/collections/get.js
tkyi: Maybe a warning at least?

ian-fox:
Found a slight problem with tokens, fixing now.

PR #717, opened 2017-08-23
feat(tokens): store tokens using pbkdf2

ian-fox:
## Context

After taking a security workshop, I learned that sha256 is not a secure way to store passwords. PBKDF2 is much better, as it takes longer (making it harder to crack).
## Objective

This PR pulls in the new `models` package to make use of the more secure password hashing algorithm for tokens.

I'm currently unable to get some of the functional tests surrounding starting a build to run locally, but all functional tests relating to tokens pass.

## References
https://github.com/screwdriver-cd/models/commit/ba8967ebe104c4c41859bfb09a4ab4eea8a7299d


ian-fox:
Pulling change into https://github.com/screwdriver-cd/screwdriver/pull/714 as that requires the new models (and thus the new argument to `TokenFactory`), and this PR is blocked on it.

PR #722, opened 2017-08-25
feat(tokens): use pbkdf2 to securely store tokens

ian-fox:
## Context

Tokens were previously hashed with SHA256, which is too fast to be resilient to brute force attacks.

`screwdriver-models@26.0.0` changes how tokens are stored to use PBKDF2, a more secure algorithm. It also takes in a server-side salt.

## Objective

Use PBKDF2 to hash tokens in a more secure manner.

## References

https://github.com/screwdriver-cd/models/pull/191


ian-fox:
@FenrirUnbound this will invalidate the tokens that currently exist but I can migrate ones that are hard to change.

PR #731, opened 2017-09-01
test(collections): Add functional tests for collections

joelseq:
## Objective

Add functional tests for collections

## References

Issue: #523

### DIFF
--- a/features/step_definitions/collection.js
+++ b/features/step_definitions/collection.js
@@ -1,0 +1,275 @@
+'use strict';
+
+const Assert = require('chai').assert;
+const request = require('../support/request');
+const { defineSupportCode } = require('cucumber');
+
+// Timeout of 15 seconds
+const TIMEOUT = 15 * 1000;
+
+/**
+ * Helper function to create a collection
+ * @method createCollection
+ * @param   {Object}  body    The body of the request
+ * @returns {Promise}
+ */
+function createCollection(body) {
+    return this.getJwt(this.apiToken)
+        .then((response) => {
+            this.jwt = response.body.token;
+
+            return request({
+                uri: `${this.instance}/${this.namespace}/collections`,
+                method: 'POST',
+                auth: {
+                    bearer: this.jwt
+                },
+                body,
+                json: true
+            });
+        });
+}
+
+/**
+ * Helper function to delete a collection
+ * @method deleteCollection
+ * @param   {Number} [id]  Id of the collection to delete
+ * @returns {Promise}
+ */
+function deleteCollection(id) {
+    if (!id) {
+        return Promise.resolve();
+    }
+
+    return request({
+        uri: `${this.instance}/${this.namespace}/collections/${id}`,
+        method: 'DELETE',
+        auth: {
+            bearer: this.jwt
+        },
+        json: true
+    }).then((response) => {
+        Assert.strictEqual(response.statusCode, 204);
+    });
+}
+
+defineSupportCode(({ Before, Given, Then, When, After }) => {
+    Before('@collections', function hook() {
+        this.repoOrg = this.testOrg;
+        this.repoName = 'functional-collections';
+        this.pipelineId = null;
+        this.firstCollectionId = null;
+        this.secondCollectionId = null;
+    });
+
+    When(/^they create a new collection "myCollection" with that pipeline$/,
+        { timeout: TIMEOUT }, function step() {
+            return this.ensurePipelineExists({ repoName: this.repoName })
+                .then(() => {
+                    const requestBody = {
+                        name: 'myCollection',
+                        pipelineIds: [this.pipelineId]
+                    };
+
+                    return createCollection.call(this, requestBody);
+                })
+                .then((response) => {
+                    Assert.strictEqual(response.statusCode, 201);
+                    this.firstCollectionId = response.body.id;
+                });
+        });
+
+    Then(/^they can see that collection$/, function step() {
+        return request({
+            uri: `${this.instance}/${this.namespace}/collections/${this.firstCollectionId}`,
+            method: 'GET',
+            auth: {
+                bearer: this.jwt
+            },
+            json: true
+        }).then((response) => {
+            Assert.strictEqual(response.statusCode, 200);
+            Assert.strictEqual(response.body.name, 'myCollection');
+        });
+    });
+
+    Then(/^the collection contains that pipeline$/, function step() {
+        const pipelineId = parseInt(this.pipelineId, 10);
+
+        return request({
+            uri: `${this.instance}/${this.namespace}/collections/${this.firstCollectionId}`,
+            method: 'GET',
+            auth: {
+                bearer: this.jwt
+            },
+            json: true
+        }).then((response) => {
+            Assert.strictEqual(response.statusCode, 200);
+            Assert.deepEqual(response.body.pipelineIds, [pipelineId]);
+        });
+    });
+
+    When(/^they create a new collection "myCollection"$/, function step() {
+        return createCollection.call(this, { name: 'myCollection' })
+            .then((response) => {
+                Assert.strictEqual(response.statusCode, 201);
+                this.firstCollectionId = response.body.id;
+            });
+    });
+
+    Then(/^the collection is empty$/, function step() {
+        return request({
+            uri: `${this.instance}/${this.namespace}/collections/${this.firstCollectionId}`,
+            method: 'GET',
+            auth: {
+                bearer: this.jwt
+            },
+            json: true
+        }).then((response) => {
+            Assert.strictEqual(response.statusCode, 200);
+            Assert.deepEqual(response.body.pipelineIds, []);
+        });
+    });
+
+    When(/^they update the collection "myCollection" with that pipeline$/,
+        { timeout: TIMEOUT }, function step() {
+            return this.ensurePipelineExists({ repoName: this.repoName })
+                .then(() => {
+                    const pipelineId = parseInt(this.pipelineId, 10);
+
+                    request({
+                        uri: `${this.instance}/${this.namespace}/collections/` +
+                            `${this.firstCollectionId}`,
+                        method: 'PUT',
+                        auth: {
+                            bearer: this.jwt
+                        },
+                        body: {
+                            pipelineIds: [pipelineId]
+                        },
+                        json: true
+                    }).then((response) => {
+                        Assert.strictEqual(response.statusCode, 200);
+                    });
+                });
+        });
+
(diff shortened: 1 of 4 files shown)


features/step_definitions/collection.js
```diff
@@ -0,0 +1,267 @@
...
+                Assert.strictEqual(response.statusCode, 400);
+            });
+    });
+
+    After('@collections', function hook() {
+        // Delete the collections created in the functional tests if they exist
+        return Promise.all([
+            deleteCollection.call(this, this.firstCollectionId),
+            deleteCollection.call(this, this.secondCollectionId)
+        ]);
```
joelseq: The cleanup needs to be done in the `After` because for each scenario a new world gets created so the `firstCollectionId` and `secondCollectionId` variables don't exist in the `Before` for the following scenario.

features/step_definitions/collection.js
```diff
@@ -0,0 +1,267 @@
...
+    When(/^they create a new collection "myCollection" with that pipeline$/, function step() {
+        return this.ensurePipelineExists({ repoName: this.repoName })
+            .then(() => {
+                const requestBody = {
+                    name: 'myCollection',
+                    pipelineIds: [this.pipelineId]
+                };
+
+                return createCollection.call(this, requestBody)
+                    .then((response) => {
```
FenrirUnbound: There's no reason to keep this chained in this block. We can move it out like:

```
return this.ensurePipelineExists(...)
    .then(() => {
        const requestBody = {...};
        return createCollection.call(...);
    })
    .then((response) => {
        ...
    })
```

features/step_definitions/collection.js
```diff
@@ -0,0 +1,267 @@
...
+    });
+
+    When(/^they create a new collection "myCollection"$/, function step() {
+        return createCollection.call(this, { name: 'myCollection' })
+            .then((response) => {
+                Assert.oneOf(response.statusCode, [409, 201]);
+
+                if (response.statusCode === 201) {
+                    this.firstCollectionId = response.body.id;
+                } else {
```
FenrirUnbound: This step should be asserting that "myCollection" is brand new and never been created before. Why is there an `else` case?

features/step_definitions/collection.js
```diff
@@ -0,0 +1,267 @@
...
+        return this.ensurePipelineExists({ repoName: this.repoName })
+            .then(() => {
+                const requestBody = {
+                    name: 'myCollection',
+                    pipelineIds: [this.pipelineId]
+                };
+
+                return createCollection.call(this, requestBody)
+                    .then((response) => {
+                        Assert.oneOf(response.statusCode, [409, 201]);
```
FenrirUnbound: This is supposed to assert that the collection "myCollection" is newly created. Why is `409` an acceptable response code?

features/step_definitions/collection.js
```diff
@@ -0,0 +1,267 @@
...
+        return createCollection.call(this, { name: 'myCollection' })
+            .then((response) => {
+                Assert.oneOf(response.statusCode, [409, 201]);
+
+                if (response.statusCode === 201) {
+                    this.firstCollectionId = response.body.id;
+                } else {
+                    const str = response.body.message;
+
+                    this.firstCollectionId = str.split(': ')[1];
```
FenrirUnbound: Can be re-written as:

```
[, this.firstCollectionId] = str.split(': );
```

features/step_definitions/collection.js
```diff
@@ -0,0 +1,267 @@
...
+        return createCollection.call(this, { name: 'anotherCollection' })
+            .then((response) => {
+                Assert.oneOf(response.statusCode, [409, 201]);
+
+                if (response.statusCode === 201) {
+                    this.firstCollectionId = response.body.id;
+                } else {
+                    const str = response.body.message;
+
+                    this.secondCollectionId = str.split(': ')[1];
```
FenrirUnbound: Can be re-written as:

```
[, this.secondCollectionId] = str.split(': );
```

features/step_definitions/collection.js
```diff
@@ -0,0 +1,267 @@
...
+                bearer: this.jwt
+            },
+            json: true
+        }).then((response) => {
+            Assert.strictEqual(response.statusCode, 200);
+        });
+    });
+
+    Then(/^they can see those collections$/, function step() {
+        return request({
```
FenrirUnbound: This should be receiving data from the `/^they fetch all their collections$/` step and not making a separate call.

features/step_definitions/collection.js
```diff
@@ -0,0 +1,267 @@
...
+    When(/^they delete that collection$/, function step() {
+        return request({
+            uri: `${this.instance}/${this.namespace}/collections/${this.firstCollectionId}`,
+            method: 'DELETE',
+            auth: {
+                bearer: this.jwt
+            },
+            json: true
+        }).then((response) => {
+            Assert.oneOf(response.statusCode, [404, 204]);
```
FenrirUnbound: The collection should actually exist. `404` should not be an acceptable status code.

features/step_definitions/collection.js
```diff
@@ -0,0 +1,267 @@
...
+                bearer: this.jwt
+            },
+            json: true
+        }).then((response) => {
+            Assert.strictEqual(response.statusCode, 404);
+        });
+    });
+
+    Then(/^that pipeline still exists$/, function step() {
+        return this.ensurePipelineExists({ repoName: this.repoName });
```
FenrirUnbound: This should not use the `this.ensurePipelineExists`, which creates a pipeline if missing. This should be making a call for the specific pipeline and not receive a `404`.

features/step_definitions/collection.js
FenrirUnbound: Curious: what happens in the event that this `After` fails? Does cucumber bother to execute the remaining scenarios?

features/step_definitions/collection.js
```diff
@@ -0,0 +1,267 @@
...
+                body,
+                json: true
+            });
+        });
+}
+
+/**
+ * Helper function to delete a collection
+ * @method deleteCollection
+ * @param   {Number} id  Id of the collection to delete
```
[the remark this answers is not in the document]
r3rastogi: Wait, why is this optional? You can't delete a collection if you don't know its ID

features/step_definitions/collection.js
joelseq: Sorry I was referencing some other tests when writing this and it had a check for `409` but I see now why that was allowed in those tests. I'll make the change

features/step_definitions/collection.js
```diff
@@ -0,0 +1,267 @@
...
+
+    return request({
+        uri: `${this.instance}/${this.namespace}/collections/${id}`,
+        method: 'DELETE',
+        auth: {
+            bearer: this.jwt
+        },
+        json: true
+    }).then((response) => {
+        Assert.oneOf(response.statusCode, [204, 404]);
```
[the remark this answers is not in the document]
joelseq: I had this before I refactored to check whether `id` exists and forgot to change this part.

features/step_definitions/collection.js
joelseq: How do I send data from that step to this one?

features/step_definitions/collection.js
joelseq: So what is the difference between the cases "they create a new collection..." and "they have a collection..."?

features/step_definitions/collection.js
joelseq: Just tested it out locally and yes it executes all the scenarios even if the `After` fails.

features/step_definitions/collection.js
r3rastogi: Agreed with @joelseq . When I read this scenario, I don't think it matters if they just created myCollection or it was created earlier, just as long as it exists in their list of collections

features/step_definitions/collection.js
```diff
@@ -0,0 +1,273 @@
...
+            json: true
+        }).then((response) => {
+            Assert.strictEqual(response.statusCode, 200);
+            this.collections = response.body;
+        });
+    });
+
+    Then(/^they can see those collections$/, function step() {
+        Assert.strictEqual(this.collections.length, 2);
+        Assert.strictEqual(this.collections[0].name, 'anotherCollection');
```
r3rastogi: Do we need to check specifically against IDs? Should we just check if they have both `anotherCollection` and `myCollection`, not necessarily, that `anotherCollection` is ID 0 and `myCollection` is ID 1?

features/step_definitions/collection.js
```diff
@@ -0,0 +1,273 @@
...
+                bearer: this.jwt
+            },
+            json: true
+        }).then((response) => {
+            Assert.strictEqual(response.statusCode, 200);
+            Assert.strictEqual(response.body.scmRepo.name, `${this.testOrg}/${this.repoName}`);
+        });
+    });
+
+    When(/^they create another collection "myCollection"$/, function step() {
```
r3rastogi: Maybe we should rename this scenario to "When they create another collection with the same name 'myCollection'"

features/step_definitions/collection.js
```diff
@@ -0,0 +1,273 @@
...
+}
+
+/**
+ * Helper function to delete a collection
+ * @method deleteCollection
+ * @param   {Number} [id]  Id of the collection to delete
+ * @returns {Promise}
+ */
+function deleteCollection(id) {
+    if (!id) {
```
minz1027: Why you need the check here and resolve?

features/step_definitions/collection.js
joelseq: I don't think we need to specifically check against IDs. I wrote it this way because the response sends them in descending order so if `anotherCollection` was created after `myCollection` it would be the first collection sent back but I'm not sure if that needs to strictly be checked against. I'll change it to just ensure that one of them is `anotherCollection` and the other is `myCollection`.

features/step_definitions/collection.js
```diff
@@ -0,0 +1,275 @@
...
+        return createCollection.call(this, { name: 'myCollection' })
+            .then((response) => {
+                Assert.oneOf(response.statusCode, [409, 201]);
+
+                if (response.statusCode === 201) {
+                    this.firstCollectionId = response.body.id;
+                } else {
+                    const str = response.body.message;
+
+                    [, this.firstCollectionId] = str.split(': ');
```
minz1027: You won't be able to parse the id out from the error object.

joelseq: Good catch! I fixed it now on the API side of things to add the id in the message

PR #738, opened 2017-09-11
fix: regexp dos with hapi-auth-jwt

petey:
Context:
========
`hapi-auth-jwt` requires an old version of jsonwebtoken (5.x), which depends on ms which has a ReDoS vulnerability.

Objective:
----------
Replace `hapi-auth-jwt` with `hapi-auth-jwt2` which is being actively maintained.

### DIFF
--- a/plugins/auth/index.js
+++ b/plugins/auth/index.js
@@ -3,7 +3,7 @@
 const authToken = require('hapi-auth-bearer-token');
 const bell = require('bell');
 const sugar = require('hapi-auth-cookie');
-const authjwt = require('hapi-auth-jwt');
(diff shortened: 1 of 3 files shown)


plugins/auth/index.js
```diff
@@ -117,6 +117,11 @@ exports.register = (server, options, next) => {
                 verifyOptions: {
                     algorithms: [ALGORITHM],
                     maxAge: EXPIRES_IN
+                },
+                // This function is run once the Token has been decoded with signature
+                validateFunc(decoded, request, cb) {
+                    // TODO: figure out what to do here
```
tkyi: Hmm

PR #739, opened 2017-09-13
feat: allow configuring cors origins

jithin1987:
## Context

Allows configuring CORS origins 

## Objective

This will allow user to configure which all Origins can connect to SD API via Ajax.

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -198,3 +198,5 @@ ecosystem:
     badges: https://img.shields.io/badge/build-{{status}}-{{color}}.svg
     # Default registry to pull build containers from. Uses Docker Hub if nothing/empty string is provided
     dockerRegistry: ""
+    # Extra origins allowed to do CORS to API
+    allowCors: []
--- a/lib/server.js
+++ b/lib/server.js
@@ -62,10 +62,15 @@ function prettyPrintErrors(request, reply) {
 module.exports = (config) => {
     // Hapi Cross-origin resource sharing configuration
     // See http://hapijs.com/api for available options
+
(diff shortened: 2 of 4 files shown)


config/default.yaml
```diff
@@ -198,3 +198,5 @@ ecosystem:
     badges: https://img.shields.io/badge/build-{{status}}-{{color}}.svg
     # Default registry to pull build containers from. Uses Docker Hub if nothing/empty string is provided
     dockerRegistry: ""
+    # Extra origins allowed to do CORS to API
+    allowCors: []
```
d2lam: I think we should set the default value to be same as ecosystem.ui

lib/server.js
```diff
@@ -62,10 +62,15 @@ function prettyPrintErrors(request, reply) {
...
+
+    if (Array.isArray(config.ecosystem.allowCors)) {
+        corsOrigins = corsOrigins.concat(config.ecosystem.allowCors);
+    }
+
     const cors = {
-        origin: [
-            config.ecosystem.ui
-        ],
+        origin: corsOrigins,
```
d2lam: and here just set it to config.ecosystem.allowCors

config/default.yaml
jithin1987: ecosystem.ui is already in by default. This for extra origins. I didn't want users to repeat the ui value here again.

lib/server.js
jithin1987: This way we will be forcing users to repeat value of ecosystem.ui again here. That's just duplication and won't be backward compatible.

jithin1987:
@minz1027 added unit test

PR #740, opened 2017-09-14
[RFC] Commands (Sharing Binaries) and Build Orchestration

stjohnjohnson:
## Context

The Design section of our repository should be describing how each of these concepts should work eventually.  That gives us something to slowly increment towards and a place of common discussion.

Right now, we don't have a documented story for how users should share executables with other pipelines or how the build should work.

## Objective

Describe how shared executables should work in Screwdriver.  And describe how the build should execute.


design/director.md
```diff
@@ -0,0 +1,104 @@
...
+```
+
+Here is the high-level overview:
+
+ 1. **Initialization** - Start required services and load information about the build.
+ 2. **Setup** - Ensure the environment is configured and loaded (git clone, environment variables, etc.).
+ 3. **User Steps** - Run all user steps inside a single shell.
+ 4. **Teardown** - Capture artifacts and publish final status.
+
+## Details
```
petey: Firstly, I think this is a really good writeup of the direction for the launcher. I'd like a visual diagram of all the shells lifecycles. 

Secondly, and this may be diving too deep into implementation details, but it would be good to talk a little about how we share environment between these shells. For example, if bookends initialize environment variables, those should be available in the user steps, even though it is a different shell.

stjohnjohnson: Good idea!  I added a visual diagram for the lifecycles.

Regarding environment variables, I was going to document that heavily in the Bookends doc, but I added a line item on how they get shared here.

design/commands.md
```diff
@@ -0,0 +1,201 @@
...
+- 11 September 2017: Proposal submitted
+
+## Proposal
+
+Have a single interface for executing a versioned command (via remote binary, docker image, or habitat package) during a Screwdriver build.  Managed and discoverable through the Screwdriver interface.
+
+This provides a single abstraction for execution of shared code.
+
+```bash
+$ sd_cmd chefdk/knife@12 search node '*:*'
```
d2lam: We should make this `sd-cmd` instead of `sd_cmd` to be consistent with shared steps.

design/commands.md
```diff
@@ -0,0 +1,201 @@
...
+
+ - **Tags:** `latest` `stable` `feature-abc`
+
+    Matches the exact version number that has been assigned to that tag. *Note: `latest` is automatically the most recently published version.*
+
+### Formats
+
+When a user wants to execute the command, we need to translate that to the command's native format.  We have three proposed command formats and it can easily be extended.
+
+For each of these, the arguments pass direectly to the underlying technology.  Here is our example:
```
d2lam: `directly`

design/commands.md
```diff
@@ -0,0 +1,201 @@
...
+namespace: foo
+# Command name itself
+command: bar
+# Description of the command and what it does
+description: |
+  Lorem ipsum dolor sit amet.
+# Major and Minor version number (patch is automatic)
+version: 1.0
+# Format the command is in (see below)
+format: habitat
```
d2lam: Is this supposed to be an array of available formats?
Why are `docker` & `binary` config defined, but `format` doesn't include them?

design/commands.md
```diff
@@ -0,0 +1,201 @@
...
+
+ - **Caret Ranges:** `^1.2.3` `^0.2.5` `^0.0.4`
+
+    Allows changes that do not modify the left-most non-zero digit in the [major, minor, patch] tuple. In other words, this allows patch and minor updates for versions 1.0.0 and above, patch updates for versions 0.X >=0.1.0, and no updates for versions 0.0.X.
+
+ - **Explicit Pinning:** `1.2.3` `1.5.3`
+
+    Allows only that exact version number, nothing else.
+
+ - **Tags:** `latest` `stable` `feature-abc`
```
d2lam: This needs to be alphanumeric. Otherwise, if tag is `123`, we won't know if it's referring to tag 123 or version 123?

design/commands.md
stjohnjohnson: This is to replace shared steps, so `sd-step` will be removed after this is implemented.  I prefer the `_` instead of `-` personally, although there is no limitation to using one vs. the other.

design/commands.md
stjohnjohnson: No, the items below are supposed to be an example.  I'll update it.

design/commands.md
stjohnjohnson: Hmm, `npm` checks if the tag is valid semver.  If it is, it rejects it: https://docs.npmjs.com/cli/dist-tag#caveats

We can either follow that or require that it starts with a letter and only matches `/[a-z][a-z0-9-]+/`

stjohnjohnson:
@d2lam updated

design/commands.md
d2lam: Yeah I think we should check, similar to how template tags are.

PR #746, opened 2017-09-22
feat: ignore commits made by certain users (bots)

d2lam:
## Context
Some projects would have bots commit back to the repo, for example, to update a package.json, etc. This will cause builds to run indefinitely. 

This PR allows ignoring commits made by certain users

### DIFF
--- a/config/custom-environment-variables.yaml
+++ b/config/custom-environment-variables.yaml
@@ -190,6 +190,10 @@ scms:
 webhooks:
     # Obtains the SCM token for a given user. If a user does not have a valid SCM token registered with Screwdriver, it will use this user's token instead.
     username: SCM_USERNAME
+    # Ignore commits made by these users
+    ignoreCommitsBy:
+        __name: IGNORE_COMMITS_BY
+        __format: json
 
 bookends:
     # List of module names, or objects { name, config } for instantiation to use in sd-setup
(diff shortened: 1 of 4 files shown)


plugins/webhooks/index.js
```diff
@@ -343,6 +345,10 @@ function pushEvent(pluginOptions, request, reply, parsed) {
                 return reply().code(204);
             }
 
+            if (ignoreUsername.includes(username)) {
```
minz1027: Let's log something like this before it's returned so that it's easier to debug.
https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/webhooks/index.js#L419

config/custom-environment-variables.yaml
```diff
@@ -190,6 +190,10 @@ scms:
 webhooks:
     # Obtains the SCM token for a given user. If a user does not have a valid SCM token registered with Screwdriver, it will use this user's token instead.
     username: SCM_USERNAME
+    # Ignore commits made by these users
+    ignore_commits_by:
```
jithin1987: This uses _casing, noticed we are using camel casing everywhere ?

d2lam: @jithin1987 good catch! Fixed

Issue #723, opened 2017-08-29
Support new Workflow config


petey:
# Context
As a pipeline owner,
I want to describe my pipeline from the job perspective,
So I do not need to write more than I have to to define a workflow.

## Objective
Using the new workflow spec, implement the serial flow of jobs, A->B->C, as it currently exists with the current workflow system. If both new and old workflow definitions exist in the same yaml, then the new workflow style takes precedence. Users of the old workflow spec should be shown a deprecation warning.

## Tasks
- [x] Publish the new workflow spec  (https://github.com/screwdriver-cd/screwdriver/blob/master/design/workflow.md)
- [x] Display a deprecation warning in build (and UI?) when a yaml defines `workflow`
- [x] Jobs defined with `requires: ~pr` are started by `pull-request` events
- [x] Jobs defined with `requires: ~commit` are started by `push` events
- [x] Jobs defined with `requires: [A]` are started after job `A` finishes successfully
- [x] Jobs without `requires` are ignored.

d2lam:
## 9/29
We decided to not use the field `workflow` to decide whether it's a new or old workflow version. 

The reasons is:
Currently, we allow implicit workflow. That is, users can have 1 job (`main`) and not defined the `workflow` keyword. It will will be interpreted as `workflow: ['main']`. Or users can have a list of job, and the workflow will be interpreted as the same order that the jobs are defined. 

If we use the keyword `workflow` to decide whether it is a new/old version, then it will be ambiguous whether we want it to be a detached job (as described in the design doc), or we want it to be `requires:~commit`, `requires:A` 

### Our decision: 
**Check if any jobs has the `requires` keyword. If it doesn't, then we treat it like the old version.**

d2lam:
## 10/05
We decided to have a new endpoint to create Event. The event will have a new field called `startFrom`. 
This field is a string that is a jobName `A` or a requires value like `~commit`
The UI can make call this endpoint when the user clicks start (`startFrom` will be `~commit` in this case)
With this change, we will get partial events.

stjohnjohnson:
@petey what is left to do here?

petey:
We need to officially deprecate and remove all the code for the old workflow system.

d2lam:
**4/9/18** 
I will work on deprecating the old workflow.

marian-r:
Is #874 going to be part of this? I'm asking as it is being referenced there.

d2lam:
**Update 4/10/18**
[...]

PR #755, opened 2017-10-04
feat: support new workflow in push event

minz1027:
Support new workflow in `push` event:

If the pipeline has `workflow: undefined` then it's using the new workflow:
- Create new event with `workflow` field as `undefined` for now, will do `source, destination` later
- Start all jobs with `requires: '~commit' `

Added a test for `auth/token.js` to make coveralls happy!

Related to https://github.com/screwdriver-cd/screwdriver/issues/723
Blocked by: https://github.com/screwdriver-cd/models/pull/195

### DIFF
--- a/plugins/webhooks/index.js
+++ b/plugins/webhooks/index.js
@@ -4,6 +4,47 @@ const boom = require('boom');
 const joi = require('joi');
 
 /**
+ * Create a pipeline event amd start the job
+ * @method startCommitJob
+ * @param  {Object}       config            Configuration for starting a commit job
+ * @param  {Object}       config.request    Request object
+ * @param  {Object}       config.parsed     Parsed webhook payload
+ * @param  {Object}       config.pipeline   Pipeline to be triggered
+ * @param  {Object}       config.job        Job to be triggered
+ * @return {Promise}                        Resolves to null if successfully start the build
+ */
+function startCommitJob(config) {
+    const { request, parsed, pipeline, job } = config;
+    const buildFactory = request.server.app.buildFactory;
+    const eventFactory = request.server.app.eventFactory;
+    const hookId = parsed.hookId;
+    const sha = parsed.sha;
+    const username = parsed.username;
+    const scmContext = parsed.scmContext;
+
+    // create an event
+    return eventFactory.create({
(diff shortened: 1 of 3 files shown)


plugins/webhooks/index.js
```diff
@@ -346,33 +375,32 @@ function pushEvent(pluginOptions, request, reply, parsed) {
...
-                    const name = 'main';
-                    const i = jobs.findIndex(j => j.name === name); // get job's index
-                    const jobId = jobs[i].id;
+                    let commitJobs = [];
+
+                    // if it's using old workflow then find the main job
+                    if (p.workflow) {
+                        commitJobs = jobs.filter(j => j.name === 'main');
+                    } else {
+                        commitJobs = jobs.filter(j => j.requires && j.requires.includes('~commit'));
```
d2lam: actually `requires` is not stored in the job object itself, so I don't think we can get the value out of `j.requires`, will probably need to get it from `permutations`

minz1027: Ah, okay, this can be ugly..

plugins/webhooks/index.js
```diff
@@ -343,37 +342,24 @@ function pushEvent(pluginOptions, request, reply, parsed) {
...
-            return pipeline.sync()
-                // handle the PR action
-                .then(p => p.jobs.then((jobs) => {
-                    const pipelineId = p.id;
-                    const name = 'main';
-                    const i = jobs.findIndex(j => j.name === name); // get job's index
-                    const jobId = jobs[i].id;
+            const eventConfig = {
+                pipelineId: pipeline.id,
+                type: 'pipeline',
```
tkyi: Technically, you don't need to pass this in because eventFactory will default `type` to 'pipeline' if none is passed in.

test/plugins/auth.test.js
```diff
@@ -618,6 +618,22 @@ describe('auth plugin test', () => {
                     assert.notOk(reply.result.token, 'Token not returned');
                 })
             ));
+
+            it('catch err if buildFactory throws err', () => {
```
tkyi: 👍

PR #762, opened 2017-10-18
feat: Add POST route for creating events

tkyi:
## Context

With the new workflow implementation, users should be able to start a pipeline from any job (instead of only the first one).

## Objective

This PR adds a `POST` route for creating events.

## References

Related to: https://github.com/screwdriver-cd/screwdriver/issues/723


plugins/events/create.js
```diff
@@ -0,0 +1,108 @@
...
+                                scmContext,
+                                scmUri: pipeline.scmUri,
+                                token
+                            };
+
+                            // Get commit sha
+                            return scm.getCommitSha(scmConfig)
+                                .then((sha) => {
+                                    // For PRs
+                                    if (prNum) {
```
jithin1987: what will trigger this flow ? Is'nt all PR event triggering happening via webhooks ?

plugins/events/create.js
```diff
@@ -0,0 +1,108 @@
...
+            const startFrom = request.payload.startFrom;
+            const username = request.auth.credentials.username;
+            const payload = {
+                pipelineId,
+                scmContext,
+                startFrom,
+                type: 'pipeline',
+                username
+            };
+            // Match PR-prNum, then extract prNum
```
minz1027: Can we give an example for prFullName in the commet?

test/plugins/events.test.js
```diff
@@ -164,4 +184,111 @@ describe('event plugin test', () => {
...
+            const pipelineId = 123;
+            const scmContext = 'github:github.com';
+            const scmUri = 'github.com:12345:branchName';
+            const checkoutUrl = 'git@github.com:screwdriver-cd/data-model.git#master';
+            const pipelineMock = {
+                id: pipelineId,
+                checkoutUrl,
+                scmUri,
+                sync: sinon.stub().resolves(),
+                syncPR: sinon.stub().resolves()
```
minz1027: No need for these functions `sync` and `syncPR`

PR #763, opened 2017-10-18
feat: trigger next job using workflow parser

minz1027:
- Trigger next job using workflowGraph instead of workflow

When a build tries to update its status with `PUT /builds`, the pipeline must synced already and has the `workflowGraph` field. So this change is backward compatible and will work for both legacy and new config.

Will switch to user `pub-sub` to handle triggers in next pass.

Related to ttps://github.com/screwdriver-cd/screwdriver/issues/723

### DIFF
--- a/plugins/builds/update.js
+++ b/plugins/builds/update.js
@@ -21,6 +21,7 @@ module.exports = () => ({
             const buildFactory = request.server.app.buildFactory;
             const id = request.params.id;
             const desiredStatus = request.payload.status;
+            const eventFactory = request.server.app.eventFactory;
             const jobFactory = request.server.app.jobFactory;
             const username = request.auth.credentials.username;
             const scmContext = request.auth.credentials.scmContext;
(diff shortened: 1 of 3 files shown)


plugins/builds/update.js
```diff
@@ -86,28 +87,13 @@ module.exports = () => ({
...
-                                return null;
-                            }
-
-                            const nextJobName = workflow[workflowIndex + 1];
+                            const nextJobs = workflowParser.getNextJobs(workflowGraph, {
+                                trigger: job.name
+                            });
 
-                            return jobFactory.get({
+                            return Promise.all(nextJobs.map(nextJobName => jobFactory.get({
```
tkyi: Probably want to add a check in case `nextJobs` is an empty array.

tkyi: Nevermind, I see that `nextJobToTrigger` is not the result of the `Promise.all`

plugins/builds/update.js
```diff
@@ -86,28 +87,13 @@ module.exports = () => ({
                                 return null;
                             }
 
-                            const workflow = pipeline.workflow;
+                            const workflowGraph = pipeline.workflowGraph;
```
jithin1987: is this coming form event ? We would probably want to retrieve current workflow from event which this build is part of. I also think that event should have the status of all builds (jobs), so that when use case like fan-in is implemented, this API will look at status of workflowGraph in event to figure out what needs to be triggered as part of workflow for the event.

minz1027: Good point, will fix it.

PR #772, opened 2017-10-25
feat: support notification scoped package

yuichi10:
## Context
Now we can't use scoped notification package.
Due to this change user can use scoped notification package by setting "scope" in config file.

## Objective
Change how to require notification packages.
If you set "scope" at notifications config, you can use scoped notifications package. Also you can set scope value both "@xxx" and "xxx".

### DIFF
--- a/lib/registerPlugins.js
+++ b/lib/registerPlugins.js
@@ -64,6 +64,33 @@ function registerResourcePlugins(server, config, callback) {
     }, callback);
 }
 
+/**
+ * Revise npm packages's scope name
+ * @method reviseScope
+ * @param {String} scope            Name of npm package's scope        
+ */
+function reviseScope(scope) {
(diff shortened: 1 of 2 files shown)


lib/registerPlugins.js
```diff
@@ -64,6 +64,33 @@ function registerResourcePlugins(server, config, callback) {
...
+
+    return `@${scope}`;
+}
+
+/**
+ * Require notification plugin
+ * @method requireNotificationPlugin
+ * @param  {Object}    config            Configuration object for notification plugins.
+ * @param  {String}    plugin            Notification plugin name.
+ */
```
tkyi: missing @return JSDoc

yuichi10: Sorry I forget that.
I will fix.

yuichi10:
OK!

PR #777, opened 2017-11-01
fix: fix triggerFactory.list config format

minz1027:
fix triggerFactory.list config format

https://github.com/screwdriver-cd/models/blob/master/lib/baseFactory.js#L109

### DIFF
--- a/plugins/builds/update.js
+++ b/plugins/builds/update.js
@@ -115,8 +115,9 @@ module.exports = () => ({
 
                                     return null;
                                 }))))
-                                .then(() => triggerFactory.list(
-                                    { src: `~sd@${pipeline.id}:${job.name}` }))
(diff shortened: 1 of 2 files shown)


plugins/builds/update.js
```diff
@@ -115,8 +115,9 @@ module.exports = () => ({
 
                                     return null;
                                 }))))
-                                .then(() => triggerFactory.list(
-                                    { src: `~sd@${pipeline.id}:${job.name}` }))
+                                .then(() => triggerFactory.list({
+                                    param: { src: `~sd@${pipeline.id}:${job.name}` }
```
d2lam: params?

minz1027:
@jithin1987 test added :)

Issue #700, opened 2017-08-16
Workflow Fan-in


petey:
# Context
As a screwdriver user,
I want to be able to fan in jobs in my workflow,
so I run a test on all deployed regions in one job, after those regions were deployed in parallel.

## Objective
Users can define a workflow that has one jobs that is triggered by the completion of multiple other jobs [with the same initial trigger] finishing successfully.

### Tasks
- [x] Fully document the final definitions of a job fan in in the guide for workflows
- [ ] Queue a job when all the "dependent" jobs have  completed successfully
    - [ ] Engine changes
    - [ ] Worker changes(?)

d2lam:
This feature is done

PR #781, opened 2017-11-03
feat(700): support join trigger

d2lam:
Support Join Trigger

Example:
```
A: 
   requires: [~B, C,D]
```

joinList of A = [C,D]
- When B is successful, trigger A (since B is not in the joinList)
- When C is successful, check if D is successful. If yes, trigger A. If no, do nothing. 

Related: https://github.com/screwdriver-cd/screwdriver/issues/700

### DIFF
--- a/plugins/builds/index.js
+++ b/plugins/builds/index.js
@@ -7,6 +7,56 @@ const stepGetRoute = require('./steps/get');
 const stepUpdateRoute = require('./steps/update');
 const stepLogsRoute = require('./steps/logs');
 const listSecretsRoute = require('./listSecrets');
+const workflowParser = require('screwdriver-workflow-parser');
+
+/**
+ * Start the build
+ * @method startBuild
+ * @param  {Object}   config                Configuration object
+ * @param  {Factory}  config.jobFactory     Job Factory
+ * @param  {Factory}  config.buildFactory   Build Factory
+ * @param  {Number}   config.pipelineId     Pipeline Id
+ * @param  {String}   config.jobName        Job name
+ * @param  {String}   config.username       Username of build
+ * @param  {String}   config.scmContext     SCM context
+ * @param  {Build}    config.build          Build object
+ * @return {Promise}
+ */
+function startBuild(config) {
+    const { jobFactory, buildFactory, pipelineId, jobName, username, scmContext, build } = config;
+
+    return jobFactory.get({
+        name: jobName,
+        pipelineId
+    }).then((job) => {
+        if (job.state === 'ENABLED') {
+            return buildFactory.create({
+                jobId: job.id,
+                sha: build.sha,
+                parentBuildId: build.id,
+                eventId: build.eventId,
+                username,
+                scmContext
+            });
+        }
+
+        return null;
+    });
+}
+
+/**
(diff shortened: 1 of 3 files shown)


plugins/builds/index.js
```diff
@@ -70,6 +121,52 @@ exports.register = (server, options, next) => {
             });
     });
 
+    server.expose('triggerNextJobs', (config) => {
```
tkyi: JSdoc?

plugins/builds/index.js
```diff
@@ -70,6 +121,52 @@ exports.register = (server, options, next) => {
...
+        const eventFactory = server.root.app.eventFactory;
+        const jobFactory = server.root.app.jobFactory;
+        const buildFactory = server.root.app.buildFactory;
+        const { pipeline, job, build, username, scmContext } = config;
+        const currentJobName = job.name;
+        const pipelineId = pipeline.id;
+
+        return eventFactory.get({ id: build.eventId }).then((event) => {
+            const workflowGraph = event.workflowGraph;
+            const nextJobs = workflowParser.getNextJobs(workflowGraph, { trigger: job.name });
```
tkyi: Can use `trigger: currentJobName`

plugins/builds/index.js
```diff
@@ -7,6 +7,57 @@ const stepGetRoute = require('./steps/get');
...
+
+    return jobFactory.get({
+        name: jobName,
+        pipelineId
+    }).then((job) => {
+        if (job.state === 'ENABLED') {
+            return buildFactory.create({
+                jobId: job.id,
+                sha: build.sha,
+                parentBuildId: build.id,
```
tkyi: Is `parentBuildId` needed?

tkyi: Nvm I see it in the database.
We should update the docs here at some point: https://github.com/screwdriver-cd/models/blob/master/lib/buildFactory.js#L145-L156

d2lam: Yep. parentBuildId is the build that triggers this one.

test/plugins/builds.test.js
```diff
@@ -746,6 +747,161 @@ describe('build plugin test', () => {
                     });
                 });
             });
+
+            describe('join', () => {
```
jithin1987: why is this test suite nested inside `buildToken` ?

minz1027: The `PUT /builds/id` has two scopes, `user token` and `build token`. User can only update build status to `ABORTED`. Build can do all the stuff.

plugins/builds/index.js
```diff
@@ -7,6 +7,57 @@ const stepGetRoute = require('./steps/get');
...
+        return null;
+    });
+}
+
+/**
+ * Check if all the jobs in joinList are successful
+ * @method isJoinDone
+ * @param  {Array}      joinList       array of jobs(name,id) that are in join
+ * @param  {Array}      finishedBuilds array of finished builds belong to this event
+ * @param  {Object}     workflowGraph  the event workflowGraph
```
minz1027: Doesn't look like you need this

d2lam: 👍  Was leftover from before!

PR #784, opened 2017-11-06
fix: use TEST_SCM_HOSTNAME environment variable for feature test

tk3fftk:
## Context
The hostname of `checkoutUrl ` is hardcoded in the functional test. So if we want to run functional tests in own SCM, we must change the feature files.

## Objective
This PR provides `TEST_SCM_HOSTNAME` environment variable. It is used to determine checkout url. The default value is `github.com`.


d2lam:
Thanks Hiroki! 👍

Issue #768, opened 2017-10-23
Build job on an a trigger


jithin1987:
# Context

A job in a pipeline should be able to express that it should be built in response to successful completion another job (within same pipeline or external pipeline) ~~or successful completion in entirety of an external pipeline.~~ We will enhance `requires` syntax to support `~` which expresses an `ANY` relationship to a job in `requires`

Eg:  `Y : requires: [A, B, ~sd@pipeX:jobX]` means job `Y` should run whenever jobs `A` and `B` in same pipeline completes or whenever `jobX` in an external pipeline `pipeX` completes successfully.

# Objective

1. `jobA : requires: [~sd@pipeX:jobX` should  run job jobA whenever `jobX in pipeline pipeX completes. https://github.com/screwdriver-cd/screwdriver/issues/770
2. If multiple jobs in a pipeline has same `ANY` relationship  on an external job/piepline it should create a single pipeline event. Same as how `requires` work in context of `~commit` on multiple jobs.
3. A job which was triggered by `requires` on an external pipeline's job should have access to meta set by the build of the external pipeline job which started the event. - https://github.com/screwdriver-cd/screwdriver/issues/771
4. Warn users if an `ANY` relationship leads to a detached job run. Eg: `A: requires: [~commit], X: requires[~sd@pipeX:jobX, A], Y: requires[A], Z: requires: [X, Y]` (An advanced Fan-out use case, we can address this when we take up Fan-out feature)

# References
1. Original feature request https://github.com/screwdriver-cd/screwdriver/issues/433
2. Workflow config  @https://github.com/screwdriver-cd/screwdriver/blob/master/design/workflow.md#rebuild_on

d2lam:
I'm not clear on the job criteria. Why does job X need to be a detached job? or why no other jobs should require job X? I think you have some scenario you want to avoid, but it's not very clear from reading it. Why do we want such criteria?

Also, how do we define "completion of an external pipeline"? We need to figure out what the last job is for that pipeline? I think for this, it's actually easier to tie it to a job. Like `rebuild_on: sd@123:staging`

minz1027:
Agreed with @d2lam's comment about `completion of an external pipeline`. It's hard to define. For example, a pipeline has two flows, one: [A -> B], two: [C -> D]. What is the completion of the the pipeline in that case?

d2lam:
[...]

PR #787, opened 2017-11-06
feat: add feature for trigger

jithin1987:
## Context

Adds a feature for for trigger(rebuild_on) functionality 

## References

https://github.com/screwdriver-cd/screwdriver/issues/768

### DIFF
--- a/features/trigger.feature
+++ b/features/trigger.feature
@@ -1,0 +1,54 @@
+@ignore
+Feature: Trigger
+
+    Users should be allowed to tie together workflows from different build pipelines so that they can
+    express dependencies between otherwise unrelated projects.
+
+    Eg: PipeLine A for Project Foo
+        (~commit) -> (build_foo) -> (deploy_foo)
+        Pipeline B for Project Bar
+        (~commit) -> (build_bar) -> (deploy_bar)
+
+        Assume Project Bar depends on Project Foo. A user should be able to build Project Bar
+        when Project Foo has been built. A user can also make it specific to a job so that
+        Project Bar -> (build_bar) to be built when Project Foo -> (deploy_foo) is
+        finished successfully.
+
+        Combined Pipeline Workflow
+        (~commit) -> (build_foo) -> (deploy_foo) -> Foo (build_bar) -> (deploy_bar)
+
+    Rules:
+        - Connected Workflows must be one-directional (no loops)
+        - Configuration for trigger must reside in Workflow/Job being triggered
+        - Triggered workflow/job should start with latest commit on configured SCM branch.
+        - If multiple jobs in a pipeline requires the same external pipeline's Job as trigger, then
+          builds for these jobs should be part of same pipeline event.
+
+    Scenario: Failure
+        Given two pipelines PipelineA and PipelineB with following config
+            | job           | requires      |
+            | PipelineB:BAR | PipelineA:FOO |
+        When the "FOO" job in PiplelineA is started
+        And the build failed
+        Then the "BAR" job in PipelineB is not started
+
+    Scenario: Success
+        Given two pipelines PipelineA and PipelineB with following config
+            | job           | requires      |
+            | PipelineB:BAR | PipelineA:FOO |
+        When the "FOO" job in PiplelineA is started
+        And the build succeeded
+        Then the "BAR" job in PipelineB is started
+        And that "BAR" build uses the same SHA as latest commit on the branch pipeline is configured
+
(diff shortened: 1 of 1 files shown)


features/trigger.feature
```diff
@@ -0,0 +1,54 @@
...
+        (~commit) -> (build_foo) -> (deploy_foo) -> Foo (build_bar) -> (deploy_bar)
+
+    Rules:
+        - Connected Workflows must be one-directional (no loops)
+        - Configuration for trigger must reside in Workflow/Job being triggered
+        - Triggered workflow/job should start with latest commit on configured SCM branch.
+        - If multiple jobs in a pipeline requires the same external pipeline's Job as trigger, then
+          builds for these jobs should be part of same pipeline event.
+
+    Scenario: Failure
```
petey: Can this be a little more descriptive? Like: `Subsequent builds are not triggered after a build failure`

features/trigger.feature
```diff
@@ -0,0 +1,54 @@
...
+        Given two pipelines PipelineA and PipelineB with following config
+            | job           | requires      |
+            | PipelineB:BAR | PipelineA:FOO |
+        When the "FOO" job in PiplelineA is started
+        And the build failed
+        Then the "BAR" job in PipelineB is not started
+
+    Scenario: Success
```
petey: Reverse Trigger: `Builds are triggered after another build is successful`

features/trigger.feature
```diff
@@ -0,0 +1,54 @@
...
+        Assume Project Bar depends on Project Foo. A user should be able to build Project Bar
+        when Project Foo has been built. A user can also make it specific to a job so that
+        Project Bar -> (build_bar) to be built when Project Foo -> (deploy_foo) is
+        finished successfully.
+
+        Combined Pipeline Workflow
+        (~commit) -> (build_foo) -> (deploy_foo) -> Foo (build_bar) -> (deploy_bar)
+
+    Rules:
+        - Connected Workflows must be one-directional (no loops)
```
petey: This is really difficult to determine when there are multiple pipelines involved. I do not think we are detecting a looping case where `A (build_bar) -> B (build_bar) -> A (build_bar)`. Also, it doesn't appear that there is a feature test here specifically about loops.

features/trigger.feature
```diff
@@ -0,0 +1,54 @@
...
+    Scenario: Success
+        Given two pipelines PipelineA and PipelineB with following config
+            | job           | requires      |
+            | PipelineB:BAR | PipelineA:FOO |
+        When the "FOO" job in PiplelineA is started
+        And the build succeeded
+        Then the "BAR" job in PipelineB is started
+        And that "BAR" build uses the same SHA as latest commit on the branch pipeline is configured
+
+    Scenario: Multiple
```
petey: Reverse trigger Fan-out: `Multiple builds triggered in parallel as a result of build success` I wonder if it is worth adding a line that the `BAR` and `HELLO` builds belong to the same event?

features/trigger.feature
```diff
@@ -0,0 +1,54 @@
+@ignore
+Feature: Trigger
```
petey: There's actually at least three different features described in this feature file, I'm not sure if you want them all in one or want to break them up: rebuild on/reverse trigger, fan-out, and standard intra-workflow triggers. Will fan-in be added to this, or a different test? The scenarios in this file are mostly dealing with the rebuild on/reverse trigger feature.

jithin1987: Fan-in and Fan-out scenarios are different and is already in capture for `workflow`. Fan-out scenario is reemphasized here to capture the fact that there should be only one event. All the scenarios in this feature file arise from the successful completion of a single external pipeline's job.

features/trigger.feature
```diff
@@ -0,0 +1,54 @@
...
+        Combined Pipeline Workflow
+        (~commit) -> (build_foo) -> (deploy_foo) -> Foo (build_bar) -> (deploy_bar)
+
+    Rules:
+        - Configuration for trigger must reside in Workflow/Job being triggered
+        - Triggered workflow/job should start with latest commit on configured SCM branch.
+        - If multiple jobs in a pipeline requires the same external pipeline's Job as trigger, then
+          builds for these jobs should be part of same pipeline event.
+
+    Scenario: Reverse Trigger. Builds are not triggered if required build is not succesful.
```
petey: Minor nit. If you title the feature Reverse Trigger, it doesn't need to be in the scenarios.

PR #790, opened 2017-11-07
fix: use environment variable on github.js

tk3fftk:
## Context
In addition https://github.com/screwdriver-cd/screwdriver/pull/784, [node-github](https://github.com/octokit/node-github) should be also set configuration for using GHE.

## Objective
This PR adds configurations when create Github constractor. If `TEST_SCM_HOSTNAME` is not set, default values are used.

## References
https://github.com/octokit/node-github

### DIFF
--- a/features/support/github.js
+++ b/features/support/github.js
@@ -3,7 +3,8 @@
 const Github = require('github');
 const github = new Github({
     host: process.env.TEST_SCM_HOSTNAME || 'api.github.com',
-    pathPrefix: process.env.TEST_SCM_HOSTNAME ? '/api/v3' : ''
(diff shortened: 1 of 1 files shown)


features/support/github.js
```diff
@@ -1,7 +1,10 @@
 'use strict';
 
 const Github = require('github');
-const github = new Github();
+const github = new Github({
+    host: process.env.TEST_SCM_HOSTNAME || 'api.github.com',
+    pathPrefix: process.env.TEST_SCM_HOSTNAME ? '/api/v3' : ''
```
d2lam: should probably set a default value for `protocols` as well?

tk3fftk: added!

Issue #775, opened 2017-10-31
Bug: Notification packages fail to parse config when scope option is set


yuichi10:
**What happened**:

I sent this [PR](https://github.com/screwdriver-cd/screwdriver/pull/772). At that time, I just set scope in config and allow scoped notification packages. However, this will cause some issues that each notification plugin must allow a scope setting (scope key in config) when validatation.

So I think we should change the way to change the schema to the notification config.

My idea is as below.
```
email:
  config:
    host: ***
    port: ***
    from: ***
  module: '@module/package'
```
`module` is optional.  We use `module` as a notification package, when it is set.  If `module` is not set, we use _screwdriver-notification-xxx_ (same as the old way).
It makes constructor of notification package simpler, because we just pass `email.config` object. Also this looks similar to [a config for bookends](https://github.com/screwdriver-cd/build-bookend/blob/master/README.md).

If we will keep to use the [PR](https://github.com/screwdriver-cd/screwdriver/pull/772), we have to delete scope setting before calling notification plugin constructor. It makes the code a bit complex.

What should I do to fix this problem?

**reference**
- [previous PR](https://github.com/screwdriver-cd/screwdriver/pull/772)
- [notification packages validate place](https://github.com/screwdriver-cd/notifications-email/blob/master/index.js#L58)
- [bookends config](https://github.com/screwdriver-cd/build-bookend/blob/master/README.md)

tk3fftk:
This should be closed😃

PR #791, opened 2017-11-07
fix(775): add module field to notification modules.

tk3fftk:
## Context
https://github.com/screwdriver-cd/screwdriver/pull/772 enabled to use scoped packages for notification modules. But when scope is set, validation error is happened. 
I think we should not fix all notification modules. So this PR changes notifications schema.

## Objective
New notifications schema is like below:
```
email:
  config:
    host: foo
    port: bar
    from: baz
  scopedPackage: '@module/package'
```
If "scopedPackage" is set, the "module" is required. If "config" is set, the contents of "config" is passed to a notification module.
The previous schema is also accepted for backward compatibility.

## References
Related: https://github.com/screwdriver-cd/screwdriver/issues/775

### DIFF
--- a/lib/registerPlugins.js
+++ b/lib/registerPlugins.js
@@ -67,13 +67,14 @@ function registerResourcePlugins(server, config, callback) {
 /**
  * Require notification plugin
  * @method requireNotificationPlugin
- * @param  {Object}    config           Configuration object for notification plugins.
- * @param  {String}    plugin           Notification plugin name.
- * @return {Function}                   Function of notifier
+ * @param  {Object}    config                         Configuration object for notification plugins.
+ * @param  {String}    config.scopedPackage           Package name with scope (optional).
+ * @param  {String}    plugin                         Notification plugin name.
+ * @return {Function}                                 Function of notifier
  */
 function requireNotificationPlugin(config, plugin) {
-    if (config[plugin].module) {
-        return require(config[plugin].module);
+    if (config.scopedPackage) {
+        return require(config.scopedPackage);
     }
 
     return require(`screwdriver-notifications-${plugin}`);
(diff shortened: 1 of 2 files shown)


lib/registerPlugins.js
```diff
@@ -65,29 +65,15 @@ function registerResourcePlugins(server, config, callback) {
...
  * Require notification plugin
  * @method requireNotificationPlugin
  * @param  {Object}    config           Configuration object for notification plugins.
  * @param  {String}    plugin           Notification plugin name.
  * @return {Function}                   Function of notifier
  */
 function requireNotificationPlugin(config, plugin) {
-    if (config[plugin].scope) {
-        return require(`${reviseScope(config[plugin].scope)}/screwdriver-notifications-${plugin}`);
+    if (config[plugin].module) {
```
d2lam: I think the keyword `module` is kinda confusing here. This one is used only when you want to load some scoped package right? 
I'm not sure what's a better keyword. Maybe `scopedPackage`?

tk3fftk: I think this property is used only when we want to load scoped package.
Fixed to `scopedPackage`.

lib/registerPlugins.js
```diff
@@ -65,29 +65,15 @@ function registerResourcePlugins(server, config, callback) {
...
-        return scope;
-    }
-
-    return `@${scope}`;
-}
-
-/**
  * Require notification plugin
  * @method requireNotificationPlugin
  * @param  {Object}    config           Configuration object for notification plugins.
```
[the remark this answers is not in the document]
tk3fftk: Updated comment and I changed `notificationConfig[plugin]` is passed to this function https://github.com/screwdriver-cd/screwdriver/pull/791/commits/3935b93ff2ba31cd4d66f1a16ad52248b956ad7f#diff-bfa5689203892e653ca5d21a2b88658cR100

PR #797, opened 2017-11-14
feat: configuration flag for enabling Redis TLS communication

FenrirUnbound:
## Context

Although the Redis server may have a TLS proxy in front of it, the clients must also be configured to communicate to it via TLS. This change allows a configuration flag to be available for cluster maintainers to enable this feature.

This feature does _NOT_ support self-signed certificates.

## Objective

Allow a flag for configuring the Redis client to communicate with the Redis server via TLS.

### DIFF
--- a/config/custom-environment-variables.yaml
+++ b/config/custom-environment-variables.yaml
@@ -138,6 +138,7 @@ executor:
                 port: QUEUE_REDIS_PORT
                 options:
                     password: QUEUE_REDIS_PASSWORD
+                    tls: QUEUE_REDIS_TLS_ENABLED
                 database: QUEUE_REDIS_DATABASE
 
 scms:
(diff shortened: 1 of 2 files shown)


tkyi:
Do you need to set this variable in `in-a-box.py`?
https://github.com/screwdriver-cd/screwdriver/blob/master/in-a-box.py#L40

FenrirUnbound:
> Do you need to set the QUEUE_REDIS_TLS_ENABLED variable in in-a-box.py?
> https://github.com/screwdriver-cd/screwdriver/blob/master/in-a-box.py#L40

I'm inclined not to add it to `in-a-box`. Right now, `in-a-box` doesn't immediately configure itself for a queue-based implementation, so there isn't a dire need to add the Redis TLS flag there.

PR #814, opened 2017-12-05
fix: allow to search any build

yuichi10:
## Context
Now Screwdriver uses new workflow and not to need to use main. But in tests we can't find builds without main. 
Sometime this make it difficult to create new acceptance tests.

## Objective
Add jobName option to searchForBuild and allow user to find any job.


yuichi10:
Sorry I forgot to update that.  I will update.

PR #825, opened 2017-12-14
fix: Adding guid to JWT and changing timeout to 2h

tkyi:
## Context

We would like to issue one time use build JWT tokens for improved security.

## Objective

This PR adds a guid to the JWT token and changes the timeout from 12h to 2h.

## References

Design doc: https://github.com/screwdriver-cd/screwdriver/blob/master/design/auth.md
Related to: https://github.com/screwdriver-cd/hapi-auth-dynamic-jwt/blob/master/index.js#L163
jsonwebtoken npm: https://www.npmjs.com/package/jsonwebtoken
uuid npm: https://www.npmjs.com/package/uuid

### DIFF
--- a/plugins/auth/index.js
+++ b/plugins/auth/index.js
@@ -110,7 +110,7 @@ exports.register = (server, options, next) => {
 
             server.auth.strategy('session', 'cookie', {
                 cookie: 'sid',
-                ttl: 12 * 60 * 60 * 1000, // 12 hours in milliseconds
+                ttl: 2 * 60 * 60 * 1000, // 2 hours in milliseconds
                 password: pluginOptions.cookiePassword,
                 isSecure: pluginOptions.https
             });
--- a/plugins/auth/index.js
+++ b/plugins/auth/index.js
@@ -1,21 +1,22 @@
 'use strict';
 
+const authjwt = require('hapi-auth-jwt2');
 const authToken = require('hapi-auth-bearer-token');
 const bell = require('bell');
-const sugar = require('hapi-auth-cookie');
-const authjwt = require('hapi-auth-jwt2');
-const crumb = require('crumb');
-const jwt = require('jsonwebtoken');
-const joi = require('joi');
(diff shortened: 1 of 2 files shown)


plugins/auth/index.js
```diff
@@ -70,14 +71,15 @@ exports.register = (server, options, next) => {
...
+     * Generates a jwt that is signed and has a 2h lifespan
      * @method generateToken
      * @param  {Object} profile Object from generateProfile
      * @return {String}         Signed jwt that includes that profile
      */
     server.expose('generateToken', profile => jwt.sign(profile, pluginOptions.jwtPrivateKey, {
         algorithm: ALGORITHM,
-        expiresIn: EXPIRES_IN
+        expiresIn: EXPIRES_IN,
+        jwtid: guid()
```
petey: Since we are simply generating a random uuid each time, we aren't really using the features that aguid was built to perform. 

We should probably just use [uuid](https://www.npmjs.com/package/uuid) instead, as that is a self-contained implementation that does everything we want.

tkyi: I don't really have a preference for either module.
Updating

plugins/auth/index.js
```diff
@@ -108,7 +110,7 @@ exports.register = (server, options, next) => {
 
             server.auth.strategy('session', 'cookie', {
                 cookie: 'sid',
-                ttl: 12 * 60 * 60 * 1000, // 12 hours in milliseconds
+                ttl: 2 * 60 * 60 * 1000, // 2 hours in milliseconds
```
minz1027: Don't know if we should change this. If we change this, then users need to log in for every 2 hours.. it's kinda annoying

tkyi: I don't think this ttl is for the user session, so it should be okay

d2lam: Asked @petey about this. This is for transaction between API and github to make sure the calls come from the same user.

PR #826, opened 2017-12-15
fix: improve command format spec.

wdstar:
## Context

I am implementing commands data-schemas now.
https://github.com/screwdriver-cd/data-schema/pull/195

## Objective

I think we should add `command` optional key  to the Docker format that allows users not to set an extra executable (knife, ...).
And we should rename `binary` key of the Habitat format to `command`.
Please review it.

## References

https://github.com/screwdriver-cd/data-schema/pull/195#discussion_r157115519

### DIFF
--- a/design/commands.md
+++ b/design/commands.md
@@ -45,12 +45,14 @@ habitat:
     # If local
     # mode: local
     # package: ./foobar.hart
-    binary:  git
+    command: git
 
 # Docker specific config
 # if format: docker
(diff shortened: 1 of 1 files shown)


tk3fftk:
I think this PR and commit title should not start `feat`. It's not a feature release, just fix design docs.

wdstar:
I fixed titles. Thanks.

Issue #827, opened 2017-12-21
Support k8s node taints when creating pods


jithin1987:
## Context

k8s cluster admins can set custom taints to the nodes to support specific cluster requirements.
Eg: Dedicated nodes for builds, customer powerful nodes for builds etc. This requires Screwdriver kubernetes executor to be able to express tolerations in the pod spec while creating it.

## Objective

1. Ability to express toleration while creating k8s pod

tkyi:
This feature has been verified in production.

PR #828, opened 2017-12-21
feat(827): add config for tolerations

jithin1987:
## Context

We intend to support k8s toleration when creating pods so that k8s admins can specify taints on k8s nodes where they want the builds to be scheduled 

## Objective

This adds config for toleration to be used in `executor-k8s-vm`

## References

#827

### DIFF
--- a/config/custom-environment-variables.yaml
+++ b/config/custom-environment-variables.yaml
@@ -85,7 +85,9 @@ executor:
                     high: K8S_MEMORY_HIGH
             # k8s tolerations for approprate pod scheduling.
             # Value is Array of format { key: 'key', value: 'value', effect: 'NoSchedule'}
-            tolerations: K8S_POD_TOLERATIONS
+            tolerations:
+              __name: K8S_POD_TOLERATIONS
+              __format: json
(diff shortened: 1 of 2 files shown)


config/custom-environment-variables.yaml
```diff
@@ -83,6 +83,11 @@ executor:
                 memory:
                     low: K8S_MEMORY_LOW
                     high: K8S_MEMORY_HIGH
+            # k8s tolerations for approprate pod scheduling.
+            # Value is Array of format { key: 'key', value: 'value', effect: 'NoSchedule'}
+            tolerations:
+              __name: K8S_POD_TOLERATIONS
```
d2lam: `K8S_TOLERATIONS` to be consistent since `K8S_CPU_LOW/HIGH` are also config for pods.

jithin1987: `K8S_CPU_LOW/HIGH` was not POD configs, they were args for `hyper-runner`

d2lam: ok 👍  I'm fine with it

PR #833, opened 2018-01-04
fix: add config for sequalize connection pool

jithin1987:
## Objective 

This provides a placeholder for setting sequalize connection pool config See http://docs.sequelizejs.com/class/lib/sequelize.js~Sequelize.html#instance-constructor-constructor for options and default values


chestery:
👍 lgtm

PR #840, opened 2018-01-12
fix: use binary.file name for exec

yuichi10:
## Context
It will be better to download binary command as name of binary.file rather than name of version. 

It makes easy to understand what users are executing. 
Also some people want to use the command name which is executing in their shell or binary. So I think it is better to set binary.file for binary command name.

## Objective
Change this way to download binary command
`/opt/sd/commands/foo/bar/1.0.1` 
to 
`/opt/sd/commands/foo/bar/1.0.1/<binary.file>`


design/commands.md
```diff
@@ -103,9 +103,9 @@ $ sd_cmd exec foo/bar@1 -baz sample
  - **binary** - A script or binary that can be downloaded and directly executed.
 
 ```bash
-$ curl -o /opt/sd/commands/foo/bar/1.0.1 https://store/v1/commands/foo/bar/1.0.1
-$ chmod +x /opt/sd/commands/foo/bar/1.0.1
-$ /opt/sd/commands/foo/bar/1.0.1 -baz sample
+$ curl -o /opt/sd/commands/foo/bar/1.0.1 https://store/v1/commands/foo/bar/1.0.1/foobar.sh
```
catto: The path for store API wouldn't be changed. On the other hand, you need to modify the argument for `-o` switch.

yuichi10: You're right I will fix.

design/commands.md
```diff
@@ -103,9 +103,9 @@ $ sd_cmd exec foo/bar@1 -baz sample
  - **binary** - A script or binary that can be downloaded and directly executed.
 
 ```bash
-$ curl -o /opt/sd/commands/foo/bar/1.0.1 https://store/v1/commands/foo/bar/1.0.1
-$ chmod +x /opt/sd/commands/foo/bar/1.0.1
-$ /opt/sd/commands/foo/bar/1.0.1 -baz sample
+$ curl -o /opt/sd/commands/foo/bar/1.0.1/foobar.sh https://store/v1/commands/foo/bar/1.0.1
```
minz1027: This way, are we suggesting that there can be multiple `.sh` files under `/foo/bar/`?

yuichi10: I think there will be no multiple `.sh` files under `/foo/bar`, because the `.sh` files will be under `/foo/bar/<version>`. Also the `.sh` file and version is one-to-one relationship, so there will be one `.sh` file under `/foo/bar/<version>`.

minz1027: @yuichi10 Cool, thanks for the explanation

Issue #233, opened 2016-10-03
Pagination


d2lam:
We temporarily removed pagination due to DynamoDB limitation [#24](https://github.com/screwdriver-cd/datastore-dynamodb/pull/24). We should think of another way to implement pagination and deal with DynamoDB.

stjohnjohnson:
I was thinking we could use something like `?from=5bf1fd927dfb8679496a2e6cf00cbe50c1c87145&count=10`

jithin1987:
This is not valid anymore since we don't use DynamoDB

PR #843, opened 2018-01-16
feat(842): Add route for getting template tags

ScottyFillups:
## Context

It's hard for users to figure out what tags are available for each template and what versions they correspond to. It would be nice to have an endpoint where we could list the available tags for a specific template name.

## Objective

Create an endpoint, GET `/templates/{name}/tags`, which returns template tags.

### DIFF
--- a/plugins/templates/listTags.js
+++ b/plugins/templates/listTags.js
@@ -25,7 +25,7 @@ module.exports = () => ({
                 sort: request.query.sort
             }).then((templates) => {
                 if (templates.length === 0) {
-                    throw boom.notFound('Template does not exist');
+                    throw boom.notFound('No tags found for template');
                 }
 
                 reply(templates.map(p => p.toJson()));
--- a/plugins/templates/listTags.js
+++ b/plugins/templates/listTags.js
@@ -11,7 +11,7 @@ module.exports = () => ({
     path: '/templates/{name}/tags',
     config: {
         description: 'Get all template tags for a given template name with pagination',
-        notes: 'Returns all template records for a given template name',
+        notes: 'Returns all template tags for a given template name',
         tags: ['api', 'templates', 'tags'],
         handler: (request, reply) => {
             const factory = request.server.app.templateTagFactory;
--- a/plugins/templates/listTags.js
+++ b/plugins/templates/listTags.js
@@ -23,12 +23,12 @@ module.exports = () => ({
                     count: request.query.count
                 },
                 sort: request.query.sort
-            }).then((templates) => {
-                if (templates.length === 0) {
+            }).then((tags) => {
+                if (tags.length === 0) {
                     throw boom.notFound('No tags found for template');
                 }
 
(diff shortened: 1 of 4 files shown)


plugins/templates/listTags.js
```diff
@@ -0,0 +1,45 @@
...
+const joi = require('joi');
+const schema = require('screwdriver-data-schema');
+const listSchema = joi.array().items(schema.models.templateTag.base).label('List of templates');
+const nameSchema = joi.reach(schema.models.templateTag.base, 'name');
+
+module.exports = () => ({
+    method: 'GET',
+    path: '/templates/{name}/tags',
+    config: {
+        description: 'Get all template tags for a given template name with pagination',
```
d2lam: We actually don't have pagination right now: https://github.com/screwdriver-cd/screwdriver/issues/233

plugins/templates/listTags.js
```diff
@@ -0,0 +1,45 @@
...
+        notes: 'Returns all template records for a given template name',
```
d2lam: This one returns the template tags only, not the whole template records right?

test/plugins/data/templateTags.json
```diff
@@ -0,0 +1,18 @@
+[{
+    "id": 7969,
+    "name": "screwdriver/build",
+    "tag": "stable",
+    "version": "0.0.1"
+},
+{
+    "id": 7970,
+    "name": "screwdriver/build",
+    "tag": "stable",
```
d2lam: Hmm to be accurate, 1 tag can only be mapped to 1 version. There cannot be multiple versions with the `stable` tag. For example, `stable` can only map to `3.2.1`. 
However, 1 version can have multiple tags. For example, `3.2.1` can be `stable` and `latest` at the same time. It's very similar to docker hub tags.

plugins/templates/listTags.js
```diff
@@ -0,0 +1,45 @@
...
+            return factory.list({
+                params: { name: request.params.name },
+                paginate: {
+                    page: request.query.page,
+                    count: request.query.count
+                },
+                sort: request.query.sort
+            }).then((templates) => {
+                if (templates.length === 0) {
+                    throw boom.notFound('Template does not exist');
```
minz1027: It's not templates, it's template tags. A template can have zero tags.

plugins/templates/listTags.js
d2lam: @ScottyFillups  We should remove the `with pagination`

plugins/templates/listTags.js
```diff
@@ -0,0 +1,45 @@
...
+            const factory = request.server.app.templateTagFactory;
+
+            return factory.list({
+                params: { name: request.params.name },
+                paginate: {
+                    page: request.query.page,
+                    count: request.query.count
+                },
+                sort: request.query.sort
+            }).then((templates) => {
```
minz1027: I will prefer we use `tags` instead of `templates` everywhere

PR #848, opened 2018-01-19
fix: bump the wait time

minz1027:
currently it's getting the status for the old PR builds and that's failing the function tests. Wait longer to get the right status

### DIFF
--- a/features/step_definitions/git-flow.js
+++ b/features/step_definitions/git-flow.js
@@ -121,7 +121,7 @@ defineSupportCode(({ Before, Given, When, Then }) => {
     When(/^new changes are pushed to that pull request$/, {
         timeout: TIMEOUT
     }, function step() {
-        return this.promiseToWait(15) // Find & save the previous build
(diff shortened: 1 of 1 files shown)


features/step_definitions/git-flow.js
```diff
@@ -121,7 +121,7 @@ defineSupportCode(({ Before, Given, When, Then }) => {
     When(/^new changes are pushed to that pull request$/, {
         timeout: TIMEOUT
     }, function step() {
-        return this.promiseToWait(3) // Find & save the previous build
+        return this.promiseToWait(10) // Find & save the previous build
```
petey: This is really just kicking the can further down the road. It isn't solving the underlying problem. A better solution would be to have an exponential-backoff retry to get the right data.

minz1027: Agree, should record the `buildId` for the previous build and make sure the new one has a different `buildId`. I will create a issue for this. This PR is to unblock the pipeline for now.

PR #872, opened 2018-02-02
feat: Allow users to specify a metric host in config files

ScottyFillups:
## Context

It would be convenient if users could specify a URL to send executor data to their metric services.

## Objective

Add an example of `metricHost` usage in `default.yaml` and read an environment variable, `ECOSYSTEM_METRIC_HOST`, in `custom-environment-variables.yaml`.

## References

Merge for executor-k8s-vm `metricHost` implementation: https://github.com/screwdriver-cd/executor-k8s-vm/pull/23

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -232,7 +232,7 @@ ecosystem:
     badges: https://img.shields.io/badge/build-{{status}}-{{color}}.svg
     # Default registry to pull build containers from. Uses Docker Hub if nothing/empty string is provided
     dockerRegistry: ""
-    # Hostname for metrics service, or pushgateway
-    metricHost: http://pushgateway.example.org:9091
+    # Pushgateway URL for Prometheus
+    pushgatewayUrl: http://pushgateway.example.org:9091
     # Extra origins allowed to do CORS to API
     allowCors: []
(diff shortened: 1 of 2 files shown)


config/default.yaml
```diff
@@ -232,5 +232,7 @@ ecosystem:
     badges: https://img.shields.io/badge/build-{{status}}-{{color}}.svg
     # Default registry to pull build containers from. Uses Docker Hub if nothing/empty string is provided
     dockerRegistry: ""
+    # Hostname for metrics service, or pushgateway
+    metric_host: http://pushgateway.example.org:9091
```
petey: Should this be camelCase instead of snake_case?

ScottyFillups: Good point, I'll fix that.

config/default.yaml
```diff
@@ -232,5 +232,7 @@ ecosystem:
     badges: https://img.shields.io/badge/build-{{status}}-{{color}}.svg
     # Default registry to pull build containers from. Uses Docker Hub if nothing/empty string is provided
     dockerRegistry: ""
+    # Pushgateway URL for Prometheus
+    pushgatewayUrl: http://pushgateway.example.org:9091
```
minz1027: Can we comment this out? If we don't set the ENV, the executor will try to use this.

ScottyFillups:
Changes moved to https://github.com/screwdriver-cd/queue-worker/pull/43

PR #882, opened 2018-02-13
feat: add jwtEnv

d2lam:
Generate an additional field `jwtEnvironment` to indicates what environment this jwt should be generated for. By default it is `''`. We can set this to `test` or `beta` for our beta environment.

### DIFF
--- a/config/custom-environment-variables.yaml
+++ b/config/custom-environment-variables.yaml
@@ -10,6 +10,8 @@ auth:
     # Generate one by running
     # $ openssl rsa -in jwt.pem -pubout -out jwt.pub
     jwtPublicKey: SECRET_JWT_PUBLIC_KEY
+    # Environment of the JWTs. For example: 'prod' or 'beta'
+    jwtEnvironment: JWT_ENVIRONMENT
     # A password used for encrypting session data.
     # **Needs to be minimum 32 characters**
     cookiePassword: SECRET_COOKIE_PASSWORD
--- a/test/plugins/auth.test.js
+++ b/test/plugins/auth.test.js
@@ -195,6 +195,33 @@ describe('auth plugin test', () => {
             }, next);
         });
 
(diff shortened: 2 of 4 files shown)


test/plugins/auth.test.js
```diff
@@ -511,7 +540,37 @@ describe('auth plugin test', () => {
...
+                assert.equal(reply.statusCode, 200, 'Login route should be available');
+                assert.ok(reply.result.token, 'Token not returned');
+                expect(reply.result.token).to.be.a.jwt
+                    .and.have.property('username', 'robin');
+                expect(reply.result.token).to.be.a.jwt
+                    .and.have.property('scope')
+                    .with.lengthOf(1);
+                expect(reply.result.token).to.be.a.jwt
+                    .and.deep.property('scope[0]', 'user');
+            })
```
tkyi: Can you add this assertion here:
```
 expect(reply.result.token).to.be.a.jwt
                    .and.have.property('environment', '');
```

config/custom-environment-variables.yaml
```diff
@@ -10,6 +10,8 @@ auth:
     # Generate one by running
     # $ openssl rsa -in jwt.pem -pubout -out jwt.pub
     jwtPublicKey: SECRET_JWT_PUBLIC_KEY
+    # Environment of the JWTs. For example: 'prod' or 'beta'
+    jwtEnvironment: JWT_ENVIRONMENT
```
jithin1987: is this value the same as `PREFIX` we have for a bunch of settings ?

d2lam: No, it can be anything (`test`, `beta`, `prod`) depending on the cluster manager.

jithin1987: makes sense, I was just checking if we already have any duplicate settings which captures this value already.

Issue #677, opened 2017-07-31
Shared Step Improvement


stjohnjohnson:
List of things we should probably do:

 - [x] Add Readme to https://github.com/screwdriver-cd/sd-step
 - [x] Change CLI interface per https://github.com/screwdriver-cd/screwdriver/issues/650
 - [x] Add design doc to https://github.com/screwdriver-cd/screwdriver
 - [x] Check for root, if not, sudo during install per https://github.com/screwdriver-cd/screwdriver/issues/666
 - [x] Capture logs for debugging if needed instead of sending to `/dev/null`: https://github.com/screwdriver-cd/sd-step/blob/b9cc1857f5d7a39626860193e3da7dddfab1fb99/sd-step.go#L79
 - [ ] Provide option for using alternative sources for hab packages (like artifactory)

yuichi10:
### Summary 16/01/2018
I create sd-cmd which can exec binary commands. But there are no API for sd-cmd now, so I just try sd-cmd by [mock](https://github.com/yuichi10/mock_sd-cmd). 
In this project I do not use `urfave/cli` library because we received many kind of arguments or flags.

I think I can crate almost all function for execute binary but I have one problem for logging.
The problem summary is that I can't output log to file and terminal in right order in real time.
It will be a long sentence, though I will write the detail of this problem to 677-sd-cmd room on Slack.

drewfish:
Some ideas/questions about the [commands design](https://github.com/screwdriver-cd/screwdriver/blob/master/design/commands.md):
* does it make sense to add a link to "habitat"?  I have no idea what that is
* can the same command be called multiple times within the same build? does each invocation get a separate log file?
* presumably there'll be shared steps to provision/release selenium. which team will author those?
* the `$SD_ARTIFACTS_DIR/.sd/commands/namespace/name/version/timestamp.log` directory structure seems overly long
    * something like `$SD_ARTIFACTS_DIR/.sd/commands/{timestamp}-{namespace}-{name}.log` would be nice. would be nice if this log also contains the fully-resolved version of the command as well as it's full invocation (all arguments)

stjohnjohnson:
@drewfish 

 - Good point, sorry. https://www.habitat.sh/
 - Can be called unlimited times.
 - Right now there is no name space restrictions, so it can be anyone.  The thing we haven't thought of yet is how to bundle commands or even templates with SD.
 - I like it, good idea.

[...]

PR #892, opened 2018-02-20
feat(677): Add listVersion command endpoint

kumada626:
## Context

Like `template`, we need an endpoint that can get all versions of commands to show all versions of commands in UI.

## Objective

Like `GET /templates/{name}` endpoint, I added an endpoint that gets all versions of commands.

## References

Related #677

### DIFF
--- a/test/plugins/commands.test.js
+++ b/test/plugins/commands.test.js
@@ -8,6 +8,7 @@ const urlLib = require('url');
 const hoek = require('hoek');
 const testcommand = require('./data/command.json');
 const testcommands = require('./data/commands.json');
+const testcommandversions = require('./data/commandVersions.json');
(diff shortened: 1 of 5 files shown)


test/plugins/commands.test.js
```diff
@@ -8,6 +8,7 @@ const urlLib = require('url');
 const hoek = require('hoek');
 const testcommand = require('./data/command.json');
 const testcommands = require('./data/commands.json');
+const testcommandversions = require('./data/commandVersions.json');
```
tkyi: Non-blocking comment, but can we convert these variables to camel case?

kumada626: Converted variables to camel case.

PR #893, opened 2018-02-20
fix: use node:8 as image

minz1027:
## Context
We are having security vulnerabilities in some of our packages. But some of the latest pkgs require node:8.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:6
+FROM node:8
 
 # Screwdriver Version
 ARG VERSION=latest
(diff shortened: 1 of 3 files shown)


stjohnjohnson:
Please update the Dockerfiles as well: https://github.com/screwdriver-cd/screwdriver/blob/cbfcd17d61fa345e64c291825165350c7185f7c8/Dockerfile#L1 and https://github.com/screwdriver-cd/screwdriver/blob/cbfcd17d61fa345e64c291825165350c7185f7c8/Dockerfile.local#L1

Issue #806, opened 2017-11-22
Start build for any job in pipeline


jithin1987:
## Context
As a user I want to start my pipeline at a specific job.

## Objective

As of today auto trigger for a pipeline's event is either `commit` or `pr` or an external pipeline's job. Also user can manually start a pipeline which behaves similar to `commit` triggered from last commit on configured SCM repo's branch.

Expose in UI a way for user to start pipeline from a specific job.

**Tasks**

- [x] UI pipeline graph for an event should expose a `Start Build` button against a job in the pipeline.

- [x] On clicking `Start Build` button, Screwdriver should create a new event which should inherit all properties from the event from which user clicked `Start Build`. 

- [x] UX should be clear about what happens when user clicks `Start Build` button. User should be aware of the following Execution Flow

Execution Flow.
1. User is viewing pipeline graph for event `T1` at commit SHA `SHA100`. 
1. From the loaded pipeline graph for event `T1` at commit SHA `SHA100` user selects job `X` and clicks on `Start Build` for job `X`. 
1. Screwdriver should create a new event `T2` at commit SHA `SHA100`.  Pipeline for event `T2` should start at job `X` and should have access to all `meta` which  event `T1` had access to.

minz1027:
@jithin1987 Do we want to allow user to just pick any SHA to start from or just the most recent one?

jithin1987:
@minz1027 
1. It will be the SHA of the event user started the build from.
2. If the event was from an external pipeline then it should behave like this feature requirement https://github.com/screwdriver-cd/screwdriver/issues/809

minz1027:
### Update 02/13

## Backend
I think for backend, we already have a function that can achieve most requirements. It can preserve meta from external trigger and previous build:
https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/builds/index.js#L81
Just need to tune it a bit to take in SHA from config and add the logic to /POST event endpoint.

The only case not covered will be `Join`. But we don't even support fully preserve meta for `Join`.https://github.com/screwdriver-cd/screwdriver/issues/883

## UI
After the backend change, when users want to start a build in middle, the UI need to pass in 
- parentBulidId
- pipelineId
- startFrom
- SHA
- CauseMessage(optional)

petey:
UI 
======
[...]

PR #905, opened 2018-02-26
fix: pass info from parentEvent when create new event

minz1027:
## Context
We want to support user to restart from any event. To allow that, we need to pass in `sha` and `workflowGraph` from the `parentEvent` so that model can use those info to recreate the event.

## References
Related to https://github.com/screwdriver-cd/screwdriver/issues/806
Blcoked by: https://github.com/screwdriver-cd/models/pull/224


FenrirUnbound:
`package.json` should be updated to reflect the minimum `screwdriver-cd/models` version

PR #906, opened 2018-02-28
feat(677): Command create endpoint receives multipa…

kumada626:
…rt/form-data

## Context

When register the `command`, the sd-cmd cli will POST the command meta and the binary file to the API. Then, the API POST the binary file to the store and register the meta to the DB.
I have enabled the API to receive multipart/form-data to receive both command meta and binary.

## Objective

The API is able to receive multipart/form-data. If the API receives multipart/form-data, it POST the binary file to the store and register the meta to the DB.
If the API receives not multipart/form-data, it only register the meta to the DB.

## References
Below PR shows overview of publishing sd-command in description.
https://github.com/screwdriver-cd/sd-cmd/pull/13


plugins/commands/create.js
```diff
@@ -1,10 +1,72 @@
...
 const hoek = require('hoek');
 const urlLib = require('url');
+const req = require('request');
+const VERSION_REGEX = schema.config.regex.VERSION;
+
+/**
+ * Publish binary command
+ * @method binaryCommandPublish
+ * @param  {commandFactory} commandFactory      commandFactory
+ * @param  {config}         config              Command config
```
tkyi: Should be `Object`

plugins/commands/create.js
```diff
@@ -1,10 +1,72 @@
...
+ * @param  {binary}         binary              Binary published to the store
+ * @param  {string}         storeUrl            URL to the store
```
tkyi: Can you capitalize these (aka `String`, `Binary`, etc)?

plugins/commands/create.js
kumada626: fixed.

plugins/commands/create.js
kumada626: Capitalized these.

plugins/commands/create.js
```diff
@@ -22,58 +84,89 @@ module.exports = () => ({
...
         },
-        handler: (request, reply) => validator(request.payload.yaml)
-            .then((config) => {
-                if (config.errors.length > 0) {
-                    throw boom.badRequest(
-                        `Command has invalid format: ${config.errors.length} error(s).`,
-                        config.errors);
+        payload: {
+            parse: true,
+            allow: ['multipart/form-data', 'application/json']
```
yoshwata: I'd like you to set `payload.maxBytes` here. I think the default 1MB setting is too small.
https://hapijs.com/api/3.0.0#-routeoptionspayloadmaxbytes

plugins/commands/create.js
```diff
@@ -22,58 +84,89 @@ module.exports = () => ({
...
+            let commandBin;
+
+            // if Content-type is multipart/form-data, both command binary and meta are posted
+            if (request.headers['content-type'].startsWith('multipart/form-data')) {
+                if (data.file === undefined || data.file.length !== 2) {
+                    return reply(boom.badRequest('Posted invalid number of files.'));
                 }
+                data.file.forEach((file) => {
+                    if (typeof file === 'string') {
+                        commandSpec = file;
```
yoshwata: It also might be needed to parse `yaml` key here like you've done on non-multipart.
`commandSpec = data.yaml;`

plugins/commands/create.js
kumada626: I added `payload.maxBytes` to 1GB.

plugins/commands/create.js
yoshwata: This is not need to fix. I'll try the implementation of that sd-cmd sends multipart with no `yaml` key, but sends application/json sends with a `yaml` key.

plugins/commands/create.js
```diff
@@ -97,16 +97,11 @@ module.exports = () => ({
 
             // if Content-type is multipart/form-data, both command binary and meta are posted
             if (request.headers['content-type'].startsWith('multipart/form-data')) {
-                if (data.file === undefined || data.file.length !== 2) {
+                if (data.binary === undefined || data.spec === undefined) {
```
kumada626: According to the sd-cmd PR comment (https://github.com/screwdriver-cd/sd-cmd/pull/13#issuecomment-371022838), sd-cmd cli POST a multipart/form-data which includes spec part and binary part. Therefore, I fixed commands API depending on this change.

plugins/commands/create.js
```diff
@@ -97,16 +97,11 @@ module.exports = () => ({
...
                     return reply(boom.badRequest('Posted invalid number of files.'));
```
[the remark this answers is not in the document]
kumada626: Fixed message.

plugins/commands/create.js
```diff
@@ -1,10 +1,73 @@
...
+ */
+function binaryCommandPublish(commandFactory, config, binary, storeUrl, authToken) {
+    const [, major, minor] = VERSION_REGEX.exec(config.version);
+    const searchVersion = minor ? `${major}${minor}` : major;
+    let publishVersion;
+
+    return commandFactory.getCommand(`${config.namespace}/${config.name}@${searchVersion}`)
+        .then((latest) => {
+            if (!latest) {
+                publishVersion = minor ? `${major}${minor}.0` : `${major}.0.0`;
```
d2lam: I'm wondering why we need to do the duplicate work of this https://github.com/screwdriver-cd/models/blob/master/lib/commandFactory.js#L66-L78 here?

Can we call `return commandFactory.create(config)` (line 68) first, and that will return a command, and we can get `command.version`?

kumada626: Actually, we call `return commandFactory.create(config)` (line 68) first, and that will return a command, and we can get `command.version`. However, this will cause the meta to be saved and the binary files not to be saved if an error occurs while saving to the store.

d2lam: Okay. So you want the command to be created **ONLY AFTER** the binary is successfully saved to the store. That makes sense.

Issue #904, opened 2018-02-26
Meta should exist at event level


jithin1987:
## Context

Today a build's `meta` information is copied over from the parent build (triggering build). This creates a problem when we re-start a failed build or launch an independent job from an existing event. https://github.com/screwdriver-cd/screwdriver/issues/806 Having meta stored at event level, instead of a build level will simplify use case of accessing existing meta from the previous event.

## Objective

1. Store `meta` information at an Event level.
1. With in builds for jobs triggered as part of same event read `meta` from Event instead of copying from parent build.
1. When starting an independent job (https://github.com/screwdriver-cd/screwdriver/issues/806) read initial meta for the new event from the old event from which the job was started.

d2lam:
## **Update 3/7/18**
### After discussing, we have decided:
- Case 1: build does not have a parent build, use the event's metadata.
In this example, restarting the detached job to rollback a deployment will use the event's metadata.
![image uploaded from ios](https://user-images.githubusercontent.com/3401924/37127754-fe37aa92-222b-11e8-8758-1cdad505d13a.jpg)

- Case 2: build has 1 parent build, use the parent build's metadata
In this example, if the user restarts job C, job C will be rerun using job A's metadata.
![image uploaded from ios 3](https://user-images.githubusercontent.com/3401924/37127765-123036a4-222c-11e8-80e2-8e9569b9e71a.jpg)

- Case 3: build has multiple parent builds, use the merged metadata from all the parents
In this example , if the user restarts job F, it will be run using the combined metadata from job A & B. 
![image uploaded from ios 4](https://user-images.githubusercontent.com/3401924/37128422-f727df9e-222e-11e8-9d74-a732ac496242.jpg)


### Reasons why we do this:
If we just use event's metadata for all cases, it will be very hard for users to consume templates. 

For example, in case 2: 
- There are 2 templates: `template_test` that does `meta set image`, and `template_publish` that does `meta get image` inside the code and does something with it. 
- job A and job B both use `template_test`
- job C and job D both use `template_publish`
- If we always use event's metadata, then it's really inconvenient. Users will need to add a poststep for jobA to set `image` to `image_A` and for jobB to set `image` to `image_B`. Then inside jobC and jobD, users will need to set `image_A` to `image` and `image_B` to `image`. 

The idea is to minimize extra overhead on the user's part to consume templates.

Another example, in case 3:
[...]

PR #915, opened 2018-03-08
feat(904): support event meta

d2lam:
This PR updates the event meta data when updating a build. 
The new meta should be merged into the event meta. If they have the same key, the new meta will override the old one. 

Blocked by: https://github.com/screwdriver-cd/data-schema/pull/209

### DIFF
--- a/plugins/builds/update.js
+++ b/plugins/builds/update.js
@@ -19,6 +19,7 @@ module.exports = () => ({
         },
         handler: (request, reply) => {
             const buildFactory = request.server.app.buildFactory;
+            const eventFactory = request.server.app.eventFactory;
             const id = request.params.id;
             const desiredStatus = request.payload.status;
             const statusMessage = request.payload.statusMessage;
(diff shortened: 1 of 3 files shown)


plugins/builds/update.js
```diff
@@ -53,20 +54,23 @@ module.exports = () => ({
...
+                    }
+
+                    return eventFactory.get(build.eventId).then(event => ({ build, event }));
+                }).then(({ build, event }) => {
+                    switch (desiredStatus) {
+                    case 'SUCCESS':
+                    case 'FAILURE':
+                    case 'ABORTED':
+                        build.meta = request.payload.meta || {};
+                        event.meta = { ...event.meta, ...build.meta };
```
tkyi: Shouldn't the event meta only be updated on `SUCCESS` and `ABORTED`? Is there a use case for updating event meta when the build is `FAILURE`?

tkyi: Nevermind, after discussing with you I think it's more consistent/intuitive to update the event metadata where we update the build metadata.

plugins/builds/update.js
```diff
@@ -53,20 +54,23 @@ module.exports = () => ({
...
-                            break;
-                        case 'RUNNING':
-                            build.startTime = (new Date()).toISOString();
-                            break;
-                        default:
-                            throw boom.badRequest(`Cannot update builds to ${desiredStatus}`);
-                        }
+                    }
+
+                    return eventFactory.get(build.eventId).then(event => ({ build, event }));
```
tkyi: I think `.then(event => ({ build, event }))` can be skipped, but it's more readable if we keep it the way you have it.

Issue #870, opened 2018-02-01
Allow public access to build page and artifacts


jithin1987:
## Context

https://github.com/screwdriver-cd/screwdriver/issues/858 introduced authentication for accessing store data. While this is important in enterprise applications this is not desirable in public instances of SD. 

## Objective

1. Allow SD cluster admin to configure public access to build page and artifacts.

stjohnjohnson:
# 3/16

Guest mode is now available.  There are still two bugs:
 - When not logged in, everything shows as blank pages (due to 403 errors), **need UI redirects.**
 - When logged in as guest and trying to perform write operations, user is presented with a strange error message "Missing scope".  **Need better UI message.**

jithin1987:
> When not logged in, everything shows as blank pages (due to 403 errors), need UI redirects.

Will be fixing this on a priority.

tkyi:
These changes are working in prod 👍

Issue #572, opened 2017-05-31
Enforce Login


stjohnjohnson:
We've noticed that some teams would like to require that only authenticated users can view the pipelines.  Thus preventing leaking of information to those without authentication.

We should have a feature to require all calls to be authenticated.

d2lam:
This is a little confusing to me. Wouldn't this be accomplished by https://github.com/screwdriver-cd/screwdriver/issues/466? 

If read protection is implemented, then only people who have read permission to the repo can view the pipeline. What is the difference between these 2 issues?

stjohnjohnson:
@d2lam, would #466 require logins?  Would there still be a logged-out experience?  If not, then this would be resolved by that one.

If there is still a logged-out experience, where I can visit things like https://cd.screwdriver.cd/pipelines/7.  Then it should be an option to require login (for companies with GHE as an example).

d2lam:
I see. I didn't think that we would keep the logged-out experience. I'm not sure if I would care about some random public pipelines if I'm a user.

stjohnjohnson:
@d2lam then that will affect how we design the UI.  There's no point to a login button since all access will send me there.

d2lam:
Yup. Just my personal opinion. We can always discuss as a team & keep this ticket opened.

stjohnjohnson:
Closing in favor of https://github.com/screwdriver-cd/screwdriver/issues/870

PR #919, opened 2018-03-12
feat(870): Adding Guest Mode

stjohnjohnson:
## Context

We're in a mixed mode right now, where anonymous browsing is allowed but logs and artifacts are not.  As per [this issue](https://github.com/screwdriver-cd/screwdriver/issues/870) our goal should be to allow server admins to chose wether guests can browse the site or not.  For example, cd.screwdriver.cd should allow guests to view the logs while internal deployments may not.

## Objective

This adds a configuration that will enable users to browse your pipelines, jobs, and builds without authenticating against a SCM.  It forces read-only access.

As a side effect, all routes (minus health check, stats, webhooks, and badges) now require token-based auth.  It also includes specific references as to what a build can access during a build.

## Breaking changes

 - Default routes of `/auth/login/` are no longer redirected to the first SCM.
 - Endpoints other than `/auth/token` no longer accept session-based auth.
 - Most endpoints now require authentication (except status, stats, webhooks, and badges)
 - Builds have explicit access to 19 routes.
 - Guests are forbidden from 26 routes.

## References

 - https://github.com/screwdriver-cd/screwdriver/issues/870
 - https://github.com/screwdriver-cd/screwdriver/issues/572

### DIFF
--- a/config/custom-environment-variables.yaml
+++ b/config/custom-environment-variables.yaml
@@ -21,6 +21,8 @@ auth:
     # A flag to set if the server is running over https.
     # Used as a flag for the OAuth flow
     https: IS_HTTPS
+    # A flag to set if you want guests to browse your pipelines
+    allowGuestAccess: AUTH_GUEST_ACCESS
     whitelist:
         __name: SECRET_WHITELIST
         __format: json
--- a/plugins/auth/contexts.js
+++ b/plugins/auth/contexts.js
@@ -1,19 +1,19 @@
 'use strict';
 
 /**
- * Get all scm contexts
+ * Get all auth contexts
  * @method login
- * @param  {Object}      config           Configuration from the user
- * @param  {Array}       config.whitelist List of allowed users to the API
- * @return {Object}                       Hapi Plugin Route
+ * @param  {Object}      config                  Configuration from the user
+ * @param  {Boolean}     config.allowGuestAccess Letting users browse your system
+ * @return {Object}                              Hapi Plugin Route
  */
-module.exports = () => ({
+module.exports = config => ({
     method: ['GET'],
     path: '/auth/contexts',
     config: {
-        description: 'Get all scm contexts',
-        notes: 'Get all scm contexts',
-        tags: ['api', 'scmContext'],
+        description: 'Get all auth contexts',
+        notes: 'Get all auth contexts',
+        tags: ['api', 'auth', 'context'],
         handler: (request, reply) => {
             const scm = request.server.app.userFactory.scm;
             const scmContexts = scm.getScmContexts();
@@ -27,6 +27,13 @@ module.exports = () => ({
 
                 contexts.push(context);
             });
+
+            if (config.allowGuestAccess) {
+                contexts.push({
+                    context: 'guest',
+                    displayName: 'Guest Access'
+                });
+            }
 
             return reply(contexts);
         }
--- a/plugins/auth/login.js
+++ b/plugins/auth/login.js
@@ -1,97 +1,145 @@
 'use strict';
 
 const boom = require('boom');
-const urlLib = require('url');
+const uuid = require('uuid/v4');
+
+/**
+ * Add a guest route for those who want to be in read-only mode
+ * @method addGuestRoute
+ * @param  {Hapi}     server                        Hapi Server
+ * @param  {Object}   config                        Configuration object
+ * @param  {Boolean}  config.allowGuestAccess       Letting users browse your system
+ * @return Array                                    List of new routes
+ */
+function addGuestRoute(server, config) {
+    return [{
+        method: ['GET'],
+        path: '/auth/login/guest/{web?}',
+        config: {
+            description: 'Login as an guest user',
+            notes: 'Authenticate an guest user',
+            tags: ['api', 'login'],
+            auth: null,
+            handler: (request, reply) => {
+                // Check if guest is allowed to login
+                if (!config.allowGuestAccess) {
+                    return reply(boom.forbidden('Guest users are not allowed access'));
+                }
+
+                const username = `guest/${uuid()}`;
+                const profile = request.server.plugins.auth.generateProfile(
+                    username, 'guest', ['user', 'guest'], {}
+                );
+
+                // Log that the user has authenticated
+                request.log(['auth'], `${username} has logged in`);
+
+                profile.token = request.server.plugins.auth.generateToken(profile);
+                request.cookieAuth.set(profile);
+
+                if (request.params.web === 'web') {
+                    return reply('<script>window.close();</script>');
+                }
+
+                return reply().redirect('/v4/auth/token');
+            }
+        }
+    }];
+}
+
+/**
+ * Add OAuth Routes for all SCM Contexts
+ * @method addOAuthRoutes
+ * @param  {Hapi}     server                 Hapi Server
+ * @param  {Object}   config                 Configuration object
+ * @param  {Array}  config.whitelist         List of users allowed into your system
+ * @return Array                             List of new routes
+ */
+function addOAuthRoutes(server, config) {
+    const scmContexts = server.root.app.userFactory.scm.getScmContexts();
+
+    return scmContexts.map(scmContext => ({
+        method: ['GET', 'POST'],
+        path: `/auth/login/${scmContext}/{web?}`,
+        config: {
+            description: 'Login using oauth',
+            notes: 'Authenticate user with oauth provider',
+            tags: ['api', 'login'],
+            auth: {
+                strategy: `oauth_${scmContext}`,
+                mode: 'try'
+            },
+            handler: (request, reply) => {
+                if (!request.auth.isAuthenticated) {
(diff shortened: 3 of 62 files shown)


config/custom-environment-variables.yaml
```diff
@@ -21,6 +21,8 @@ auth:
     # A flag to set if the server is running over https.
     # Used as a flag for the OAuth flow
     https: IS_HTTPS
+    # A flag to set if you want guests to browse your pipelines
+    allowGuestAccess: SECRET_GUEST_ACCESS
```
d2lam: maybe just `GUEST_ACCESS` since it's just a flag. Secret makes it sounds like a password.

plugins/auth/login.js
```diff
@@ -1,97 +1,145 @@
...
 
-                const location = urlLib.format({
-                    host: request.headers.host,
-                    port: request.headers.port,
-                    protocol: request.server.info.protocol,
-                    pathname: pathName
-                });
+                const username = `guest/${uuid()}`;
+                const profile = request.server.plugins.auth.generateProfile(
+                    username, 'guest', ['user', 'guest'], {}
```
d2lam: So guest user has 2 scopes: `user` and `guest`?

plugins/builds/create.js
```diff
@@ -12,8 +12,8 @@ module.exports = () => ({
         notes: 'Create and start a specific build',
         tags: ['api', 'builds'],
         auth: {
-            strategies: ['token', 'session'],
-            scope: ['user']
+            strategies: ['token'],
+            scope: ['user', '!guest']
```
d2lam: I have a question. From my understanding, this `scope` array is an OR. For example if `scope: ['user', 'build']` means either a build or a user is good to access.
In this case, don't we want an `AND`? It's a user AND it's not a guest (since guest users have both `user` and `guest` scope). Maybe I missed something. Could you explain?

config/custom-environment-variables.yaml
stjohnjohnson: That's fair, I was just following the previous standard in this section.  The original name of the group was "secret" prior to being renamed to "auth".  I would prefer `AUTH_GUEST_ACCESS` instead.

plugins/auth/login.js
stjohnjohnson: Yes, so a guest is a user, but is also a guest.

plugins/builds/create.js
stjohnjohnson: So for HapiJS, scope works like this:
 - If it starts with `+` that scope is required
 - If it starts with `!` that scope is not allowed
 - Otherwise it's an OR

More info [here](https://hapijs.com/api#-routeoptionsauthaccessscope).

config/custom-environment-variables.yaml
d2lam: sure. that works for me.

plugins/auth/login.js
```diff
@@ -1,97 +1,145 @@
...
-
-                    return reply().redirect('/v4/auth/token');
-                })
-                .catch(err => reply(boom.wrap(err)));
+/**
+ * Add OAuth Routes for all SCM Contexts
+ * @method addOAuthRoutes
+ * @param  {Hapi}     server                 Hapi Server
+ * @param  {Object}   config                 Configuration object
+ * @param  {Boolean}  config.whitelist       List of users allowed into your system
```
tkyi: Array

plugins/auth/contexts.js
```diff
@@ -1,19 +1,19 @@
...
 
 /**
- * Get all scm contexts
+ * Get all auth contexts
  * @method login
- * @param  {Object}      config           Configuration from the user
- * @param  {Array}       config.whitelist List of allowed users to the API
- * @return {Object}                       Hapi Plugin Route
+ * @param  {Object}      config                  Configuration from the user
+ * @param  {Array}       config.allowGuestAccess Letting users browse your system
```
tkyi: Boolean

plugins/builds/get.js
```diff
@@ -13,6 +13,15 @@ module.exports = () => ({
         description: 'Get a single build',
         notes: 'Returns a build record',
         tags: ['api', 'builds'],
+        auth: {
```
jithin1987: is there a way in `hapi` to push this to a higher middleware, adding to each and every old and newer routes is cumbersome.

Else we could defined these configs as constants in a separate file and import it here.
```
auth: USER_BUILD_TOKEN/ BUILD_TOKEN/USER_TOKEN
```

plugins/builds/artifacts/get.js
```diff
@@ -16,12 +16,12 @@ module.exports = config => ({
         notes: 'Redirects to store with proper token',
         tags: ['api', 'builds', 'artifacts'],
         auth: {
-            strategies: ['session', 'token'],
-            scope: ['user']
+            strategies: ['token'],
```
jithin1987: This will break current flow of opening artifacts as a link to API which in-turn redirects to store with a short lived token.

One option is to modify UI so that links are populated dynamically with token in URL or re-enable session back here.

Also one question, once a user scoped token comes does auth checks if user is actually logged in or does it ony check if token is not expired ?

plugins/builds/get.js
stjohnjohnson: Ehh, it's technically possible via [server.auth.default(options)](https://hapijs.com/api#-serverauthdefaultoptions).  But what would you really set it to?  Because our APIs have a really mixed permission:
 - Some allow only a user to access
 - Some allow only a build to access
 - Some allow both to access
 - Some don't allow guests to access
 - Some don't have any auth (status, stats, etc.)
 - Some have custom auth (webhook, oauth, etc.)

plugins/builds/artifacts/get.js
stjohnjohnson: This shouldn't.  Interacting with the links should be making an AJAX call then redirecting the user.

plugins/auth/contexts.js
stjohnjohnson: Thanks! Fixed!

plugins/auth/login.js
stjohnjohnson: Thanks! Fixed!

config/custom-environment-variables.yaml
stjohnjohnson: Fixed!

plugins/builds/artifacts/get.js
jithin1987: >  Interacting with the links should be making an AJAX call then redirecting the user.

This won't work on AJAX, UI can't redirect on AJAX and have browser provide a download dialogue, it needs JS. Simplicity of current approach is once user is logged in via session browser can handle redirection with no more JS involved after clicking the link.

If we must token, then one option is UI to download the entire file in AJAX  in memory and use JS (`URL.createObjectURL()`) to either change the current url or show file download dialog using an `a`  

Another option is for this end point to just return the store end point and use `window.open` to open a new window with the url. But that might run foul of popup settings.

I would prefer doing former approach while leaving `session` auth as it is for now and revisit this once UI has changed.

stjohnjohnson: I'll be happy to roll this back now, but we really should go with the second option.  However, instead of doing `window.open` why not `window.location`?  Since the URLs should be download only it won't _really_ cause the page to change location.

jithin1987: > I'll be happy to roll this back now, but we really should go with the second option. However, instead of doing window.open why not window.location? Since the URLs should be download only it won't really cause the page to change location.

Either with `window location/open` will need change in the current UI to not follow the redirect automatically but do a `window.location` from `Location` header send back in the redirect response. But that is a change in UI, so can you put `session` back just for this route and remove it once there is a PR in UI to do this.

stjohnjohnson: Updated.

tkyi:
Build failure:
```
11:32:57
  1) build plugin test "before each" hook for "registers the plugin":
11:32:57
     Error: Unknown authentication strategy session in /builds/{id}/artifacts/{name*}
```

plugins/builds/artifacts/get.js
```diff
@@ -17,11 +17,11 @@ module.exports = config => ({
         tags: ['api', 'builds', 'artifacts'],
         auth: {
             strategies: ['session', 'token'],
-            scope: ['user']
+            scope: ['user', 'build']
```
jithin1987: I think something went wrong with last commit, was supposed to add `session` back here based on discussion https://github.com/screwdriver-cd/screwdriver/pull/919#discussion_r174239529

stjohnjohnson: Not to scope, but to strategies (which it is there).

stjohnjohnson: Builds _should_ be able to access this endpoint, no?

jithin1987: aah sorry I got confused. Build does not call this `get` endpoint, so don't think it's required.

PR #921, opened 2018-03-14
Multiple parents

d2lam:
## Context
Allow a build to have multiple parents, so that for a `join`, a build will have access to all the parents' meta, instead of just the one that triggers it. 

## Objective
Allow builds to have multiple parents. There are a couple cases:
1. [A B] -> C. A passed -> create C
2. [A B] -> C. A passed -> C created; B passed -> update C to add B as parent build ID, then start
3. [A B] -> C. A passed -> C created; B failed -> delete C
4. [A B D] -> C. A passed -> C created; B failed -> delete C; D passed -> do nothing
5. [A B] -> C. A failed -> C not created; B failed -> do nothing

This boils down to: if there are any failed builds so far, then either delete (if the build exists), or do nothing. If all the builds are successful so far, then either create, or update to have an additional parentBuildId. 

## References
Blocked by: 
- https://github.com/screwdriver-cd/data-schema
- https://github.com/screwdriver-cd/models/pull/229
- https://github.com/screwdriver-cd/launcher/pull/157


plugins/builds/index.js
```diff
@@ -21,10 +21,13 @@ const workflowParser = require('screwdriver-workflow-parser');
  * @param  {String}   config.username       Username of build
  * @param  {String}   config.scmContext     SCM context
  * @param  {Build}    config.build          Build object
+ * @param  {Build}    config.start          Whether to start the build or not
```
tkyi: Boolean

plugins/builds/index.js
```diff
@@ -34,10 +37,11 @@ function startBuild(config) {
             return buildFactory.create({
                 jobId: job.id,
                 sha: build.sha,
-                parentBuildId: build.id,
+                parentBuildId: [build.id],
                 eventId: build.eventId,
                 username,
-                scmContext
+                scmContext,
+                start: start !== false
```
tkyi: should just be able to pass `start` since it's already a boolean

d2lam: `start` is not always passed in. So it could be `undefined`.

tkyi: oh okay 👍

plugins/builds/index.js
tkyi: add `[]` around it to indicate it's optional

plugins/builds/index.js
```diff
@@ -203,22 +223,55 @@ exports.register = (server, options, next) => {
...
                 // 2. ([~D,B,C]->A) currentJob=D, nextJob=A, joinList(A)=[B,C]
                 //    joinList doesn't include C, so start A
                 if (joinList.length === 0 || !joinListNames.includes(currentJobName)) {
-                    return startBuild(buildConfig);
+                    return createBuild(buildConfig);
                 }
 
-                // If no parent event id, start if all jobs in the list are done
                 if (!event.parentEventId) {
+                    let finishedBuilds = [];
```
minz1027: We also need this logic for restarting jobs from a past event. Should probably move logic down to after https://github.com/screwdriver-cd/screwdriver/pull/921/files#diff-7a514e440aa937beee722c7ac84dfb93R289

plugins/builds/index.js
```diff
@@ -59,6 +63,84 @@ function isJoinDone(joinList, finishedBuilds) {
...
+/**
+ * Handle next build logic: create, update, start, or remove
+ * @method handleNextBuild
+ * @param  {Object}   config                    configuration object
+ * @param  {Object}   config.buildConfig        config to create the build with
+ * @param  {Array}    config.joinList           list of job that join on this current job
+ * @param  {Array}    config.finishedBuilds     list of finished builds
+ * @param  {String}   config.jobName            jobname for this build
+ * @param  {Number}   config.parentBuildId      parent build Id of the new build
+ */
```
tkyi: Missing `@return`

plugins/builds/index.js
```diff
@@ -59,6 +63,83 @@ function isJoinDone(joinList, finishedBuilds) {
...
+            .map(b => b.id);
+
+            // Do a replace instead of push because of the restart join case
+        nextBuild.parentBuildId = successBuildsIds.concat(parentBuildId);
+
+        return nextBuild.update();
+    }).then((nextBuild) => {
+        const done = isJoinDone(joinList, finishedBuilds);
+
+        return done ? nextBuild.start() : null;
```
petey: can this be simply:
```
.then(nextBuild => isJoinDone(joinList, finishedBuilds) ? nextBuild.start() : null);
```

plugins/builds/index.js
```diff
@@ -24,11 +24,8 @@ const workflowParser = require('screwdriver-workflow-parser');
  * @param  {Boolean}  [config.start]        Whether to start the build or not
  * @return {Promise}
  */
-function createBuild(config) {
-    const {
-        jobFactory, buildFactory, pipelineId, jobName, username, scmContext, build, start
-    } = config;
-
+function createBuild({ jobFactory, buildFactory, pipelineId, jobName, username,
```
petey: Did you want to add `start=true` to this so it has a predefined value if not passed?

PR #938, opened 2018-03-20
Event create build

d2lam:
Join build now has access to all the metadata. However, restarting join doesn't work because in the UI, `parentBuildId` is not accepted as an array yet. 

Instead of changing the UI to accept `parentBuildId` as an array or int, we will now change eventFactory's create to take in only `buildId` and figure out the `parentBuildId` in the API instead of the UI. 

Blocked by: https://github.com/screwdriver-cd/ui/pull/266


plugins/events/create.js
```diff
@@ -24,93 +24,109 @@ module.exports = () => ({
...
-                                    });
-                            }
+                if (buildId) { // restart case
+                    return buildFactory.get(buildId)
+                        .then(b => jobFactory.get(b.jobId)
+                            .then(j => ({
+                                pipelineId: j.pipelineId,
+                                startFrom: j.name,
+                                parentBuildId: b.parentBuildId,
+                                parentEventId: b.parentEventId
```
tkyi: Should just be `b.eventId`

plugins/events/create.js
```diff
@@ -24,93 +24,109 @@ module.exports = () => ({
...
+                    .then((permissions) => {
+                        if (!permissions.push) {
+                            throw boom.unauthorized(`User ${username} `
+                              + 'does not have push permission for this repo');
+                        }
+                    })
+                    // User has good permissions, create an event
+                    .then(() => user.unsealToken())
+                    .then((token) => {
+                        // If there is parentEvent, pass workflow, worklfowGraph and sha to payload
```
tkyi: workflowGraph*

Issue #911, opened 2018-03-07
Control job trigger based on source path


jithin1987:
## Context

Today screwdriver listens to any changes in the SCM repo on `commit` and `pr` and runs builds for jobs which have `requires` on them. This is not ideal for SCM repositories following monorepo pattern, where individual sub directories need to be built by themselves. 

## Objective

A job configuration should support an annotation to specify paths within the SCM repo which should trigger it. 

```
jobs:
  buildPathA:
    requires: [~commit, ~pr]
    image: node:8
    sourcePaths: [src/A, src/AConfig]
    steps:
      - buildA: make build
```

Above configuration should run a build for job `buildPathA` on `commit` or `pr` if and only if the SCM commit change include files under path `src/A` or `src/AConfig`

This along with supporting `meta` at event level will allow SCMs following a monorepo source pattern to build multiple parallel pipeline workflows to build various parts of the repo.

## References 
Subdirectory Feature -  https://github.com/screwdriver-cd/screwdriver/issues/471
Meta at event level - https://github.com/screwdriver-cd/screwdriver/issues/904

## Updates
* 3/8/18 Moved config from annotations to settings 
* 3/9/18 Moved config to job config, settings was wrong place to put

jithin1987:
@stjohnjohnson  I plan to use this to support sub directories #471  Thoughts ?

stjohnjohnson:
@jithin1987 Seems like a good approach.  I'd love to see more things like this in the future.  Although we should probably implement a generic method in the models to access annotations easier (with defaults) from the pipeline and job level.

jithin1987:
## *Update 3/8/2018*

Rather than putting the source config in `annotations` we have decided to put the config under `settings` since this is a permanent setting. Update config for this feature will look like

```
jobs:
  buildPathA:
    requires: [~commit, ~pr]
    image: node:8
    settings:
      sourcePaths: [src/A, src/AConfig]
    steps:
      - buildA: make build
```

stjohnjohnson:
Why? Isn’t that the point of annotations? I thought those were also
permanent. Why do we even need settings vs annotations?

The only specificity of permanence is if it has beta.sd.cd vs sd.cd.

jithin1987:
@stjohnjohnson we had a discussion and what came out was features of experimental nature and not tied to core functionality should go into annotations, since it's a free flow config. More defined feature configuration should either be put in settings or in the main job config itself. 

[...]

PR #940, opened 2018-03-22
feat(911): support subdirectory [5]

d2lam:
## Context
To support Subdirectory, `scm` plugin `getChangedFiles` will return an array of files that were changed. The API will pass this extra field to model (what this PR is doing). The model will do the logic to figure out whether to start jobs or not.

Blocked by: 
- https://github.com/screwdriver-cd/data-schema/pull/213
- https://github.com/screwdriver-cd/config-parser/pull/56
- scm change

Related: https://github.com/screwdriver-cd/screwdriver/issues/911


package.json
```diff
@@ -87,14 +87,14 @@
...
+    "screwdriver-executor-docker": "^2.3.2",
+    "screwdriver-executor-k8s": "^10.7.2",
     "screwdriver-executor-k8s-vm": "^1.0.0",
     "screwdriver-executor-queue": "^1.1.0",
     "screwdriver-executor-router": "^1.0.0",
-    "screwdriver-models": "^26.8.4",
+    "screwdriver-models": "^26.8.6",
     "screwdriver-notifications-email": "^1.1.2",
     "screwdriver-notifications-slack": "^2.1.2",
     "screwdriver-scm-github": "^5.3.2",
```
tkyi: Will need to bump this to `^6.0.0`

d2lam: I want this change to be independent of the models. It can pass in the extra param without models doing anything with it. Then after confirming this doesn't break anything, we can pull in the new model

tkyi: Talking about `screwdriver-scm-github`

d2lam: Apparently `scm-github` was not bumped major. Keeping it as `5.3` now

PR #941, opened 2018-03-23
nomad executor

ethan-haynes:
## Context

This allows for a user to take advantage of a nomad executor in their screwdriver CICD pipeline. 

## References
https://www.npmjs.com/package/screwdriver-executor-nomad


config/default.yaml
```diff
@@ -80,6 +80,19 @@ executor:
                     high: 12
         # Container tags to use
         launchVersion: stable
+    nomad:
+      enabled: true
+      options:
+        nomad:
+            # The host or IP of the nomad cluster
+            host: http://10.75.130.53:4646/v1/jobs
```
tkyi: This should probably be some fake value here

config/default.yaml
```diff
@@ -80,6 +80,19 @@ executor:
                     high: 12
         # Container tags to use
         launchVersion: stable
+    nomad:
```
minz1027: Can you comment this part out just like jenkins? Our team is not supporting this by default. If someone wants to use it, can use the custom-environment-variables.yaml.

config/custom-environment-variables.yaml
```diff
@@ -94,6 +94,24 @@ executor:
         launchVersion: LAUNCH_VERSION
         # Prefix to the pod
         prefix: EXECUTOR_PREFIX
+    nomad:
+      enabled: EXECUTOR_NOMAD_ENABLED
+      options:
+        nomad:
+            # The host or IP of the kubernetes cluster
```
d2lam: Change comment

Issue #924, opened 2018-03-16
Template  - Delete template


jithin1987:
## Context

Once a template is published there is not template lifecycle management functionality for users. If a user wants to delete the template he has no option.

## Objective
1. API to delete a template. Only the owner of the template must be allowed to delete.
1. Expose in UI functionality to delete template.
1. Warn template owner if there are active pipelines using the template.

pranavrc:
Splitting up tasks here for convenience:

- [ ] Setup templates in local environment for testing.
- [ ] Create new module w/tests to delete templates through the API (https://github.com/screwdriver-cd/screwdriver/tree/master/plugins/templates).
- [ ] Design a delete button for the templates page in the UI.
- [ ] Get feedback on design.
- [ ] Implement a delete button for the templates page in the UI.
- [ ] Test everything.
- [ ] Verify everything.

ScottyFillups:
Regarding:

> 3. Warn template owner if there are active pipelines using the template.

For now, we'll warn users that pipelines may depend on their template when they attempt to delete it. We'll add template dependents in another pass; implementing it now seems like a bit of a stretch for this story, and is very similar issue https://github.com/screwdriver-cd/screwdriver/issues/804

ScottyFillups:
@drewfish 

https://github.com/screwdriver-cd/template-main/pull/15

tkyi:
The delete template API and UI portion is finished and verified.
![687474703a2f2f672e7265636f726469742e636f2f6462677077423537337a2e676966](https://user-images.githubusercontent.com/3230529/38328209-23614010-37ff-11e8-9b4b-a37805874b42.gif)

PR #948, opened 2018-03-27
feat(924): API endpoint to delete templates

pranavrc:
## Context

https://github.com/screwdriver-cd/screwdriver/issues/924

```
Once a template is published there is no template management functionality for users. If a user wants to delete the template he has no option.
```

## Objective

- [x] API to delete a template. Only the owner of the template must be allowed to delete.
- [x] Unit tests
- [x] Update template README

## References

Github issue - https://github.com/screwdriver-cd/screwdriver/issues/924
Template lifecycle - https://github.com/screwdriver-cd/screwdriver/issues/616


plugins/templates/remove.js
```diff
@@ -0,0 +1,71 @@
...
+module.exports = () => ({
+    method: 'DELETE',
+    path: '/templates/{name}',
+    config: {
+        description: 'Delete a template',
+        notes: 'Returns null if successful',
+        tags: ['api', 'templates'],
+        auth: {
+            strategies: ['token'],
+            scope: ['user', '!guest']
```
d2lam: So only `user` can call this endpoint?
The permission is tied to the pipeline, and builds belong to that pipeline can publish, so I don't see why it's not allowed to delete?

plugins/templates/remove.js
```diff
@@ -0,0 +1,71 @@
...
+                    return user.getPermissions(pipeline.scmUri);
+                })
+                    .then((permissions) => {
+                        if (!permissions.admin) {
+                            throw boom.unauthorized(`User ${username} `
+                                + 'does not have admin permission for this repo');
+                        }
+                    })
+                    .then(() => {
+                        templates.forEach(template => template.remove());
```
d2lam: `template.remove()` return a promise. You would need to use `Promise.all`, otherwise this won't wait.

plugins/templates/remove.js
```diff
@@ -0,0 +1,73 @@
...
+                    if (!user) {
+                        throw boom.notFound(`User ${username} does not exist`);
+                    }
+
+                    return user.getPermissions(pipeline.scmUri);
+                })
+                    .then((permissions) => {
+                        if (!permissions.admin) {
+                            throw boom.unauthorized(`User ${username} `
+                                + 'does not have admin permission for this repo');
```
tk3fftk: this `template`?

ScottyFillups: Good catch! Thanks 😄

plugins/templates/remove.js
```diff
@@ -0,0 +1,73 @@
...
+
+                return Promise.all([
+                    userFactory.get({ username, scmContext }),
+                    pipelineFactory.get(templates[0].pipelineId)
+                ]).then(([user, pipeline]) => {
+                    if (!user) {
+                        throw boom.notFound(`User ${username} does not exist`);
+                    }
+
+                    return user.getPermissions(pipeline.scmUri);
```
d2lam: This would only apply to `user` scope. For `build` scope, the check for permission would be different. 
You can take a look at this function for reference: https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/secrets/index.js#L44-L63

plugins/templates/README.md
```diff
@@ -83,7 +83,18 @@ Example payload:
...
+Template tag allows fetching on template version by tag. For example, tag `mytemplate@1.1.0` as `stable`.
+
+##### Delete a template
+Deleting a template will delete a template and all of its associated tags and versions.
+
+`DELETE /templates/{name}`
+
+###### Arguments
+
+'name'
```
tkyi: can remove this line

plugins/templates/index.js
```diff
@@ -16,13 +18,62 @@ const removeTagRoute = require('./removeTag');
...
+     * @param {Array}   credentials.scope        Scope of the credential (user, build, admin)
+     * @param {String}  [credentials.pipelineId] If credential is a build, this is the pipeline ID
+     * @param {Object}  template                 Target template object
+     * @param {String}  permission               Required permission level
+     * @return {Promise}
+     */
+    server.expose('canRemove', (credentials, template, permission) => {
+        const { username, scmContext, scope } = credentials;
+        const { userFactory, pipelineFactory } = server.root.app;
+        const forbiddenMsg = `User ${username} does not have ` +
```
d2lam: You use this same error message for `build` scope (line 62), so the message will be kinda weird since it's not a `User`

plugins/templates/README.md
```diff
@@ -83,7 +83,16 @@ Example payload:
 ```
 
 #### Template Tag
-Template tag allows fetching on template version by tag. For example, tag `mytemplate@1.1.0` as `stable`. 
+Template tag allows fetching on template version by tag. For example, tag `mytemplate@1.1.0` as `stable`.
+
+##### Delete a template
```
tkyi: Actually, the `##### Delete a template` section should probably go under the `#### Template` section rather than `#### Template Tag` section.

Issue #969, opened 2018-04-02
Delete associated templates and tags upon pipeline removal


ScottyFillups:
## Context

Templates are associated with a pipeline. Currently, when you delete a pipeline, Screwdriver _only_ deletes the pipeline. Any templates and template tags linked to that pipeline are not cleaned up: https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/pipelines/remove.js#L52

Since the pipeline model is used for checking build and user credentials, templates with deleted pipelines are essentially "unreachable"; you cannot delete or modify them.

## Objective

* Remove templates and template tags associated with a pipeline when it's deleted

ScottyFillups:
After some discussion (see https://github.com/screwdriver-cd/screwdriver/pull/970) we've concluded that this feature is too risky; users could easily break other people's pipelines inadvertently (eg. A user deletes a pipeline responsible for publishing 100+ popular templates; all those templates are then deleted as well).

We'll revisit how we store templates SOON™

Issue #967, opened 2018-04-02
Delete template UX


tkyi:
## Proposal
When a user doesn't have permissions to delete a template, there shouldn't be an option to do so (the icon shouldn't show up).

PR #970, opened 2018-04-02
[DNM] fix(969): Clean up templates and template tags when pipeline deleted

ScottyFillups:
## Context

From https://github.com/screwdriver-cd/screwdriver/issues/969:

> Templates are associated with a pipeline. Currently, when you delete a pipeline, Screwdriver _only_ deletes the pipeline. Any templates and template tags linked to that pipeline are not cleaned up: https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/pipelines/remove.js#L52

 > Since the pipeline model is used for checking build and user credentials, templates with deleted pipelines are essentially "unreachable"; you cannot delete or modify them.

## Objective

* Remove templates and template tags associated with a pipeline when it's deleted

## Related links

* Story: https://github.com/screwdriver-cd/screwdriver/issues/924

### DIFF
--- a/plugins/pipelines/README.md
+++ b/plugins/pipelines/README.md
@@ -72,6 +72,7 @@ Example payload:
 ```
 
 #### Delete a pipeline
+* Deleting a pipeline will also delete all dependent templates and template tags
 
 `DELETE /pipelines/{id}`
 
--- a/plugins/pipelines/remove.js
+++ b/plugins/pipelines/remove.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const hoek = require('hoek');
 const boom = require('boom');
 const joi = require('joi');
 const schema = require('screwdriver-data-schema');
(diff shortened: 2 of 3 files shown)


ScottyFillups:
@d2lam 

I completely agree that this PR gives users a lot of power. That being said, you could argue that it's the user's responsibility to delete pipelines conservatively. Regardless of whether this gets merged or not, it's going to force us to re-consider how we store templates.

IMHO I'd prefer if templates were tied only to users. I see templates as the "CD equivalent" of NPM packages. Why does a template have to be tied to a single pipeline? Why can't template authors have multiple pipelines for a given template?

The main purpose behind this PR is really just to address "dead" templates on the template registry page which cannot be deleted. Alternatively, we could just grey out dead templates, or its delete button.

ScottyFillups:
We could also just hide the delete button, inspired from this issue here: https://github.com/screwdriver-cd/screwdriver/issues/967

Closing the PR for now.

d2lam:
I agree with you.

PR #986, opened 2018-04-06
fix: bump artifact-bookend version

sakka2:
I fixed artifact-bookend, so I updated package.json to use latest artifact-bookend package.

Related: https://github.com/screwdriver-cd/artifact-bookend/pull/13


d2lam:
Docker hub is having an incident https://status.docker.com/pages/history/533c6539221ae15e3f000031 so I haven’t been able to push out any changes...

sakka2:
I see. Any time is OK.

Issue #485, opened 2017-03-04
API, UI, and Launcher should display current version number


stjohnjohnson:
This is to help debugging.

PR #988, opened 2018-04-06
[DNM] fix: Fix /versions

tkyi:
## Context

The `/versions` endpoint only works locally, not in production, because node modules are stored differently locally. Need to default to checking for node modules two levels up, because there is always node modules locally.

## Objective

This PR defaults to checking two levels up for the node modules.

## References

Related to https://github.com/screwdriver-cd/screwdriver/pull/986
Original issue: https://github.com/screwdriver-cd/screwdriver/issues/485


tkyi:
Not needed.

PR #990, opened 2018-04-07
feat(deleteCommands): API endpoint to delete commands

pranavrc:
## Context

There's no option right now to delete commands from the UI once they're published. This PR adds an API endpoint for command deletion.

## Objective

- [x] API to delete a command. Only the owner of the command (pipeline admin) must be allowed to delete.
- [x] Unit tests
- [x] Update command README

## References

Commands doc - https://docs.screwdriver.cd/user-guide/commands.html

Blocked by: https://github.com/screwdriver-cd/store/pull/41


plugins/commands/remove.js
```diff
@@ -0,0 +1,55 @@
...
+            ]).then(([commands, tags]) => {
+                if (commands.length === 0) {
+                    throw boom.notFound(`Command ${namespace}/${name} does not exist`);
+                }
+
+                return canRemove(credentials, commands[0], 'admin').then(() => {
+                    const commandPromises = commands.map(command => command.remove());
+                    const tagPromises = tags.map(tag => tag.remove());
+
+                    return Promise.all(commandPromises.concat(tagPromises));
```
tkyi: I think you should also remove the command binaries from the store. So you'll have to make some call to the store similar to https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/commands/create.js#L43-L51
But I don't think that endpoint exists yet in the store (https://github.com/screwdriver-cd/store/blob/master/plugins/commands.js). You might want to coordinate with Yahoo Japan. Not sure if they're already working on this.

plugins/commands/remove.js
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * Remove command from store and API
+ * @method removeCommand
+ * @param  {Object}   command         The Command object
+ * @param  {String}   storeUrl        URL to the store
+ * @param  {String}   authToken       Bearer Token to be passed to the store
+ * @return {Promise}
+ */
+function removeCommand(command, storeUrl, authToken) {
```
d2lam: You can put `function removeCommand({namespace, name, version}, storeUrl, authToken)` here

plugins/commands/remove.js
```diff
@@ -0,0 +1,95 @@
...
+    const options = {
+        url: `${storeUrl}/v1/commands/${command.namespace}/${command.name}/${command.version}`,
+        method: 'DELETE',
```
d2lam: Does this endpoint exist? Can't find it.

d2lam:
@pranavrc  Can you put Blocked by: https://github.com/screwdriver-cd/store/pull/41 in your PR description

plugins/commands/remove.js
pranavrc: There's a `command.remove()` later on in the function

pranavrc:
Dismissing @tkyi's review since she's OOO for a while and her comments have been addressed.

Issue #115, opened 2016-08-19
Swagger response schemas missing for auth, status, and stats


nicolaifsf:
response schemas for login, status, and stats are currently not defined

Filbird:
Response schemas (for GETs) are also missing for **versions**, **contexts**, **badge**, and **artifacts**.

Filbird:
As **badge** and **artifacts** are redirects, it may not make sense to add response schemas for theseendpoints.

The majority of non GET endpoints do not have response schemas defined. Likewise, the usefulness of adding response schemas for these endpoints is questionable in my opinion.

Filbird:
Since we've covered the bulk of the response schemas, I'll be closing this PR.

Please feel free to reopen if you feel that more response schemas are needed.

PR #992, opened 2018-04-09
fix(115): Add response schemas for versions, contexts, stats, and status

Filbird:
## Context

There are response object schemas missing for several endpoints as detailed in screwdriver-cd/screwdriver#115

Pending merging of https://github.com/screwdriver-cd/data-schema/pull/222

## Objective

- [x] Add response object schemas for:
  - [x] versions
  - [x] contexts
  - [x] stats
  - [x] status 
- [x] Remove incorrect response schema for step update
- [x] Add auth tag to login and logout routes

## References

Swagger documentation:
https://api.screwdriver.cd/v4/documentation

### DIFF
--- a/plugins/builds/steps/update.js
+++ b/plugins/builds/steps/update.js
@@ -57,9 +57,6 @@ module.exports = () => ({
                 .then(build => reply(build.steps[stepIndex]).code(200))
                 .catch(err => reply(boom.wrap(err)));
         },
-        response: {
-            schema: schema.models.build.getStep
-        },
         validate: {
             params: schema.api.loglines.params,
             payload: schema.models.build.updateStep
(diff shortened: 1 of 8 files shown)


plugins/builds/steps/update.js
```diff
@@ -57,9 +57,6 @@ module.exports = () => ({
                 .then(build => reply(build.steps[stepIndex]).code(200))
                 .catch(err => reply(boom.wrap(err)));
         },
-        response: {
```
FenrirUnbound: I might be missing context on this, but why was this particular schema dropped?

Filbird: I may be wrong but it seems like this response schema was intended only for GETs on this endpoint.

tkyi: I think the schema still applies to the PUT endpoint; they were probably just reusing it

Filbird: ooh, my mistake, I will remove that commit!

PR #1001, opened 2018-04-10
fix: Echo docker tag

tkyi:
This will help with debugging


screwdriver.yaml
```diff
@@ -24,6 +24,7 @@ jobs:
             - save-tag-to-meta: |
                 NPM_PKG_VERSION=`npm dist-tag ls | awk -F: '{ print $2 }' | tr -d '[:space:]'`
                 meta set docker_tag "v$NPM_PKG_VERSION"
+                echo `meta get docker_tag`
```
d2lam: we can't just do `meta get`? that should print out right?

PR #1004, opened 2018-04-11
fix: bump artifact-bookend

sakka2:
## Context
Artifact-bookend has updated. So I want to use the latest version.
https://github.com/screwdriver-cd/artifact-bookend/pull/14


sakka2:
Thank you!

PR #1012, opened 2018-04-13
fix: fix functional tests for the new workflow

d2lam:
Currently, it's starting the wrong job:
screen shot 2018-04-13 at 11 51 08 am

the `/jobs` endpoint sorts by `workflow` if `workflow` is there, but since we don't have it anymore, the order is not guaranteed. We cannot get the first element and assume it's the `main` job. 

Instead, we can create a new event with `startFrom` as `main`, and it will create the `main` build.

### DIFF
--- a/features/step_definitions/authorization.js
+++ b/features/step_definitions/authorization.js
@@ -107,7 +107,7 @@ defineSupportCode(({ Before, Given, Then }) => {
             json: true
         })).then((resp) => {
             Assert.equal(resp.statusCode, 201);
-            this.buildId = resp.body.id;
+            this.buildId = resp.body[0].id;
         });
     });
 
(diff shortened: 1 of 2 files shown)


features/step_definitions/authorization.js
```diff
@@ -85,35 +85,30 @@ defineSupportCode(({ Before, Given, Then }) => {
...
-                    Assert.equal(resp.statusCode, 201);
-
-                    this.buildId = resp.body.id;
-                })
-            );
+            },
+            json: true
+        })).then((resp) => {
+            Assert.equal(resp.statusCode, 201);
+            this.buildId = resp.body.id;
```
minz1027: should be an array of builds

Issue #688, opened 2017-08-08
Build Periodically


petey:
# Context
As a user,
I would like to schedule a pipeline to execute periodically,
So I can ensure that my code and dependencies are always up-to-date.

## Objective
Allow the user to supply an annotation describing a cron-style string which describes the interval for which a job should run. This interval should *not* be allowed to occur more than once hourly. We can enforce this by requiring the "minutes" configuration to be `H`.

### H Explanation
Note: To allow periodically scheduled tasks to produce even load on the system, the symbol "H" (for "hash") should be used wherever possible. For example, using 0 0 * * * for a dozen daily jobs will cause a large spike at midnight UTC. In contrast, using H H * * * would still execute each job once a day, but not all at the same time, better using shared resources. The "H" symbol can be used with a range. For example, H H(0-7) * * * means some time between 12:00 AM UTC (midnight) to 7:59 AM UTC. You can also use step intervals with "H", with or without ranges.

### Tasks
- [x] Spike
- [x] Define a workflow annotation for storing periodic build configuration
- [x] A workflow automatically creates an event when the periodic build configuration triggers
- [x] Enable `H` configurations
- [x] Document annotation
- [ ] Monitor frequently running pipelines so that we can detect abuse

minz1027:
### Possible solution
- Kubernetes cron job
- micro service
- at queue level

pranavrc:
@Filbird and I were looking into this, and https://github.com/resque/resque-scheduler#dynamic-schedules seems like the cleanest way to implement this since it would happen at the executor-queue level when we create the job in the queue. It supports recurring jobs with dynamic schedules.

Filbird:
**Update 4/18**

Implementing this at a the `executor-queue` [level](https://github.com/screwdriver-cd/executor-queue/blob/master/index.js#L64) requires us to persist the JWT beyond its intended lifespan. One idea is to generate a new JWT by using the OAuth token to hit the `/v4/auth/login/` endpoint as detailed [here](https://docs.screwdriver.cd/user-guide/api#authentication-and-authorization). The JWT would then be stored in Redis prior to enqueueing the scheduled build in Resque.

Several complications:
- How do we get a unique build ID for every scheduled build?
- How do we ensure only a single instance of the scheduler is enqueueing builds?

Filbird:
**Update 4/19**

[...]

PR #1032, opened 2018-04-25
[4] feat: support build periodically

d2lam:
## Context
Need to pass additional `tokenGen` for the executor queue to use. 
Ideally, we want to pass it directly to the executor plugin. However, the `executor` plugin is registered **before** the `auth` plugin. If we register the executor plugin **after** the `auth` plugin, then other plugins (like buildFactory) cannot get `executor` plugin yet. To pass it directly to the executor requires a refactor. 

## References
https://github.com/screwdriver-cd/screwdriver/issues/688


d2lam:
Let's merge this last so we can pull in the new models at the same time.

Issue #1007, opened 2018-04-12
Request: Code coverage and Test result


cynthiax:
**What feature**:
Right now Screwdriver is not showing Code coverage nor Test result for each build. As a person who cares about code quality and hoping to see each commit would increase the code coverage or at least have it stay the same, I would want a easy way to view it inside Screwdriver. Similarly for test result, I would like to know if all test passed, if not, which one failed at a easy glance in Screwdriver.

**Detail**:
There are different service different developer could use, like coverall mentioned in https://github.com/screwdriver-cd/screwdriver/issues/500, [codecov](https://codecov.io/) and [sonar](https://www.sonarqube.org/). Screwdriver should provide (default)plugin for using these services.

- [x] Spike on how to integrate different code analysis service in Screwdriver 
        Criteria: costs, ease of setup

tkyi:
There's a related issue open already: https://github.com/screwdriver-cd/screwdriver/issues/674

minz1027:
## Updates 04/30
After talking with the team, we have decided to go with SonarQube as default code coverage plugin. 

### Proposed implementation:
Flow: 
- Build has a teardown step `push-code-coverage` 
- Inside the step
   - build will call screwdriver API and use build `jwt` to trade for a sonar user token
   - push to sonar host with the user token
Notes: the user token will only be authorized to push its own project. Every pipeline maps to a project in sonar, and every single PR has its own project.

First pass
To do:
- [ ] sonar pipeline with customized config
- [ ] code-coverage-base and code-coverage-sonar
- [ ] `push-code-coverage` bookend

tkyi:
This feature is working in production:
<img width="1440" alt="screen shot 2018-05-15 at 11 29 50 am" src="https://user-images.githubusercontent.com/3230529/40078501-9aab183a-5839-11e8-9df3-ddafd103873a.png">
Clicking on the link takes you to the SonarQube project page:
<img width="1440" alt="screen shot 2018-05-15 at 11 30 14 am" src="https://user-images.githubusercontent.com/3230529/40078515-a9af35f0-5839-11e8-88f6-2ccb2ddddf30.png">
Documentation can be found: https://docs.screwdriver.cd/user-guide/configuration/coverage

marian-r:
[...]

PR #1050, opened 2018-05-07
feat: support coverage plugin

minz1027:
support coverage plugin
- if coverage is configured, it will register new routes for `/coverage/token`
- by default, the coverage bookend is not pulled in. will modify the env `BOOKENDS_TEARDOWN ` to pull it in: https://github.com/screwdriver-cd/screwdriver/blob/master/config/custom-environment-variables.yaml#L293

Related to: https://github.com/screwdriver-cd/screwdriver/issues/1007


plugins/coverage/token.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+const boom = require('boom');
+
+module.exports = config => ({
+    method: 'GET',
+    path: '/coverage/token',
+    config: {
+        description: 'Get an access token to talk to coverage server',
+        notes: 'Returns a job record',
```
tkyi: Returns a `token`

plugins/coverage/token.js
```diff
@@ -0,0 +1,29 @@
...
+        plugins: {
+            'hapi-swagger': {
+                security: [{ token: [] }]
+            }
+        },
+        handler: (request, reply) => {
+            const buildCredentials = request.auth.credentials;
+
+            return config.coveragePlugin.getAccessToken(buildCredentials)
+                .then(reply)
```
tkyi: so we'll just return a string here?

minz1027: yes, I also update the description for the endpoint to make that clear

PR #1052, opened 2018-05-07
fix: add remote trigger for artifact bookend

jithin1987:
## Contexts

API needs to be rebuilt after it's dependencies like `artifact-bookend` are published to npm.

### DIFF
--- a/screwdriver.yaml
+++ b/screwdriver.yaml
@@ -3,7 +3,11 @@ shared:
 
 jobs:
     main:
-        requires: [~pr, ~commit]
+        requires: [
+            ~pr,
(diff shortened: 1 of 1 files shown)


screwdriver.yaml
```diff
@@ -3,7 +3,11 @@ shared:
 
 jobs:
     main:
-        requires: [~pr, ~commit]
+        requires: [
```
minz1027: how about change it to this format 
```
- ~pr
-  ~commit
- ....
```?
That's easier to read

jithin1987: added

Issue #854, opened 2018-01-24
Some pipelines would be permanently unavailable depending on cases


catto:
**What happened**:
* `User A` has GitHub account and logged in to a Screwdriver.cd instance.
* `User A` has created new pipeline.
* Users who have write access to the repo can start the pipeline.
* `User A` removes or renames his/her GitHub account. (including the suspended state for GitHub Enterprise)
* Now no one can start the pipeline because of no admins for it. UI shows a JSON including a message like "suspended"

**What you expected to happen**:
Users who have write access can start build when a creator of the pipeline is not available.

**How to reproduce it**:
These steps are for an environment with GitHub Enterprise.
* Create new user on your GitHub Enterprise instance.
* Create new pipeline on Screwdriver.cd instance with the created user.
* Suspend the GitHub account
* No one can start new build for the pipeline even if there are admins of the repository.

catto:
possible solution: 
* Add new admins to a pipeline when operations to the pipeline occurs
* Try next admin if API accesses are failed with auth error.
* Remove the failed admin from the pipeline admins.

stjohnjohnson:
Good catch, @catto.  I recommend we actually do all three of those.

yoshwata:
@minz1027 @d2lam @tkyi @jithin1987 Could someone close this issue?

yoshwata:
This is related issue.
https://github.com/screwdriver-cd/screwdriver/issues/1117

catto:
I confirmed that the issue has been fixed. Thanks @yoshwata!

yoshwata:
@catto Thank you!

PR #1061, opened 2018-05-14
[3] fix(854): Add or delete an admin when pipeline operation occurs

yoshwata:
## Context

This pr fixes the issue [854](https://github.com/screwdriver-cd/screwdriver/issues/854)

## Objective

Now there is only one admin exists for a pipeline. If the admin release the permission of the repository of the pipeline, no one can starts the build. To fix it, I add operations like below.

- Add the user as an admin if the github permission is OK.
- Delete the user from admins if the github permission is not OK.

## References

https://github.com/screwdriver-cd/screwdriver/issues/854

## Related PRs
https://github.com/screwdriver-cd/executor-queue/pull/25
https://github.com/screwdriver-cd/models/pull/247
https://github.com/screwdriver-cd/screwdriver/pull/1061


test/plugins/builds.test.js
```diff
@@ -235,7 +235,10 @@ describe('build plugin test', () => {
                 scmRepo,
                 admins: { foo: true },
                 sync: sinon.stub().resolves(),
-                syncPR: sinon.stub().resolves()
+                syncPR: sinon.stub().resolves(),
+                admin: Promise.resolve({
```
tkyi: can probably do 
```
admin: sinon.stub().resolves({
    username: 'foo'
})
```
for consistency

plugins/builds/create.js
```diff
@@ -46,10 +46,14 @@ module.exports = () => ({
...
                         .then((permissions) => {
                             if (!permissions.push) {
+                                // the user who are not permitted is deleted from admins table
+                                delete pipeline.admins[username];
                                 throw boom.unauthorized(`User ${username} `
                             + 'does not have push permission for this repo');
                             }
                         })
+                        // user has good permissions, add the user as an admin
+                        .then(() => { pipeline.admins[username] = true; })
```
tkyi: Do you need to call `pipeline.update()` after this line and before the `pipeline.sync()`? Like here: https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/pipelines/update.js#L92-L93

yoshwata: I need it. I add `update` to call `sync`.  Thank you!

test/plugins/builds.test.js
yoshwata: Let me add more information. `admin` is not a function but a variable, so applying sinon here will be error. `builds.secrets` is also variable, and it applies `Promise`.  The `then` function appears only in case of function. In case of variable, `then` will be undefined, and test failed.

tkyi: ah I understand
thanks for explaining 🙇‍♀️

plugins/builds/create.js
```diff
@@ -46,10 +46,17 @@ module.exports = () => ({
...
                             + 'does not have push permission for this repo');
                             }
                         })
+                        // user has good permissions, add the user as an admin
+                        .then(() => {
+                            pipeline.admins[username] = true;
+                            pipeline.update();
+                        })
                         // user has good permissions, sync and create a build
                         .then(() => (job.isPR() ? pipeline.syncPR(job.prNum) : pipeline.sync()))
```
tkyi: do you need to save the `updatedPipeline` and use it to syncPR()? like
```
.then((updatedPipeline) => (job.isPR() ? updatedPipeline.syncPR(job.prNum) : updatedPipeline.sync()))
```

plugins/builds/index.js
```diff
@@ -221,33 +221,35 @@ exports.register = (server, options, next) => {
...
+                .then((adminName) => {
+                    const scmUri = pipeline.scmUri;
+                    const admin = adminName;
+                    const scmContext = pipeline.scmContext;
+
+                    payload.scmContext = scmContext;
+                    payload.username = admin;
+
+                    // get pipeline admin's token
+                    return userFactory.get({ username: admin, scmContext })
```
minz1027: No need to do this since `pipeline.admin` is already an user object, should be able to do `readAdmin.unsealToken()` directly. https://github.com/screwdriver-cd/models/pull/247/files

yoshwata: I fixed. Thank you!

minz1027:
@yokawara Can we also add/delete admins in `/plugins/events/create` and `/plugins/pipelines/sync`? the `builds/create` endpoint is not used very often... now most things are event based, and i think a `sync` should also fix the admin problem

yoshwata:
@minz1027 Thank you. I will confirm that.

Issue #1030, opened 2018-04-25
JWT is hardcoded to expire in 2 hours


d2lam:
**What happened**:
JWT is hardcoded to expire in 2 hours.

**What you expected to happen**:
This value should be configurable based on the build timeout. It does not make sense that JWT expires before the build timeout value. For example, cluster admin set build timeout to 10 hours, but the JWT can only be valid in 2 hours.

jithin1987:
Not JWT expiration is aligned with build timeout.

Issue #998, opened 2018-04-10
Build artifacts cannot be saved after 2 hours from starting a build


yoshwata:
**What happened**:
- If a build takes more than 2 hours, the logs are not saved on store. 
- It is also a problem that a user cannot notice the situation.
  - It looks a build is stopped, but the build is proceeding actually.

The cause might be a expired token.
https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/auth/index.js#L18
It might be reasonable that SDadmin can change this variable, but that is now fixed value.

**What you expected to happen**:
Token's expired period can be set by SDadmin, and store can save logs in the period.
or
Show messages that store does not save the build logs anymore.

**How to reproduce it**:
Start build that spends more than 2 hours.

sakka2:
To resolve this Issue, I fixed Screwdriver API and models.

API: https://github.com/screwdriver-cd/screwdriver/pull/1062
models: https://github.com/screwdriver-cd/models/pull/245

## Objective
Expires time of JWT should be configurable based on the build timeout.
- Add `expiresSec` argument into [generateToken](https://github.com/screwdriver-cd/screwdriver/pull/1062/files#diff-cc45542b37e76ced3e366e4aa20f512cR91) method.
- `expiresSec` argument is defined in models, [lib/build.js#start()](https://github.com/screwdriver-cd/models/pull/245/files#diff-26b681bd4dac10f25d5293c7595c435bR154) method.
- `expiresSec` is configured based on build timeout value which is set by annotation or cluster config. If neither of them is set, default value (90minutes) will be set.
- Max value of `expiresSec` is 12 hours. Even if build timeout value which is set by annotation or cluster config is over 12 hours, JWT expires time value will be 12 hours.
- `expiresSec` will have buffer time (5minutes) more than build time out value, because sometimes it will take a time to start build since JWT is created.

jithin1987:
Should be fixed with temporal JWT's

PR #1062, opened 2018-05-14
fix(1030): Expires time of JWT should be configurable based on the build timeout.

sakka2:
## Context
JWT is hardcoded to expire in 2 hours.
So, there is a case that JWT expires before the build timeout.
See this [issue](https://github.com/screwdriver-cd/screwdriver/issues/998).

## Objective
Expires time of JWT should be configurable based on the build timeout.
- Add `expiresIn` argument into [generateToken](https://github.com/screwdriver-cd/screwdriver/pull/1062/files#diff-cc45542b37e76ced3e366e4aa20f512cR91) method.
- `expiresIn` argument is defined in models, [lib/build.js#start()](https://github.com/screwdriver-cd/models/pull/245/files#diff-26b681bd4dac10f25d5293c7595c435bR153) method.
- `expiresIn` is configured based on build timeout value which is set by annotation or cluster config. If neither of them is set, default value (90minutes) will be set.
- Build timeout setting of annotation must not be over the cluster config value (`maxBuildTimeout`).  So if it is over it, value of `maxBuildTimeout` will be set.
- `expiresIn` will have buffer time (5minutes) more than build time out value, because sometimes it will take a time to start build since JWT is created.

## References
- Issues: 
　https://github.com/screwdriver-cd/screwdriver/issues/998
　https://github.com/screwdriver-cd/screwdriver/issues/1030
- Related: (PR in screwdriver models)
　https://github.com/screwdriver-cd/models/pull/245


plugins/auth/index.js
```diff
@@ -15,7 +15,9 @@ const sugar = require('hapi-auth-cookie');
 const tokenRoute = require('./token');
 const uuid = require('uuid/v4');
 
-const EXPIRES_IN = '2h';
+const EXPIRES_IN = 7200; // 2h
+const MAX_EXPIRES_IN = 43200; // 12h
+const EXPIRES_BUFFER = 300; // 5m
```
sakka2: Expires time of JWT will have buffer time ( 5m )

plugins/auth/index.js
```diff
@@ -15,7 +15,9 @@ const sugar = require('hapi-auth-cookie');
 const tokenRoute = require('./token');
 const uuid = require('uuid/v4');
 
-const EXPIRES_IN = '2h';
+const EXPIRES_IN = 7200; // 2h
```
sakka2: Changed from string to integer in order to be calculated.

plugins/auth/index.js
```diff
@@ -83,13 +85,20 @@ exports.register = (server, options, next) => {
...
-        jwtid: uuid()
-    }));
+    server.expose('generateToken', (profile, expiresSec = EXPIRES_IN) => {
+        const expiresIn = (expiresSec <= MAX_EXPIRES_IN)
+            ? expiresSec + EXPIRES_BUFFER
+            : MAX_EXPIRES_IN;
+
+        return jwt.sign(profile, pluginOptions.jwtPrivateKey, {
+            algorithm: ALGORITHM,
+            expiresIn,
```
sakka2: `expiresIn` arguments of `jwt.sign() ` can be set by seconds.
https://www.npmjs.com/package/jsonwebtoken

lib/server.js
```diff
@@ -123,9 +123,9 @@ module.exports = (config) => {
             // Initialize common data in buildFactory and jobFactory
             if (server.app.buildFactory) {
                 server.app.buildFactory.apiUri = server.info.uri;
-                server.app.buildFactory.tokenGen = (buildId, metadata, scmContext) =>
+                server.app.buildFactory.tokenGen = (buildId, metadata, scmContext, expiresSec) =>
```
sakka2: `tokenGen()` method is called in [models](https://github.com/screwdriver-cd/models/pull/245/files#diff-26b681bd4dac10f25d5293c7595c435bR170)

plugins/auth/index.js
```diff
@@ -83,13 +85,20 @@ exports.register = (server, options, next) => {
...
      */
-    server.expose('generateToken', profile => jwt.sign(profile, pluginOptions.jwtPrivateKey, {
-        algorithm: ALGORITHM,
-        expiresIn: EXPIRES_IN,
-        jwtid: uuid()
-    }));
+    server.expose('generateToken', (profile, expiresSec = EXPIRES_IN) => {
+        const expiresIn = (expiresSec <= MAX_EXPIRES_IN)
+            ? expiresSec + EXPIRES_BUFFER
+            : MAX_EXPIRES_IN;
```
wdstar: We may as well plus the `EXPIRES_BUFFER` to the `MAX_EXPIRES_IN` value too.

plugins/auth/index.js
```diff
@@ -83,13 +85,20 @@ exports.register = (server, options, next) => {
      * Generates a jwt that is signed and has a 2h lifespan
      * @method generateToken
      * @param  {Object} profile Object from generateProfile
+     * @param  {Integer} expires_sec JWT Expires time (must be seconds)
```
s-yoshika: `expires_sec` is missing, maybe `expiresSec`

plugins/auth/index.js
wdstar: Missing label? `expiresIn: expiresIn,`

wdstar:
We should add the buffer time to it?
https://github.com/screwdriver-cd/screwdriver/pull/1062/files#diff-cc45542b37e76ced3e366e4aa20f512cR138

plugins/auth/index.js
s-yoshika: Please align thes lines.

plugins/auth/index.js
```diff
@@ -15,7 +15,9 @@ const sugar = require('hapi-auth-cookie');
 const tokenRoute = require('./token');
 const uuid = require('uuid/v4');
 
-const EXPIRES_IN = '2h';
+const EXPIRES_IN = 7200; // 2h
+const MAX_EXPIRES_IN = 43200; // 12h
```
s-yoshika: It is better to use a calculation formula so that there is no mistake.
https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/auth/index.js#L121

plugins/auth/index.js
sakka2: I see, I fixed!

plugins/auth/index.js
sakka2: I fixed!

plugins/auth/index.js
sakka2: That's right. I fixed!

plugins/auth/index.js
sakka2: When label and value is same, it can be omitted in ES6.

plugins/auth/index.js
```diff
@@ -126,7 +135,7 @@ exports.register = (server, options, next) => {
                 key: pluginOptions.jwtPublicKey,
                 verifyOptions: {
                     algorithms: [ALGORITHM],
-                    maxAge: EXPIRES_IN
+                    maxAge: EXPIRES_IN + EXPIRES_BUFFER
```
s-yoshika: Sorry, I don't known as well, but `maxAge` for `verifyOption` never be changed?

sakka2: I will check this.

sakka2: I checked this.

According to [jsonwebtoken library](https://github.com/auth0/node-jsonwebtoken/blob/master/README.md#jwtverifytoken-secretorpublickey-options-callback) , `maxAge` is the maximum allowed age for tokens to still be valid. 
So, It should be the same value of `MAX_EXPIRES_IN (+ EXPIRES_BUFFER)`.

If `expiresIn` value in `generateToken()` is less than `maxAge`, JWT expires by `expiresIn` value.
And if `expiresIn` value in `generateToken()` is greater than `maxAge`,  it will become error.

lib/server.js
```diff
@@ -123,9 +123,9 @@ module.exports = (config) => {
...
                     server.plugins.auth.generateToken(server.plugins.auth
-                        .generateProfile(buildId, scmContext, ['build'], metadata)
+                        .generateProfile(buildId, scmContext, ['build'], metadata), expiresSec
                     );
             }
             if (server.app.jobFactory) {
```
d2lam: Don't we also need to add for `jobFactory.tokenGen`?

d2lam: It is used in here: https://github.com/screwdriver-cd/executor-queue/blob/master/index.js#L166
So we need to make the changes there as well.

sakka2: In `generalToken` method, argument `expiresSec` can be omitted.
https://github.com/screwdriver-cd/screwdriver/pull/1062/files#diff-cc45542b37e76ced3e366e4aa20f512cR91

If there is no `expiresSec` argument, `DEFAULT_EXPIRES_IN` value will be set. It is 2 hours, same as before.
I think if it is not during build event (like periodic events), it is OK that expires time of JWT is fixed 2 hours.

plugins/auth/index.js
```diff
@@ -15,7 +15,9 @@ const sugar = require('hapi-auth-cookie');
 const tokenRoute = require('./token');
 const uuid = require('uuid/v4');
 
-const EXPIRES_IN = '2h';
+const EXPIRES_IN = 2 * 60 * 60; // 2h in seconds
+const MAX_EXPIRES_IN = 12 * 60 * 60; // 12h in seconds
```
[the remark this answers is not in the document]
sakka2: I fixed `MAX_EXPIRES_IN` to be able to configured in cluster config setting.

plugins/auth/index.js
```diff
@@ -80,16 +82,30 @@ exports.register = (server, options, next) => {
...
-    server.expose('generateToken', profile => jwt.sign(profile, pluginOptions.jwtPrivateKey, {
-        algorithm: ALGORITHM,
-        expiresIn: EXPIRES_IN,
-        jwtid: uuid()
-    }));
+    server.expose('generateToken', (profile, expiresSec = DEFAULT_EXPIRES_IN) => {
+        const executor = server.root.app.buildFactory.executor;
+
+        // max build timeout is set only when executor is 'k8s' or 'k8s-vm'
+        const maxExpiresIn = (executor.kubernetes && executor.kubernetes.maxBuildTimeout)
```
jithin1987: can we base our logic on whether build timeout is present or not ? rather than explicitly based on executor value.
```
const buildTimeOut = getBuildTimeout(executor); 
const maxExpiresIn = buildTimeOut ? buildTimeOut * 60 : DEFAULT_MAX_EXPIRES_IN
```
`getBuildTimeout` can hide the details of explicit checks, without overloading this function.

sakka2: I fixed to use `getBuildTimeout()` method in models and use `executor` there, so it will work even with `executor-queue` .
And I also fixed the logic of `getBuildTimeout()` method.

Issue #966, opened 2018-04-02
feature: There is no way to announce the downtime.


yoshwata:
What happened:
It is useful if users are able to know about the downtime from web ui.  We'd like to show the message like `2018/4/2 10:00 - 11:00 : Screwdrivrer will down for maintenance` at top of pages.

What you expected to happen:
Screwdriver UI can show maintenance message.

tkyi:
## Proposal
We should add endpoints to update and delete banner messages in the Screwdriver UI. They should require `admin` permissions (https://github.com/screwdriver-cd/screwdriver/blob/master/config/default.yaml#L31).
1. __PUT__ endpoint to `/banner`
__body__: `{ message: <STRING> }`
2. __DELETE__ endpoint to `/banner`

## Tasks
- [ ] **data-schema**: Add a banner table and schema validation (somewhere here https://github.com/screwdriver-cd/data-schema/tree/master/models)
- [ ] **models**: Add banner model (should be very similar to https://github.com/screwdriver-cd/models/blob/master/lib/collectionFactory.js, https://github.com/screwdriver-cd/models/blob/master/lib/collection.js)
- [ ] **API**: 
  - [ ] Take in auth admin information
  - [ ] Add PUT and DELETE endpoints in a banner folder (see https://github.com/screwdriver-cd/screwdriver/tree/master/plugins) with banner README.md
- [ ] **UI**: Display banner on UI

PR #1066, opened 2018-05-16
fix(966): Remove auth for GET banners endpoints

tkyi:
## Context

Users shouldn't need to login to Screwdriver to see banner announcements.

## Objective

This PR removes auth for GETs to banners endpoints.

## References

Related links: https://github.com/screwdriver-cd/screwdriver/issues/966


plugins/banners/get.js
```diff
@@ -13,10 +13,6 @@ module.exports = () => ({
         description: 'Get a single banner',
         notes: 'Return a banner record',
         tags: ['api', 'banners'],
-        auth: {
-            strategies: ['token'],
-            scope: ['user', '!guest']
-        },
         plugins: {
             'hapi-swagger': {
                 security: [{ token: [] }]
```
jithin1987: should this be removed ?

plugins/banners/list.js
```diff
@@ -10,10 +10,6 @@ module.exports = () => ({
         description: 'Get banners',
         notes: 'Returns all banner records',
         tags: ['api', 'banners'],
-        auth: {
-            strategies: ['token'],
-            scope: ['user']
-        },
         plugins: {
             'hapi-swagger': {
                 security: [{ token: [] }]
```
jithin1987: should this be removed ?

plugins/banners/get.js
tkyi: removed 👍

plugins/banners/list.js
tkyi: removed

Issue #1075, opened 2018-05-22
Build JWT should be generated at execution phase


jithin1987:
## Context

API generates build JWT token when it sends a build config to an executor. This approach works as long as there is not much delay from the time API generates the build config and the time the build is actually executed. With uses cases like `blockedBy`, `freezeWindows` there could be significant delay from the time API generates build config and when the build is actually executed. Also when used with `executor-queue`, API does not have all relevant information to generate JWT, such as time out config for the underlying executor. This warrants a need to generate JWT token when an executor actually runs the build so that it's expiry time starts from actual build execution and not count for the time it's awaiting it's execution.

## Objective

1. API generates a temporal scoped JWT when it sends a build to executor.
2. Executor exchanges the temporal JWT with actual build scoped JWT.
3. Temporal scoped JWT should not be exposed inside a build.

## References
1. API Design https://github.com/screwdriver-cd/models/pull/245#issuecomment-390619885 Thanks @sakka2

sakka2:
@jithin1987 
> can they exchange temporal JWT for build JWT token without making network call ?

I thought it would be better not to make network call too, but Inside each executors, it is difficult to exchange JWT without calling new API.
The network should be internal and it calls only one time at the start of builds, so the network load will not be so much.

Another way I thought is that make If statement [here](https://github.com/screwdriver-cd/models/blob/master/lib/build.js#L214), before the `this[executor].start` method. Only in case the `executor-queue`, generate temporal JWT and in case the others, generate build JWT as before. This way do not need to call API in non-queue executor, but cannot avoid some case you mentioned above. And It seems making if statement may not a smart way.

Do you have any good idea?

jithin1987:
@sakka2 API call makes sense. Let each executor except `queue` have logic to exchange temporal JWT. Thanks for nailing down the flow.

sakka2:
@jithin1987 
I added new API and fixed executors to use it.
Could you review these changes please?

PR #1087, opened 2018-05-25
fix(1030): [4] Add API to generate JWT for build using temporal JWT

sakka2:
## Context
JWT is hardcoded to expire in 2 hours.
So, there is a case that JWT expires before the build timeout.
See this [issue #998](https://github.com/screwdriver-cd/screwdriver/issues/998).
See this [issue #1075](https://github.com/screwdriver-cd/screwdriver/issues/1075).

## Objective 
Expires time of JWT should be configurable based on the build timeout.
JWT token need to generate when an executor actually runs the build so that it's expiry time starts from actual build execution.

### 1.  Add the new API endpoint to generate build JWT.
**request_URL:**  `/v4/builds/{id}/token`
**request_body:** 
{
    `"expiresSec": "3700"`
}
**rquestHeader:** Authorization: Bearer `{temporal JWT}`

**response_body:**
{
    `"jwt": xxxxxxxxxxxxxxxxx (build JWT)`
}

### 2. In every build, firstly generate temporal JWT.
In `buildFactory.tokenGen()` method, make changes to always generate temporal JWT.
JWT scope will be changed from `['build']` to `['temporal']`.
https://github.com/screwdriver-cd/screwdriver/pull/1062/files#diff-c945a46d13b34fcaff544d966cffcabaR128

### 3.  In each executor, generate build JWT with calling new API.
For each `executor.start()` method,  add the process to call the new API endpoint for getting build JWT using temporal JWT passed in, except for `executor-queue`.
In case `executor-queue`, set the temporal JWT to Redis table, 
and in `queue-worker`, `executor.start()` method such as executor-k8s and executor-k8s-vm will call the new API with this temporal JWT as I wrote above.
https://github.com/screwdriver-cd/queue-worker/blob/master/lib/jobs.js#L49

### 4.   Add `expiresIn` argument into generationToken() method in API
Add `expiresIn` argument into [generateToken](https://github.com/screwdriver-cd/screwdriver/pull/1087/files#diff-cc45542b37e76ced3e366e4aa20f512cR90) method.

## References
- Issues: 
　https://github.com/screwdriver-cd/screwdriver/issues/998
　https://github.com/screwdriver-cd/screwdriver/issues/1030
　https://github.com/screwdriver-cd/screwdriver/issues/1075
- Related: (other PRs)
　https://github.com/screwdriver-cd/screwdriver/pull/1087
　https://github.com/screwdriver-cd/models/pull/245
　https://github.com/screwdriver-cd/executor-base/pull/42
　https://github.com/screwdriver-cd/executor-k8s/pull/84
　https://github.com/screwdriver-cd/executor-k8s-vm/pull/33
　https://github.com/screwdriver-cd/executor-docker/pull/25
　https://github.com/screwdriver-cd/executor-jenkins/pull/20


plugins/builds/token.js
```diff
@@ -0,0 +1,60 @@
...
+            return buildFactory.get(request.params.id).then((build) => {
+                if (!build) {
+                    throw boom.notFound('Build does not exist');
+                }
+
+                if (parseInt(request.params.id, 10) !== parseInt(profile.username, 10)) {
+                    throw boom.notFound('Build Id parameter and token does not match');
+                }
+
+                if (build.status !== 'QUEUED') {
```
jithin1987: We should keep an eye onthis in case we introduce new status for blocked or frozen jobs.

plugins/builds/token.js
```diff
@@ -0,0 +1,60 @@
...
+                    request.server.plugins.auth.generateProfile(
+                        profile.username,
+                        profile.scmContext,
+                        ['build'],
+                        {
+                            isPR: profile.isPR,
+                            jobId: profile.jobId,
+                            pipelineId: profile.pipelineId
+                        }
+                    ), buildTimeout * 60 // in seconds
```
jithin1987: hopefully this is in minutes. or should we put the onus of executors to pass in the seconds value, rather than doing the calculation here ?

plugins/builds/token.js
```diff
@@ -0,0 +1,60 @@
...
+            scope: ['temporal']
+        },
+        plugins: {
+            'hapi-swagger': {
+                security: [{ token: [] }]
+            }
+        },
+        handler: (request, reply) => {
+            const profile = request.auth.credentials;
+            const buildTimeout = request.payload.buildTimeout;
```
jithin1987: can we also add a validation here on `buildTimeout` like it's there for executor-base https://github.com/screwdriver-cd/executor-base/pull/42/files#diff-168726dbe96b3ce427e7fedce31bb0bcR138 ?

plugins/builds/token.js
sakka2: I see, this should be in minutes. How about calculate in `generateToken()` function?
https://github.com/screwdriver-cd/screwdriver/pull/1087/files#diff-cc45542b37e76ced3e366e4aa20f512cR89

plugins/builds/token.js
sakka2: OK! I will add this.

sakka2: I added.

Issue #1082, opened 2018-05-23
SCM Branch-specific jobs


yuichi10:
## Context
The issue of #723 was closed but [SCM Branch-specific jobs workflow](https://github.com/screwdriver-cd/screwdriver/blob/master/design/workflow.md#scm-branch-specific-jobs) is not implemented yet.
The workflow will make ease to use git flow.

## Objective
Allow to specify branch to commit trigger.

## Related links
Documentation: https://github.com/screwdriver-cd/screwdriver/blob/master/design/workflow.md#scm-branch-specific-jobs
Related to: #723

yuichi10:
I have split up the work into these tasks for now:
- [ ] get pipelines by like search of scmURI then search pipeline which have ~commit:{branch} trigger.
- [ ] fix regex TRIGGER to allow regex characters at data-schema.
- [ ] add branch to data-schema of SCHEMA_HOOK.
- [ ] use edge.src as a regex and check the trigger does match the regex or not at workflow-parser.
- [ ] if startFrom is ~commit, search jobs by trigger of ~commit and ~commit:{branch name} at models.
- [ ] udpate workflow documentation

yuichi10:
To get pipelines which you start jobs, I will add function to scm repositories which get branches from scms api.

The flow is as below
1. get branches by the new function
2. create scmUris by using the branch list
3. search pipelines from the scmUris and you get the pipelines which is related to the repository you using
4. search these pipelines which have the trigger to start by using eventFactory.getJobsFromTrigger. 
5. Then you will get the pipelines which we have to start jobs.

jithin1987:
@yuichi10 What's the use case for branch specific workflows ? Can't user specify branch while creating pipeline ?

yuichi10:
I want to verify the specifications about SCM Branch-specific jobs.

Assuming the project which have two pipelines as below.
![piiax5b18eb55475038c7601b956c](https://user-images.githubusercontent.com/4645011/41087486-cd1a4228-6a77-11e8-8ca5-bbbb93b5539f.jpg)

When developer push to staging branch, the red jobs will triggered. (I think this is no problem)
![usetf5b18eb6bdf9db0c260d0a6fb](https://user-images.githubusercontent.com/4645011/41087550-f9404424-6a77-11e8-806a-0b495f8c534f.jpg)

And now when developer click start button on staging pipeline page, the red jobs will triggered now.
[...]

PR #1097, opened 2018-05-31
feat(1082): [11] trigger specific branch

yuichi10:
## Context
To implement SCM Branch-specific jobs.

## Objective
Create events which have a branch-specific trigger.
I implement this feature as this below
1. get branches from scm api.
2. create scmUris by the branches (without request branch)
3. get pipelines by these scmUris
4. check the pipelines have trigger which should start or not
5. get request-branch's pipeline
6. create events for the all pipelines

The reason not to search request-branch pipeline at step 3 is to make code simpler.
If search request-branch at step 3, sometime the branch is declined by the step 4.
We always need to start the branch pipeline, so I separate step 3 and step 5.

## References
Related to #1082


plugins/webhooks/index.js
```diff
@@ -310,6 +311,94 @@ function pullRequestEvent(pluginOptions, request, reply, parsed) {
...
+ * @param   {String}            branch          the branch which is committed
+ * @returns {Promise}                           Promise that resolves into triggered pipelines
+ */
+function triggerPipelines(pipelineFactory, scmConfig, branch) {
+    return pipelineFactory.scm.getBranchList(scmConfig)
+        .then((branches) => {
+            const scmUris = [];
+            const splitUrl = scmConfig.scmUri.split(':');
+
+            branches.forEach((b) => {
```
tk3fftk: How about using `map()` like https://github.com/screwdriver-cd/scm-github/pull/91#discussion_r192939228

plugins/webhooks/index.js
```diff
@@ -310,6 +311,94 @@ function pullRequestEvent(pluginOptions, request, reply, parsed) {
...
+ * @param   {pipelineFactory}   pipelineFactory the pipeline factory to get branch list
+ * @param   {Object}            scmConfig       has the token and scmUri to get branches
+ * @param   {String}            branch          the branch which is committed
+ * @returns {Promise}                           Promise that resolves into triggered pipelines
+ */
+function triggerPipelines(pipelineFactory, scmConfig, branch) {
+    return pipelineFactory.scm.getBranchList(scmConfig)
+        .then((branches) => {
+            const scmUris = [];
+            const splitUrl = scmConfig.scmUri.split(':');
```
tk3fftk: This should be `splitUri`.

plugins/webhooks/index.js
```diff
@@ -310,6 +311,94 @@ function pullRequestEvent(pluginOptions, request, reply, parsed) {
...
+ * @param {EventFactory}    eventFactory    to create event
+ * @param {Array}           pipelines       the pipelines to start events
+ * @param {String}          parsed          it have a information to create event
+ * @returns {Promise}                       Promise that resolves into events
+ */
+async function createEvents(eventFactory, pipelines, parsed) {
+    const { branch, sha, username, scmContext, changedFiles } = parsed;
+    const events = [];
+
+    for (let i = 0; i < pipelines.length; i += 1) {
```
tk3fftk: How about using `forEach`? `i` is used only pipelines index.

plugins/webhooks/index.js
```diff
@@ -323,47 +412,50 @@ function pushEvent(pluginOptions, request, reply, parsed) {
...
                 request.log(['webhook', hookId],
                     `Skipping since Pipeline ${fullCheckoutUrl} does not exist`);
 
+                return [];
+            }
+
+            return createEvents(eventFactory, pipelines, parsed);
+        })
+        .then((events) => {
+            if (events.length <= 0) {
```
tk3fftk: Is there a case that `events.length` is minus?

plugins/webhooks/index.js
yuichi10: I try to use forEach, but when I try to use await it failed. So I use basic for.
https://stackoverflow.com/questions/37576685/using-async-await-with-a-foreach-loop

plugins/webhooks/index.js
```diff
@@ -310,6 +311,94 @@ function pullRequestEvent(pluginOptions, request, reply, parsed) {
...
 }
 
+/**
+ * Check the pipeline have triggered job or not
+ * @method  isTrigger
+ * @param   {Pipeline}  pipeline    The pipeline to check
+ * @param   {String}    startFrom   The trigger name
+ * @returns {boolean}               If true the pipeline has triggered job
+ */
+function isTrigger(pipeline, startFrom) {
```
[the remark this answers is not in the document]
yuichi10: That would be nicer. I will use `hasTriggeredJob`!

plugins/webhooks/index.js
yuichi10: It is better to use map(). I will fix.

plugins/webhooks/index.js
yuichi10: I will fix.

plugins/webhooks/index.js
yuichi10: The length will never be minus, it's kind my habit.  Is it better to compare by `events.length === 0`?

plugins/webhooks/index.js
tk3fftk: In this page, the answerer said you should use `for … of`. How about it?

plugins/webhooks/index.js
tk3fftk: I think so.

yuichi10: Ok then I will fix.

plugins/webhooks/index.js
yuichi10: When I use `for ... of` , I got `no-restricted-syntax` lint error, thus I use basic `for` now. 
> error  iterators/generators require regenerator-runtime, which is too heavyweight for this guide to allow them. Separately, loops should be avoided in favor of array iterations  no-restricted-syntax

tk3fftk: oh..

plugins/webhooks/index.js
```diff
@@ -323,47 +411,51 @@ function pushEvent(pluginOptions, request, reply, parsed) {
...
-            return eventFactory.create(eventConfig)
-                .then((event) => {
-                    request.log(['webhook', hookId, event.id],
-                        `event ${event.id} started`);
-
-                    return reply().code(201);
-                });
+            events.forEach((e) => {
+                request.log(['webhook', hookId, e.id],
+                    `event ${e.id} started`);
```
Filbird: Capitalize `event`

plugins/webhooks/index.js
```diff
@@ -310,6 +311,93 @@ function pullRequestEvent(pluginOptions, request, reply, parsed) {
...
+        trigger: startFrom
+    });
+
+    return nextJobs.length > 0;
+}
+
+/**
+ * Get all pipelines which has triggered job
+ * @method  triggerPipelines
+ * @param   {pipelineFactory}   pipelineFactory the pipeline factory to get branch list
```
Filbird: Capitalize the JSDoc descriptions (`The pipeline factory...`)

plugins/webhooks/index.js
```diff
@@ -310,6 +311,93 @@ function pullRequestEvent(pluginOptions, request, reply, parsed) {
...
+ * @param   {Object}            scmConfig       has the token and scmUri to get branches
+ * @param   {String}            branch          the branch which is committed
+ * @returns {Promise}                           Promise that resolves into triggered pipelines
+ */
+function triggerPipelines(pipelineFactory, scmConfig, branch) {
```
Filbird: Maybe `triggeredPipelines` works better since we return an array of pipelines.

plugins/webhooks/index.js
```diff
@@ -310,6 +311,93 @@ function pullRequestEvent(pluginOptions, request, reply, parsed) {
         .catch(err => reply(boom.wrap(err)));
 }
 
+/**
+ * Check the pipeline have triggered job or not
```
Filbird: Check if the pipeline has a triggered job or not

plugins/webhooks/index.js
yuichi10: I will fix.

plugins/webhooks/index.js
yuichi10: Oops I will fix.

plugins/webhooks/index.js
yuichi10: Yes, `triggeredPipelines` looks better, I will fix

plugins/webhooks/index.js
yuichi10: Thank you. I will fix

yuichi10:
@Filbird Thank you your comments! I fixed all.

plugins/webhooks/index.js
```diff
@@ -310,6 +311,93 @@ function pullRequestEvent(pluginOptions, request, reply, parsed) {
         .catch(err => reply(boom.wrap(err)));
 }
 
+/**
+ * Check if the pipeline has a triggered job or not
+ * @method  hasTriggeredJob
+ * @param   {Pipeline}  pipeline    The pipeline to check
+ * @param   {String}    startFrom   The trigger name
+ * @returns {boolean}               If true the pipeline has triggered job
```
tkyi: Boolean

plugins/webhooks/index.js
```diff
@@ -310,6 +311,93 @@ function pullRequestEvent(pluginOptions, request, reply, parsed) {
...
+        trigger: startFrom
+    });
+
+    return nextJobs.length > 0;
+}
+
+/**
+ * Get all pipelines which has triggered job
+ * @method  triggeredPipelines
+ * @param   {pipelineFactory}   pipelineFactory The pipeline factory to get branch list
```
tkyi: PipelineFactory

plugins/webhooks/index.js
```diff
@@ -310,6 +311,93 @@ function pullRequestEvent(pluginOptions, request, reply, parsed) {
...
+            return branches.filter(b => b.name !== branch).map((b) => {
+                splitUri[2] = b.name;
+
+                return splitUri.join(':');
+            });
+        })
+        .then(scmUris => pipelineFactory.list({ params: { scmUri: scmUris } }))
+        .then(pipelines => pipelines.filter(p => hasTriggeredJob(p, `~commit:${branch}`)))
+        .then((pipelines) => {
+            const scmUri = scmConfig.scmUri;
```
tkyi: probably don't need to declare this variable here if it is only used once

yuichi10: You are right, I fixed.

plugins/webhooks/index.js
tk3fftk: {pipelineFactory}  -> {**P**ipelineFactory}

Issue #796, opened 2017-11-14
Display past builds for PR job


jithin1987:
# Context

If user makes incremental changes to a PR, it kicks off new builds. But UI shows only last build, there by loosing context of past builds.

# Objective

Display in PR build page, past builds for the PR job.

Design mocks:
![05a_jobs_dropdown_more](https://user-images.githubusercontent.com/193126/40160347-23cbb970-5962-11e8-982f-d6685d2bcf04.png)
![05a_jobs_dropdown](https://user-images.githubusercontent.com/193126/40160348-23e56c3a-5962-11e8-8be1-5f7519ed384c.png)
![05a_jobs](https://user-images.githubusercontent.com/193126/40160349-2419a072-5962-11e8-96fb-f69f0bf847b7.png)

jithin1987:
_5/16/2018_
Updated with latest mocks.

lusol:
@pranavrc don't do this one please

tkyi:
Functionality is there, though CSS could be cleaned up:
<img width="720" alt="screen shot 2018-06-14 at 10 46 06 am" src="https://user-images.githubusercontent.com/3230529/41428869-3a342fde-6fc0-11e8-865f-5bcd54c95436.png">

PR #1098, opened 2018-05-31
feat(796): add possibility of pr to type property of event objects

lusol:
## Context

so that we could identify what type of event we have, whether it's pipeline or pr

## Objective

to allow us to have prs in the type property of event objects

## References

https://github.com/screwdriver-cd/screwdriver/issues/796

### DIFF
--- a/plugins/pipelines/listEvents.js
+++ b/plugins/pipelines/listEvents.js
@@ -23,20 +23,20 @@ module.exports = () => ({
         },
         handler: (request, reply) => {
             const factory = request.server.app.pipelineFactory;
-            
+
             return factory.get(request.params.id)
                 .then((pipeline) => {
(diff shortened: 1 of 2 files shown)


plugins/pipelines/listEvents.js
```diff
@@ -23,14 +23,20 @@ module.exports = () => ({
         },
         handler: (request, reply) => {
             const factory = request.server.app.pipelineFactory;
-
+
```
Filbird: Unsure but is this extra whitespace?

lusol: lint requires blank line after variable declaration

tkyi: I think Filbert is referring to the extra spaces in that line

Issue #937, opened 2018-03-20
External screwdriver config


jithin1987:
## Context

Allow pipelines to build source code in external SCM repository. This feature can be used to build applications who do not want to bundle `screwdriver.yaml` file in the SCM repository of their source code. 

### Use cases
1. Projects which tracks Opensource code. Such projects would want to keep their SCM repository which contains OpenSource code without extra files and host the build files in a separate SCM repository which is used solely for building the application.
1. A build project used solely for building multiple external SCM repositories. In this instance any commit on the external repository should trigger an event in the build project. Also user should have flexibility to build all external projects from the build project.

## Objective
1. Allow user to specify external repositories in job config.
1. UI should provide a drop down list of configured SCM's before Pipeline header in `Events` tab
   1.  <img width="959" alt="screen shot 2018-05-11 at 8 35 47 am" src="https://user-images.githubusercontent.com/193126/39932973-83a3dcb2-54f6-11e8-8c74-ac634afdd056.png">

1. Selecting a repository from the drop down list should show `events` and `PR` from the selected repository, exactly like existing pipeline view for non external project pipeline of today.
1. Any commits or PRs to external repository should trigger commit & PR builds in the build repository which user can see once user selects the external repository from drop down list.
1. Clicking on `Start All` (Refer screenshot) should build all external repositories.
1. Any commits to build project should build all external repositories.
   1. We should define what does it mean for a PR build in the build repository.
1. `SD_SOURCE_DIR` should point to the checked out code for the external repository.

## References

* External Repository - SCM Repository where the source code lives which is not the same repository where build code & `screwdriver.yaml` lives.

dragon512:
It would also be nice to be able to define the type of change... ie commit vs tag as an example.

Other systems define this logic as Buildbot, bitbucket pipeline (https://confluence.atlassian.com/bitbucket/configure-bitbucket-pipelines-yml-792298910.html) and others...

jithin1987:
@dragon512 This is similar to https://github.com/screwdriver-cd/screwdriver/issues/823 will support in a generic way outside of this feature.

minz1027:
## Proposed implementation

### Example config
```yaml
jobs:
    build:
      scm_list: 
          - git@github.com:screwdriver-cd/screwdriver.git
          - git@github.com:screwdriver-cd/ui.git
      requires: [~commit, ~pr]
      image: node:8
      steps:
         - install: npm instal
         - test: npm test

    publish: 
      requires: build
[...]

PR #1109, opened 2018-06-07
fix(937): add start all child pipelines endpoint

minz1027:
- add start all child pipelines endpoint
- allow to list pipelines by `configPipelineId`

Related to: https://github.com/screwdriver-cd/screwdriver/issues/937

### DIFF
--- a/plugins/pipelines/startAll.js
+++ b/plugins/pipelines/startAll.js
@@ -22,12 +22,9 @@ module.exports = () => ({
             }
         },
         handler: (request, reply) => {
-            const pipelineFactory = request.server.app.pipelineFactory;
-            const eventFactory = request.server.app.eventFactory;
-            const userFactory = request.server.app.userFactory;
+            const { pipelineFactory, eventFactory, userFactory } = request.server.app;
+            const { username, scmContext } = request.auth.credentials;
             const id = request.params.id;
-            const username = request.auth.credentials.username;
-            const scmContext = request.auth.credentials.scmContext;
             const scm = pipelineFactory.scm;
 
             return Promise.all([
--- a/plugins/pipelines/startAll.js
+++ b/plugins/pipelines/startAll.js
@@ -1,0 +1,79 @@
+'use strict';
+
+const boom = require('boom');
+const joi = require('joi');
+const schema = require('screwdriver-data-schema');
+const idSchema = joi.reach(schema.models.pipeline.base, 'id');
+
+module.exports = () => ({
+    method: 'POST',
+    path: '/pipelines/{id}/startall',
+    config: {
+        description: 'Start all child pipelines given a specific pipeline',
+        notes: 'Start all child pipelines given a specific pipeline',
+        tags: ['api', 'pipelines'],
+        auth: {
+            strategies: ['token'],
+            scope: ['user', '!guest']
+        },
+        plugins: {
+            'hapi-swagger': {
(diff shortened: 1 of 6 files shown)


plugins/pipelines/startAll.js
```diff
@@ -0,0 +1,79 @@
+'use strict';
+
+const boom = require('boom');
+const joi = require('joi');
+const schema = require('screwdriver-data-schema');
+const idSchema = joi.reach(schema.models.pipeline.base, 'id');
+
+module.exports = () => ({
+    method: 'POST',
+    path: '/pipelines/{id}/startall',
```
tkyi: camel case maybe? `startAll`?

minz1027: don't think it's good to have camel case in url...

tkyi: good point whoops

plugins/pipelines/startAll.js
```diff
@@ -0,0 +1,79 @@
...
+            strategies: ['token'],
+            scope: ['user', '!guest']
+        },
+        plugins: {
+            'hapi-swagger': {
+                security: [{ token: [] }]
+            }
+        },
+        handler: (request, reply) => {
+            const pipelineFactory = request.server.app.pipelineFactory;
```
tkyi: can do `const { pipelineFactory, eventFactory, userFactory } = request.server.app;`

plugins/pipelines/startAll.js
```diff
@@ -0,0 +1,79 @@
...
+            const id = request.params.id;
+            const username = request.auth.credentials.username;
+            const scmContext = request.auth.credentials.scmContext;
+            const scm = pipelineFactory.scm;
+
+            return Promise.all([
+                pipelineFactory.get(id),
+                userFactory.get({ username, scmContext })
+            ])
+                .then(([pipeline, user]) => {
```
tkyi: do we need to check if user exists?
```
if (!user) {
     throw boom.notFound(`User ${username} does not exist`);
}
```

plugins/pipelines/startAll.js
```diff
@@ -0,0 +1,79 @@
...
+                    if (!pipeline) {
+                        throw boom.notFound('Pipeline does not exist');
+                    }
+
+                    return user.getPermissions(pipeline.scmUri)
+                        // check if user has push access
+                        .then((permissions) => {
```
tkyi: Are we going to check if the user has permission on the child pipelines? Or is it implied?

minz1027: Don't think we need to check here, it is shared permissions.  e.g. I have permissions for child A, B, you have permission for child C and added that. All child A, B, C should start.

plugins/pipelines/startAll.js
minz1027: We have that in a lot of our endpoints, but i forgot why we need it.. We're getting `username` and `scmContext` from user's JWT, how is it possible that the user doesn't exists?

tkyi: I see what you're saying. Should be okay to not have this check then.

plugins/pipelines/startAll.js
tkyi: I see, so for startall we only care if the user has permissions against the config pipeline. If we did a Start on a childPipeline, we would still check permissions against that child repo.

Issue #614, opened 2017-06-19
Pipeline API tokens


petey:
Similar to User API tokens, but scoped to a specific pipeline.

kumada626:
## Proposal
We designed Pipeline API tokens as below.

Rules:
- Tokens should be specifically tied to the Pipeline that issued them
- Tokens should have a label/name associated with them
- No one can ever look up the raw value after initially generating the token

Users who are admins of the pipeline that issued token can below.
- Users can add a description to Pipeline tokens, so they know what they're using them for
- Users can revoke a token at any time
- User can see when a token was last used

Difference from User API token:
- scope: the Pipeline API tokens only interact with Screwdriver API which is related to the Pipeline.
- The Pipeline tokens can be created, regenerated, edited and deleted by the members who can edit its pipeline's secrets.

UI:
The Pipeline tokens can be generated, listed and deleted from the Pipeline's secret tab like below:

<img width="1116" alt="2018-05-23 14 51 39" src="https://user-images.githubusercontent.com/8113204/40406145-81434374-5e9a-11e8-8d5e-27551c0c7dcf.png">

or, from an another tab for token.
Flows of creating, regenerating, editing, and deleting pipeline tokens from the UI are the same as User API tokens.

I think the Pipeline tokens should have pipelineId to their data-schema, so we have to create new model for the Pipeline tokens or fix token model.

Please let me know if any points are missing.

yuichi10:
How to store this pipeline token data?

kumada626:
I think it should be the same way of the User API tokens are stored.

lgfausak:
I seems tricky to figure out exactly which api calls could be used.  I imagine, for example, that the ability to create an event on that pipeline would be included?  Will the pipeline token have a different swagger (/v4/documentation) to be explicit about what can be done?

kumada626:
I assume that the ability to create an event on that pipeline would be included. Like the Build token, we'll add `pipeline` to the api scope to restrict the ability of the Pipeline tokens.
https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/builds/get.js#L18
Currently, there is no different swagger for the Build tokens. Therefore, I think that we don't have to prepare another swagger for the Pipeline tokens.

tk3fftk:
What is the API endpoints which generate a pipeline token and obtain JWT based on a pipeline token?

kumada626:
[...]

PR #1123, opened 2018-06-14
feat(614): [3] Add pipeline token

s-yoshika:
## Context
To implement pipeline API token.

## Objective
This PR add new API endpoints for pipeline API token, and add pipeline scope to some API related to pipelines. pipeline scope is new scope to allow access by pipeline API tokens.

This PR add pipeline scope to a few API related to a specific pipeline like below.
```
GET   /pipelines/{id}
GET   /pipelines/{id}/events
GET   /pipelines/{id}/jobs
PUT   /pipelines/{id}
POST  /pipelines/{id}/sync
POST  /builds
```
It seems better to add more APIs, but it is difficult to grant permission to the pipeline API token.
 I think we can allow pipeline API tokens to access GET API associated with a specific pipeline like `/builds/{id},/events/{id} etc`.

Please tell me if there is an opinion.

## References
Issue and design :
https://github.com/screwdriver-cd/screwdriver/issues/614

Related PRs:
https://github.com/screwdriver-cd/data-schema/pull/257
https://github.com/screwdriver-cd/models/pull/264


tk3fftk:
We should also update [README.md](https://github.com/sonic-screwdriver-cd/screwdriver/blob/pipeline-token/plugins/pipelines/README.md) or create new README on ` plugins/pipelines/tokens/`

plugins/auth/index.js
```diff
@@ -140,30 +140,58 @@ exports.register = (server, options, next) => {
...
-                            const profile = {
-                                username: user.username,
-                                scmContext: user.scmContext,
-                                scope: ['user']
-                            };
-
+                            return Promise.reject();
+                        })
+                        .then((profile) => {
+                            request.log(['auth'], `${profile.username} has logged in via API keys`);
```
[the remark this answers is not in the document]
minz1027: To add to @tk3fftk's comment, can we also log the scope so that we know it is from a user or a pipeline token?

plugins/auth/index.js
```diff
@@ -140,30 +140,58 @@ exports.register = (server, options, next) => {
...
-
-                    return factory.get({ accessToken: token })
-                        .then((user) => {
-                            if (!user) {
-                                return cb(null, false, {});
+                    const tokenFactory = request.server.app.tokenFactory;
+                    const userFactory = request.server.app.userFactory;
+                    const pipelineFactory = request.server.app.pipelineFactory;
+
+                    tokenFactory.get({ value: tokenValue })
```
minz1027: Can we have `return` before this? just to be consistent

plugins/auth/index.js
```diff
@@ -140,30 +140,58 @@ exports.register = (server, options, next) => {
...
+                            return Promise.reject();
+                        })
+                        .then((profile) => {
+                            request.log(['auth'], `${profile.username} has logged in via API keys`);
                             profile.token = server.plugins.auth.generateToken(profile);
 
                             return cb(null, true, profile);
-                        });
+                        })
+                        .catch(() => cb(null, false, {}));
```
minz1027: For the purpose of debugging, shall we cat the err and log it out?

plugins/auth/index.js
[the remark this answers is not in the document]
s-yoshika: It's confusing, I'll fix it.

plugins/auth/index.js
s-yoshika: Okay, I'll fix it.

plugins/auth/index.js
s-yoshika: Sure, I'll fix it.

s-yoshika:
> We should also update README.md or create new README on plugins/pipelines/tokens/

Thanks!

> Do we need refresh like user token?
https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/tokens/refresh.js

Yes, the proposal says it. I forgot it. Thanks!
https://github.com/screwdriver-cd/screwdriver/issues/614#issuecomment-391232568

s-yoshika:
@tk3fftk 
> I think we should include POST /events to the scope. Most of builds are created via this endpoint.

Thank you for your opinion. it is good to allow pipeline token to create events.
I added pipeline scope to `POST /events` too.

s-yoshika:
@minz1027 Thank you for your suggestion. The [issue description](https://github.com/screwdriver-cd/screwdriver/issues/614) says pipeline API token is scoped to a specific pipeline, so I added check on pipeline ID even for GET API.

If there is no problem, I'll remove the pipeline check in the case of GET endpoints.

plugins/pipelines/tokens/refresh.js
```diff
@@ -0,0 +1,75 @@
...
+const tokenIdSchema = joi.reach(schema.models.token.base, 'id');
+const pipelineIdSchema = joi.reach(schema.models.pipeline.base, 'id');
+
+module.exports = () => ({
+    method: 'PUT',
+    path: '/pipelines/{pipelineId}/tokens/{tokenId}/refresh',
+    config: {
+        description: 'Refresh a pipeline token',
+        notes: 'Update the value of a token while preserving its other metadata',
+        tags: ['api', 'secrets'],
```
sakka2: should be tokens?

plugins/tokens/update.js
```diff
@@ -10,8 +10,8 @@ module.exports = () => ({
     path: '/tokens/{id}',
     config: {
         description: 'Update a token',
-        notes: 'Update a specific tooken',
```
tkyi: 👍 😆

plugins/pipelines/index.js
```diff
@@ -22,6 +28,20 @@ const startAllRoute = require('./startAll');
...
+    /**
+     * Returns true if the scope includes pipeline and it's pipelineId does not matches
+     * the id of the pipeline, otherwise returns false
+     * @method isInvalidToken
+     * @param  {String} id                     ID of pipeline
+     * @param  {Object} credentials            Credential object from Hapi
+     * @param  {String} credentials.pipelineId ID of pipeline which the token is allowed to access
+     * @param  {String} credentials.scope      Scope whose token is allowed
+     */
+    server.expose('isInvalidToken', (id, credentials) =>
```
tkyi: I think this would be more straightforward/clear if we changed the name to `isValidPipelineToken` and then have it return `credentials.scope.includes('pipeline') && parseInt(id, 10) === parseInt(credentials.pipelineId, 10)`

plugins/pipelines/tokens/create.js
```diff
@@ -0,0 +1,86 @@
...
+module.exports = () => ({
+    method: 'POST',
+    path: '/pipelines/{id}/tokens',
+    config: {
+        description: 'Create a new token for pipeline',
+        notes: 'Create a specific token for pipeline',
+        tags: ['api', 'tokens'],
+        auth: {
+            strategies: ['token'],
+            scope: ['user', '!guest']
```
tkyi: Wondering if we should allow pipeline tokens to create/update/delete/etc other pipeline tokens, but I think we can revisit that issue later. This looks fine for now.

plugins/pipelines/index.js
s-yoshika: Thanks, it's more clear. I'll fix.

plugins/pipelines/tokens/create.js
jithin1987: I think it's better for admins of the pipeline to have control over the management of the pipeline tokens

plugins/pipelines/index.js
s-yoshika: If non-pipeline scope is passed,  `isValidPipelineToken` returns false and [this](https://github.com/sonic-screwdriver-cd/screwdriver/blob/f54382028b0a1eedd066bc79947733c8177fff59/plugins/builds/create.js#L47-L48) thows unauthorize error. We can check the scope is not `pipeline` in if statement before calling `isValidPipelineToken`, but it's a 
 little redundant.

So, I change the name to `isValidToken` which returns true when the scope is not pipeline or the scope is pipeline and the token's `pipelineId` matches the id of pipeline.

s-yoshika:
@minz1027 @jithin1987 Thanks, I removed the check for GET APIs.

>  All POST/DELETE/PUT APi's should continue to enforce the check ?

I think they should check pipeline ID. In pipeline tokens, we use admin username as `credential.username`. Write APis check the permission for scm with `credential.username`.
So if pipeline token can also access Write APis of other pipelines, it may give unintended permissions to users who use that token.

plugins/pipelines/tokens/create.js
tkyi: Well right now user tokens can generate more user tokens, so allowing pipeline tokens to create pipeline tokens would be more consistent with what we have.

plugins/pipelines/tokens/create.js
```diff
@@ -0,0 +1,86 @@
...
+                ])
+                    .then(() => tokenFactory.create({
+                        name: request.payload.name,
+                        description: request.payload.description,
+                        pipelineId
+                    }))
+                    .then((token) => {
+                        const location = urlLib.format({
+                            host: request.headers.host,
+                            port: 100,
```
tkyi: Why is this port 100? Should it be `request.headers.port` like in https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/pipelines/create.js#L80?

s-yoshika: Thanks! It's terrible mistake. I fixed.

plugins/pipelines/tokens/create.js
jithin1987: That is logged in users accessing API via UI, If we are not differentiating between generated user access tokens and logged in users then it's a design bug.

Issue #1122, opened 2018-06-14
Introduce `Unstable` build status


jithin1987:
## Context/Objective

Give users an option to mark their build as unstable, rather than going for outright failure on certain step failures. Onus is own user to make sure their step passes and call API to mark build as `Unstable` on certain conditions. For example a user would want their build to complete all steps even after test failures and at same time have the build marked as `Unstable` if certain conditions are not met.

## Implementation
- [x] [data-schema] Allow `Unstable` status for build.
- [x] [screwdriver] Provide endpoint for build to mark its status as `Unstable`. (Via API Call)
- [x] [UI] Display `Unstable` builds with appropriate build status Icon.
- [x] [models] If a build's status is Unstable don't change it to `Success` at end of build.

d2lam:
This should be done. Here is an example of a pipeline updating unstable build: https://cd.screwdriver.cd/pipelines/1092/events

tkyi:
Related to https://github.com/screwdriver-cd/screwdriver/issues/1029

drewfish:
Can we use markdown in the `statusMessage`?

Issue #653, opened 2017-07-20
Blocked By


petey:
## Context

Build execution start should be able to be blocked by execution of another build, such that one build should not start until the other is completed.

## Objective

- [x] Jobs are always blocked by themselves. A build for any given job should not run concurrently with themself.
- [x] Jobs can be configured to be blocked by other jobs in the same pipeline

- [x] Jobs can be configured to be blocked by jobs in other pipelines.

- [x] UI should show a blocked task in status Blocked with reason. Reason should link to the running job which is blocking current job.

d2lam:
# Proposed implementations
Using an example: B is blocked by A

## First way:
- Pass in the `blocked by` field to `executor start`
- Create a new plugin like [JobLock](https://github.com/taskrabbit/node-resque/blob/master/lib/plugins/JobLock.js#L4-L5) and modify `beforePerform`. 
  - Explanation: this function is trying to [SETNX](https://redis.io/commands/setnx) the `key`. If set is not successful, that means it's already in the queue, so re-enqueue. In our case, we use `buildId` as args, which is always unique, so it will never re-enqueue. 
  - Beside, if we use the queue to check, that means we will re-enqueue B if A is currently **IN QUEUE**. Please correct me if I'm wrong, I don't think that's what we want. We want to re-enqueue B if A is currently **RUNNING**.  
- So I think we need to keep track of all `jobId` that are running in a table called `runningJobs`. In `beforePerform`, we check if **the blocking job** is running by checking if the `jobId` is in the table. If it is then re-enqueue. If not then we add the **current** `jobId` to table, same for regular case that is not blocked_by. In `afterPerform`, we remove the **current** `jobId`.
- Inside the worker `start`, we will add this new plugin like [this example](https://github.com/taskrabbit/node-resque/blob/1b8509651bf5962b987cce2de819cd1d2c9576b7/examples/example.js#L27-L32)

**TO DO:**
- Data-schema: accept `blocked_by`
- Config-parser: parse `blocked_by`
- Models: pass `blocked_by` to executor `start`
- Executor-queue: make sure the field is passed to worker job `start`
- Worker: implement the new plugin and add it to `start`

**NOTES:** 
This implementation relies on redis and queue, just like `build periodically`.

## Second way (simpler):
[...]

PR #1129, opened 2018-06-19
Support block & unstable statuses

d2lam:
Support `BLOCKED` and `UNSTABLE` status. 
Previously, we only allowed update `RUNNING` or `QUEUED`, but we want to allow updating `BLOCKED` and `UNSTABLE`. 
They should skip on updating `meta` and `endTime`.
`BLOCKED` -> `SUCCESS` is ok, but `UNSTABLE` is not allowed to be updated to something else. 

Related:
https://github.com/screwdriver-cd/screwdriver/issues/653
https://github.com/screwdriver-cd/screwdriver/issues/1122


plugins/builds/update.js
```diff
@@ -96,8 +104,9 @@ module.exports = () => ({
                         buildLink:
                             `${buildFactory.uiUri}/pipelines/${pipeline.id}/builds/${id}`
                     });
-                    // Guard against triggering non-successful builds
-                    if (desiredStatus !== 'SUCCESS') {
+
+                    // Guard against triggering non-successful or unstable builds
+                    if (desiredStatus !== 'SUCCESS' || build.status === 'UNSTABLE') {
```
jithin1987: won't `if (build.status !== 'SUCCESS')` do ?

d2lam: `build.status` is current status. `desiredStatus` is the new status

We only want to trigger if the new status is `SUCCESS`, and the current status is not `UNSTABLE`.

d2lam:
@tkyi  I implemented it based on the acceptance criteria `If a build's status is Unstable don't change it to Success at end of build.` and it shouldn't trigger next build according to @jithin1987 

I think the intention is user can update a build to `UNSTABLE` if they think something is not right. For example, coverage is too low. Therefore, it should probably not trigger the next build, and should not be reupdated to `SUCCESS` at the end of build.

plugins/builds/update.js
d2lam: Oh I see at this point `build.status` should be the new status already. We can use that check instead like you said.

jithin1987:
> I think I'm still confused what UNSTABLE is used for. Is it just an arbitrary label put on a build instead of SUCCESS or FAILURE? Should UNSTABLE builds still trigger the next build?

UNSTABLE builds should not trigger next builds. It's a build status which build can call to report itself as UNSTABLE, we don't control the logic on what makes a build UNSTABLE.

plugins/builds/update.js
```diff
@@ -74,14 +75,21 @@ module.exports = () => ({
...
 
-                    // Everyone is able to update the status
-                    build.status = desiredStatus;
-                    if (statusMessage) {
-                        build.statusMessage = statusMessage;
+                    // UNSTABLE -> SUCCESS needs to update meta and endtime.
+                    // However, the status itself cannot be updated to SUCCESS
+                    if (build.status !== 'UNSTABLE') {
+                        build.status = desiredStatus;
+                        if (statusMessage) {
```
jithin1987: should `statusMessage` be restricted for UNSTABLE ?

d2lam: I don't think so. Moved it out 👍

PR #1133, opened 2018-06-20
fix: change log lines per file to 1000

minz1027:
We changed the lines per file in log-service to 1000 lines and it ships with the latest launcher version `v4.0.112`
https://github.com/screwdriver-cd/launcher/commit/37f90c42d82210cff40dd12399f96d55508ee522

This change need to be pulled in before pulling in launcher verson `v4.0.112` otherwise there will be missing logs in UI.


minz1027:
Close in favor of this one https://github.com/screwdriver-cd/screwdriver/pull/1134

PR #1134, opened 2018-06-20
fix: support log lines per file to 1000

minz1027:
We changed the lines per file in log-service to `1000` lines and it ships with the latest launcher version `v4.0.112`
https://github.com/screwdriver-cd/launcher/commit/37f90c42d82210cff40dd12399f96d55508ee522

This change need to be pulled in before pulling in launcher verson `v4.0.112` otherwise there will be missing logs in UI.

This PR is backward compatible. It will work with both `100` lines and `1000` lines.


plugins/builds/steps/logs.js
```diff
@@ -4,7 +4,7 @@ const boom = require('boom');
 const schema = require('screwdriver-data-schema');
 const request = require('request');
 const ndjson = require('ndjson');
-const MAX_LINES = 100;
+let MAX_LINES = 100;
```
d2lam: let's make this lower case since it's not a `const` anymore.

PR #1136, opened 2018-06-21
feat(677): add GET /commands/namespace/name/tags

tk3fftk:
## Context
API does not have the endpoint of getting all command tags with namespace and name yet.
So UI does cannot show command versions with tags. 

## Objective
Add `GET /commands/namespace/name/tags` endpoint.
This PR also add auth to command endpoints. I think these endpoints should also have auth strategy same as other endpoints.

## References
Related: #677
BlockedBy: https://github.com/screwdriver-cd/ui/pull/318


plugins/commands/listTags.js
```diff
@@ -0,0 +1,54 @@
+'use strict';
+
+const boom = require('boom');
+const joi = require('joi');
+const schema = require('screwdriver-data-schema');
+const listSchema = joi.array().items(schema.models.commandTag.base).label('List of command tags');
+const nameSpaceSchema = joi.reach(schema.models.commandTag.base, 'namespace');
```
tkyi: namespaceSchema*

tk3fftk: fixed!

PR #1138, opened 2018-06-21
fix: get user by access token

minz1027:
Functional tests is broken with the last PR merged saying that `lastUsed` of accessToken is not updated: 
https://cd.screwdriver.cd/pipelines/1/builds/44534
PR: https://github.com/screwdriver-cd/screwdriver/pull/1123

Because in that PR, it switch from `userFactory.get({ accessToken: xxx })` to `userFactory.get(token.userId)`. But inside the model, we only update t`oken.lastUsed` field when you're getting the user with accessToken.

### DIFF
--- a/plugins/auth/index.js
+++ b/plugins/auth/index.js
@@ -154,7 +154,7 @@ exports.register = (server, options, next) => {
                                 return Promise.reject();
                             } else if (token.userId) {
                                 // if token has userId then the token is for user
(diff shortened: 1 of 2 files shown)


s-yoshika:
Sorry to break functional-test :bow: I overlooked about lastUsed, I will fix it for pipeline token as well.

PR #1144, opened 2018-06-22
fix: update status message

d2lam:
Update status message


plugins/builds/update.js
```diff
@@ -89,9 +89,7 @@ module.exports = () => ({
                         build.status = desiredStatus;
                     }
 
-                    if (statusMessage) {
-                        build.statusMessage = statusMessage;
-                    }
+                    build.statusMessage = statusMessage || '';
```
jithin1987: should have a case for `USTABLE` builds, else SUCCESS ones will clear it

Issue #357, opened 2016-11-11
What is "Build Cache" ?


ginoh:
Hi, I have a question.

Has Cache depicted in the Overall Archtecture been already implemented ?

<img width="282" alt="2016-11-11 11 21 13" src="https://cloud.githubusercontent.com/assets/2827530/20203250/1c0c6cfe-a809-11e6-987b-4ea771d9a910.png">

jpcollins:
Not yet.  It's dependent on execution engine and storage.  In Yahoo, we are exploring using PersistentVolumes in Kubernetes using NFS as storage backend right now.

ginoh:
Thank you. I understand it.
I hope that the document would be updated if there is progress.

PR #1160, opened 2018-06-29
Build cache proposal

elyzion:
## Context

Users require a way to cache resources between jobs within a single pipeline or consecutive executions of a particular pipeline in order to speed up build time. This PR includes a proposal for such system.

## Objective

Review design proposal for build cache.

## References

#357


yahoocla:
Thank you for submitting this pull request, however I do not see a valid CLA on file for you. Before we can merge this request please visit https://yahoocla.herokuapp.com/ and agree to the terms. Thanks! :smile:

jithin1987:
cc @Zhongtang @catto

design/build-cache.md
```diff
@@ -0,0 +1,201 @@
...
+Caching of potentially sensitive data also needs to be handled. Data provided via the secrets field should not be cached. 
+
+### Cache invalidation
+
+- [Proposed] Changing environment variables or secrets should invalidate existing caches. Mass invalidation can be avoided by including a hash of variables and secrets with the cache metadata.
+- [Proposed] The user should also be able to invalidate caches via the UI.
+
+### Conflicts
+
+Multiple caches trying to restore the same file/files will result in an error. Caches by default do not overwrite existing files or directories. This behaviour can be configured by setting selecting a merging strategy.
```
Zhongtang: When will `Multiple caches trying to restore the same file/files` happen? Is this referring to multiple builds accessing the same cache directory at the same time?

design/build-cache.md
```diff
@@ -0,0 +1,201 @@
...
+1. Build dependencies take a long time to resolve
+2. Generated resources are meant for deployment (Risk prone, not recommended unless deleting deployed artifact from cache after deployment)
+
+### (Proposal) Local and remote caching
+
+Pipeline and event level caches are stored on the build cache service node. They are copied to the build node on pipeline setup, and back to the build cache on pipeline completion.
+
+Once copied to the build node, they are persisted as a shared volume, mounted onto the appropriate path in the build container. 
+
+This proposal is currently blocked by the nature of the backend implementation for nodes. The k8s executor for example, selects a completely random node for every job in the pipeline. Caching locally is therefore not viable at the moment.
```
Zhongtang: One possibility is to use NFS. With k8s one can setup nfs backed Persistent Volume Claim.

design/build-cache.md
```diff
@@ -0,0 +1,201 @@
...
+
+### Build cache service
+
+The build cache service is essentially a specialized version of the store service. It utilizes the builtin Hapi cache system backed by Catbox.
```
Zhongtang: Would a network cache service be too slow in some cases? For example npm cache essentially uses local disk to cache network I/O. But if the entire npm cache has to be loaded through network when the build starts, doesn't it defeat the purpose? I can see gains from using tar to reduce the # of network calls, or from talking to a service in the same network instead of a remote NPM register, but the gains are probably not as significant as using disk volumes, and could be offset by the added size of the files that need to be loaded. We probably need to do some speed testing before going with this choice.

On the other hand, for cases where things have to be re-compiled (for example builds using source code dependencies like cocoapods for iOS or REPO for Android), a network cache service totally makes sense.

design/build-cache.md
```diff
@@ -0,0 +1,201 @@
...
+The following merging strategies are defined for dealing with file conflicts when extracting cached data on ta local build node.
+
+1. merge: Compares creation time of files and directories. Recursively merges files and directories from multiple caches. Selects the *newest* version of a file.
+2. [Proposed] preserve: Compares creation time of files and directories. Recursively merges files and directories from multiple caches. Selects the *oldest* version of a file.
+3. reject: Fails the build if there are any file conflicts.
+
+
+### Scoping
+
+Cache scoping levels can be expressed as `Job level < Event level < Pipeline level`. Job level scoping limits the cache to particular jobs, within a single event, within a single pipeline. Event level scoping limits the cache to a certain event within a single pipeline. Pipeline level scoping represents a globally scoped cache for the single pipeline, with the cache being available for all instances of the pipeline.
```
Zhongtang: For the job level scope, should it be a particular job within a single event or across different events? I think the later is probably a more common use case. For example I'm publishing an npm package in my component job. When I do npm install, I would like the dependencies to be in npm cache if I haven't touched my package.json file from my last commit (this logic is managed by npm itself)

design/build-cache.md
```diff
@@ -0,0 +1,201 @@
...
+shared:
+  # Caches defined in the shared section can be pipeline or exection scoped.
+  cache:
+    # A pipeline scoped cache.
+    pipeline:  
+      # The caching targets
+      target: ["build/**", "~/.sbt"]
+    # An event scoped cache
+    event:
+      target: ["target/deployment.jar"]
```
Zhongtang: While this makes sense as an example for caching files across different builds in the same event, a more common pattern we've been seeing people doing is explicitly publishing the produced artifact to registries (e.g. Artifactory) in one jobs, and pulling it in another for deployment.

design/build-cache.md
```diff
@@ -0,0 +1,201 @@
...
+Points to consider:
+
+- Should maybe just use GraphQL?
+- The data field might better be implemented as a link to a location like s3. This will increase system component count, complexity and risk
+
+#### Build node
+
+##### Command
+
+Not exposed as a user command. Used internally by director. Uses yaml data for cache details.
```
Zhongtang: `sd_director` is still in a proposal. I don't think there is guarantee that the team is going to work on it shortly.

design/build-cache.md
```diff
@@ -0,0 +1,201 @@
...
+
+#### Server
+
+##### API
+
+| Method | URL | Body | Reponse |
+| - | - | - | - |
+| GET | /pipeline/:pipeline_id/cache/:cache_id | none | A single cache entity.<br><b>Body</b><br>```{ "scope": "", "event_id": "", "pipeline_id": "", "ttl": 100,"data": base64}``` <br><b>Codes</b><br> 200, 404 |
+| GET | /pipeline/:pipeline_id/cache | none | List of currently valid caches for the pipeline.<br><b>Body</b><br>```[{ "scope": "", "event_id": "", "pipeline_id": "", "ttl": 100,"data": base64}]``` <br><b>Codes</b><br> 200, 404 | 
+| PUT | /pipeline/:pipeline_id/cache | ```{ "scope": "", "event_id": "", "pipeline_id": "", "data": base64``` | <b>Codes</b><br>202, 403 |
```
Zhongtang: Should we consider limiting the size of cache by some identifier? (e.g. pipeline ID)

design/build-cache.md
```diff
@@ -0,0 +1,201 @@
...
+
+Injecting malicious code into a build via the caching mechanism is a real concern. This can lead to theft of tokens or credentials from job, or in even lead to malicious code being deployed to a production environment. 
+
+Pull requests, as the main vector for injection of malicious code, require additional security considerations. Pull requests therefore treat caches as readonly. 
+
+Caching of potentially sensitive data also needs to be handled. Data provided via the secrets field should not be cached. 
+
+### Cache invalidation
+
+- [Proposed] Changing environment variables or secrets should invalidate existing caches. Mass invalidation can be avoided by including a hash of variables and secrets with the cache metadata.
```
Zhongtang: Invalidation imposed by SD because of change in environment variable etc. might break certain patterns of cache usage. e.g. For the case of npm install with cache. Cache invalidation should only happen when npm detects version changes etc, and npm actually can handle cache invalidation on its own.

design/build-cache.md
elyzion: Great point. thank you. I would like to expand on it:

- Total cache size is limited. The limit is the size of storage volume set up as the cache. This is a hard limit.
- Should we bother with a soft limit for cache size? A soft limit is something that can be set and updated by a system admin.
- Do we use LRU to start flushing old caches if the cache size limit is used?

design/build-cache.md
```diff
@@ -0,0 +1,201 @@
...
+
+| Method | URL | Body | Reponse |
+| - | - | - | - |
+| GET | /pipeline/:pipeline_id/cache/:cache_id | none | A single cache entity.<br><b>Body</b><br>```{ "scope": "", "event_id": "", "pipeline_id": "", "ttl": 100,"data": base64}``` <br><b>Codes</b><br> 200, 404 |
+| GET | /pipeline/:pipeline_id/cache | none | List of currently valid caches for the pipeline.<br><b>Body</b><br>```[{ "scope": "", "event_id": "", "pipeline_id": "", "ttl": 100,"data": base64}]``` <br><b>Codes</b><br> 200, 404 | 
+| PUT | /pipeline/:pipeline_id/cache | ```{ "scope": "", "event_id": "", "pipeline_id": "", "data": base64``` | <b>Codes</b><br>202, 403 |
+
+Points to consider:
+
+- Should maybe just use GraphQL?
```
elyzion: I doubt Hapi supports this. Can I scrap this?

design/build-cache.md
```diff
@@ -0,0 +1,201 @@
...
+4. Security
+   1. Is it possible to inject malicious code into pipelines via the cache mechanism? Malicious pull requests are a good example of this.
+
+### Security
+
+Injecting malicious code into a build via the caching mechanism is a real concern. This can lead to theft of tokens or credentials from job, or in even lead to malicious code being deployed to a production environment. 
+
+Pull requests, as the main vector for injection of malicious code, require additional security considerations. Pull requests therefore treat caches as readonly. 
+
+Caching of potentially sensitive data also needs to be handled. Data provided via the secrets field should not be cached.
```
elyzion: Testing for this might be expensive/troublesome. A simple `grep` could turn out to be quite expensive. Any suggestions for this?

design/build-cache.md
```diff
@@ -0,0 +1,201 @@
...
+1. merge: Compares creation time of files and directories. Recursively merges files and directories from multiple caches. Selects the *newest* version of a file.
+2. [Proposed] preserve: Compares creation time of files and directories. Recursively merges files and directories from multiple caches. Selects the *oldest* version of a file.
+3. reject: Fails the build if there are any file conflicts.
+
+
+### Scoping
+
+Cache scoping levels can be expressed as `Job level < Event level < Pipeline level`. Job level scoping limits the cache to particular jobs, within a single event, within a single pipeline. Event level scoping limits the cache to a certain event within a single pipeline. Pipeline level scoping represents a globally scoped cache for the single pipeline, with the cache being available for all instances of the pipeline.
+
+#### [Tentative] Regarding commits, branches and scoping
```
elyzion: Any comments on this? It's only tentative, so any indication of inclination would be appreciated. (Keep/disgard etc.)

elyzion:
@jithin1987 CC @Zhongtang @catto  Any comments? I would appreciate actually being able to have a meaningful discussion about this, hearing some more opinions, and then proceeding with a second edit of the proposal. If time is an issue, please give me a rough indication of when you will be able to look at this so that I can put it on the back burner for the time being.

design/build-cache.md
Zhongtang: Hmm so I guess to clarify, we are talking about the event-level scope. And there are 2 edge cases:
1. The same job reads multiple cache directories, and there could be overlap between these directories
2. Multiple jobs can write back the same cache directory if they run in parallel

For the 1st case, I think it would make sense for us to simply say that we don't support the use case and it's user's responsibility to make sure that there is no overlap, otherwise the result could be unpredictable. We can implement validation for the user, but it can be of low priority.

The second case is actually a bit trickier IMO. If job A and B run in parallel and A finishes before B, we might lose what's added to the cache dir in job A. But I guess since this is just cache, the only impact should be build time.

design/build-cache.md
Zhongtang: - I'm not against implementing local storage for only a single backend (this actually won't be very different from the multiple node + NFS case, as NFS will be mounted to all the nodes as local directory). But the conflict you talked about could exist even in the single backend scenario, if multiple jobs in the same event run in parallel on the same node?
- Yes NFS will also be slow. I guess the main value would be the abstraction as if they are local disk volumes, and the reduced overhead. But will need testing to compare

design/build-cache.md
Zhongtang: Could you elaborate on the use case of a Event scoped - Job level cache?

design/build-cache.md
Zhongtang: I can see your point of build dependency cache deviates from the purpose of SD. But build dependencies within the same event is fine IMO, since you generally want to lock down the dependencies in the same event. Also even for dependencies across different events, it should still be fine as long as people are using dependency management tools that can resolve dependency versions correctly (e.g. npm).

design/build-cache.md
Zhongtang: Yes I thought the `deployment.jar` was the final artifact the pipeline produces and was concerned that the example would encourage people to put these types of resources to cache dir. I think `generated intermediate resources` is probably a better description.

design/build-cache.md
elyzion: Yeah, the first case deals with writing to build space, the second case deals with writing to the cache.

## Case 1

> I think it would make sense for us to simply say that we don't support the use case and it's user's responsibility to make sure that there is no overlap, otherwise the result could be unpredictable.

Ditto. Leaving this as a userland responsibility is the best choice in my opinion, as well. 

> We can implement validation for the user, but it can be of low priority.

I thought about implementing a validation lint for the user, but it could get tricky if the user is using templates and an update in the templates causes case (1) to happen. 

It would still be a good idea to clearly define the order multiple caches get applied in. An on the spot suggestion would be to apply from broadest to narrowest scope. Multiple caches within the same scope get applied in order they are listed in (for job scopes) or defined in (for pipeline or event scopes).

## Case 2
> But I guess since this is just cache, the only impact should be build time.

Essentially, yes. The caching side of things needs to properly handle this case though. (Not really a problem using modern frameworks.)

## Conclusion

1. We leave case 1 as a user issue, but we do take care to explicitly define the order of cache application. The implementation should avoid parallelization of cache extraction as this can lead to increased unpredictability, unless special care is taken to ensure reproducibility.
2. The cache implementation must be able to handle multiple concurrent write requests for the same cache. The last write gets persisted.

design/build-cache.md
```diff
@@ -0,0 +1,201 @@
...
+    requires: [~pr]
+    # Indicate which job scoped caches to use.
+    cache: ["my-cache-1"]
+    image: rust:latest
+    steps:
+      - test: cargo build
+  publish-preview:
+    # Indicate which job scoped caches to use.
+    cache: ["my-cache-1"]
+    requires: [~commit]
```
tk3fftk: I think the cache does not work in this example. `build` is triggered by a pull request event and `publish-preview` is triggered by a commit event. According to scoping section, job level scoping works within a single event. These events are difference so SD creates caches in each job and these caches are not shared.

design/build-cache.md
```diff
@@ -0,0 +1,201 @@
...
+  cache:
+    # A pipeline scoped cache.
+    pipeline:  
+      # The caching targets
+      target: ["build/**", "~/.sbt"]
+    # An event scoped cache
+    event:
+      target: ["target/deployment.jar"]
+    # A job scoped cache 
+    my-cache-1:
```
tk3fftk: To adjust to other formats, how about below format?
```
job:
  my-cache-1:
    target: ["build/some-artifact.zip"]
```

On the other hand, `target` key looks verbose. I think these can be more simply.

```
  cache:
    # A pipeline scoped cache.
    pipeline: ["build/**", "~/.sbt"]
    # An event scoped cache
    event: ["target/deployment.jar"]
    # A job scoped cache 
    job:
      my-cache-1: ["build/some-artifact.zip"]
      my-cache-2: ["build/some-artifact2.zip"]
```

design/build-cache.md
elyzion: Yeah, it's a simple purist statement. I foresee caching of build dependencies on a pipeline level. (For things like npm or sbt managed dependencies)

design/build-cache.md
elyzion: > Could you elaborate on the use case of a Event scoped - Job level cache?

This is the existing definition, right? As is says in the proposal, _Job level scoping limits the cache to particular jobs, within a single event, within a single pipeline. Event level scoping limits the cache to a certain event within a single pipeline. _.

Use case for job scoped caches, as in the current proposal:
 - Caching of files/artifacts/data that are only used from certain jobs.  

Use case for event scoped caches, as in the current proposal:
- Caching of files/artifacts/data that should be available to all jobs within an event.

design/build-cache.md
elyzion: ## Conclusion

- [ ] Fix `deployment.jar` and related wording to avoid confusion.

design/build-cache.md
elyzion: Yeah, I've discussed this with @catto. The proposal is to use the director, but the initial implementation will use the cli worker or whatever makes sense.

design/build-cache.md
elyzion: Thanks @tk3fftk, that's a great suggestion. Will reflect change the syntax to reflect this.

- [ ] @elyzion Update job level cache definition example syntax.

design/build-cache.md
elyzion: Yeah, this example is meant to be a syntax proposal. Consequently, I am leaving out addition jobs that might use caches and just serve to increase the amount of intellectual load required to parse the example.

Check the comment in line 107. I'm proposing a slightly different way of scoping the caches.

design/build-cache.md
Zhongtang: There does need to be a global total cache size limit and we will need some form of cache eviction maintained at the server end. Although my comment was more about rate-limiting each individual pipeline's cache space usage. Both are important IMO.

design/build-cache.md
Zhongtang: - 👍 I agree that r/w mount of NFS could be slow. Let's maybe focus on 1) the remote caching implementation and 2) local caching with a single backend implementation, until we have enough data to see how NFS performs.
- One other possibility is to copy the content in NFS to local files when build starts, and write them back when build finishes. So it would be identical to the remote caching proposal, but NFS will be used in place instead of HTTP
- If we go with implementing both the remote caching and a local disk caching, do we provide them as options that an SD admin can pick?
- I would still like to address the potential concurrency issue in the local storage caching proposal, where multiple builds can the same directory on the single execution node.

design/build-cache.md
Zhongtang: I guess there might be some misunderstanding, but I don't see how a event-scoped + job-level cache could be useful? Cache is only useful if multiple builds use it. But `event-scoped + job-level` essentially means only one build will be able able to use it?

design/build-cache.md
Zhongtang: We don't have to use Hapi to implement the cache server if it doesn't fit the design (unless we plan to reuse the store API). But I'm actually not very familiar with GraphQL... What's the reason of using GraphQL instead of REST?

design/build-cache.md
Zhongtang: There might not be security issue if we only limit read access of the cache content to the particular pipeline?

Also maybe we can defer the responsibility of not putting sensitive data to users for now.

design/build-cache.md
Zhongtang: I think these could be valid use cases. Maybe we add more scoping in addition to event-level and pipeline-level (and implement them later)?

design/build-cache.md
elyzion: @Zhongtang @tk3fftk @catto Any thoughts on security issues?

design/build-cache.md
elyzion: - [ ] Create issue for testing performance of NFS backed shared volume vs copying for every job.

design/build-cache.md
elyzion: The current proposal with an event and job scoped cache is based on the premise that you might have something you want to cache, but for whatever reason you do not want said cache to be extracted into certain jobs. Maybe the data is only being used in 2 jobs, but you have 20 jobs in total. In such a case it might make more sense performance wise to only use caching on the 2 relevant jobs, instead of all jobs.

I still think reworking scoping from the current proposed format to the following might be a good idea.

> However, scoping could be reworked to be like this:

 - Event scoped, only valid for a single event
   - Job level caching
   - Pipeline level caching
 - Statically scoped, valid across all executions of a pipeline
   - Job level caching
   - Pipeline level caching

This will allow for more flexibility and it also removes the ambiguity between event and job scoped caches.

jithin1987:
We will scope out the work required and create separate issues for these. 

A remote network bases cache might not be that bad depending on network conditions. Let's say the NFS with k8s might not be bad in terms of performance provided the NFS servers and k8s build nodes have network proximity. So implementation of this proposal would be dependent on the type of SD build cluster. We should define a an API/CLI contract and leave implementation to cluster admins.

PR #1168, opened 2018-07-05
fix: split encryptionPassword

minz1027:
Currently, we're using the `encryptionPassword` for two different purposes.
- encrypt/decrypt pipeline secrets/user Oauth token
- as a one way hashing key for aceess token 
https://github.com/screwdriver-cd/models/blob/master/lib/tokenFactory.js#L54

This PR will split it into two to serve its own purpose and for better key management.

### DIFF
--- a/config/custom-environment-variables.yaml
+++ b/config/custom-environment-variables.yaml
@@ -15,9 +15,12 @@ auth:
     # A password used for encrypting session data.
     # **Needs to be minimum 32 characters**
     cookiePassword: SECRET_COOKIE_PASSWORD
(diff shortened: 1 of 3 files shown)


config/custom-environment-variables.yaml
```diff
@@ -18,6 +18,9 @@ auth:
     # A password used for encrypting stored secrets.
```
d2lam: can we add comment for this also used to encrypt user oath?

minz1027: Sure

Issue #450, opened 2017-02-07
Fails to delete a Pipeline belonging to a non-existent repository


FenrirUnbound:
## Expected

Given a deleted repository, when you delete the corresponding Pipeline, it will remove the Pipeline and all data associated with it.

## Actual

Delete a repository. When you delete the corresponding Pipeline, the server returns a 500. The Pipeline also continues to persist.

## Initial Investigation

There is [a permissions check](https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/pipelines/remove.js#L37) to validate that the User can modify the repository. This, in turn, [queries the Github API](https://github.com/screwdriver-cd/models/blob/master/lib/user.js#L55) on that repository. Since it's deleted, the `scm.getPermissions()` call throws an error.

FenrirUnbound:
Given how Github treats a protected repository as a missing repository, I'm unsure of a clean way to handle this scenario. At least serving a proper 404 would be a nice start, but that doesn't resolve the specific workflow of deleting things out of order.

d2lam:
Yeah this is a known problem. 
It's good you created an issue, so we can track it here.

stjohnjohnson:
I think there should be two failure cases:
 - Branch doesn't exist
 - Repo doesn't exist

In the event of Repo doesn't exist, the project is technically "orphaned" and we should either auto-remove it or inform the admins to remove it.

Filbird:
Indeed, if a user attempts to delete a private repository it does not have access to, a 404 is thrown. For the time being I think we can enable deletion of pipelines with missing repositories if the `privateRepo` flag is set to false.

It seems that in the case of a branch not existing, the API does NOT return a 404. I wonder if it would be worth making use of web hooks to automatically remove a pipeline when a branch is deleted?

Filbird:
@tkyi Brought up an alternative solution in https://github.com/screwdriver-cd/screwdriver/pull/1187.

> It's possible we could get a 404 code for a different reason, and I'm concerned we're not covering all cases right now. What if we don't allow users to delete a pipeline when there's a 404, but we allow Screwdriver cluster admins. That way, users can file a support request if they really need the pipeline to be deleted and we can manually do it.
We can reuse the check used here: https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/banners/index.js#L24

I think it makes sense to err on the side of caution and go with this approach.

PR #1187, opened 2018-07-16
fix(450): Allow pipeline to be removed if repository does not exist

Filbird:
## Context
Currently, pipelines cannot be removed if its underlying repository does not exist.

## Objective
Allow pipeline to be removed if `getPermissions` returns a 404 and only if `privateRepo` is disabled.

## References
Pending:
  - https://github.com/screwdriver-cd/scm-bitbucket/pull/52
https://github.com/screwdriver-cd/screwdriver/issues/450

### DIFF
--- a/plugins/pipelines/remove.js
+++ b/plugins/pipelines/remove.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const boom = require('boom');
+const hoek = require('hoek');
 const joi = require('joi');
 const schema = require('screwdriver-data-schema');
 const idSchema = joi.reach(schema.models.pipeline.base, 'id');
@@ -26,6 +27,8 @@ module.exports = () => ({
             const userFactory = request.server.app.userFactory;
             const username = request.auth.credentials.username;
             const scmContext = request.auth.credentials.scmContext;
+            const scms = hoek.reach(pipelineFactory, 'scm.scms') || {};
+            const isPrivateRepo = hoek.reach(scms[scmContext], 'config.privateRepo') || false;
 
             // Fetch the pipeline and user models
             return Promise.all([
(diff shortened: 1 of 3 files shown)


plugins/pipelines/remove.js
```diff
@@ -52,6 +52,12 @@ module.exports = () => ({
                                 + 'does not have admin permission for this repo');
                         }
                     })
+                    .catch((error) => {
+                        // Allow pipeline to be removed if the repository does not exist
+                        if (error.code !== 404) {
```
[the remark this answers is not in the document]
Filbird: Yep, added the check!

Will look into the Bitbucket and Gitlab cases

tkyi:
It's possible we could get a 404 code for a different reason, and I'm concerned we're not covering all cases right now. What if we don't allow users to delete a pipeline when there's a 404, but we allow Screwdriver cluster admins. That way, users can file a support request if they really need the pipeline to be deleted and we can manually do it.
We can reuse the check used here: https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/banners/index.js#L24

Filbird:
@tkyi  I see, this would certainly be a safer way of solving the issue!

I'll post your comment on the issue page and update this PR.

plugins/pipelines/remove.js
```diff
@@ -52,6 +55,18 @@ module.exports = () => ({
                                 + 'does not have admin permission for this repo');
                         }
                     })
+                    .catch((error) => {
+                        // Lookup whether user is admin
+                        const adminDetails = request.server.plugins.pipelines
```
tkyi: I think you can reuse the banners code like this instead: https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/builds/update.js#L67
Not sure what makes more sense

Filbird: Makes sense! Better than duplicating it.

PR #1209, opened 2018-07-20
fix: do not allow PR to do POST/DELETE/PUT for templates/commands 

minz1027:
do not allow PR to do POST/DELETE/PUT for templates/commands

### DIFF
--- a/plugins/commands/create.js
+++ b/plugins/commands/create.js
@@ -126,6 +126,7 @@ module.exports = () => ({
         },
         handler: (request, reply) => {
             const data = request.payload;
+            const isPR = request.auth.credentials.isPR;
             let commandSpec;
             let commandBin;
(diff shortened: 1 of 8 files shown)


plugins/commands/create.js
```diff
@@ -171,7 +172,8 @@ module.exports = () => ({
 
                         // If command name exists, but this build's pipelineId is not the same as command's pipelineId
                         // Then this build does not have permission to publish
-                        if (commands.length !== 0 && pipeline.id !== commands[0].pipelineId) {
+                        if (isPR ||
+                                (commands.length !== 0 && pipeline.id !== commands[0].pipelineId)) {
                             throw boom.unauthorized('Not allowed to publish this command');
```
tkyi: Should this be `unauthorized` or `forbidden`?

Issue #1114, opened 2018-06-11
Improve step log load time


jithin1987:
## Context

Loading step logs is slow for steps which generates large number of log lines. This is because we store step logs as files with 100 lines per file. Also we lack metadata on number of log files per step. This results in UI issues requests to read step logs file by file until we reach end of files, which makes loading step logs really slow for steps with large number of lines. 

UI waits to load entire log for a step before it scrolls to bottom to show latest, which does not perform well because of above reasons. So we need a way to just show latest logs without having to download the entire logs and progressively load rest of them. 

## Objective

 - [x] Reduce number of files generated per step. #1043
 - [x] Store number of log files for a step in a metadata file per step.
 - [x] Use metadata file to load latest logs first.
 - [x] Load logs in reverse in UI
 - [x] Start loading older logs when user starts scrolling.
 - [x] Option to download a step's log

## References
1. https://github.com/screwdriver-cd/screwdriver/issues/953
1. https://github.com/screwdriver-cd/screwdriver/issues/1024
1. https://github.com/screwdriver-cd/screwdriver/issues/1043

catto:
I found a bug that the latest api and log-service combination cannot handle old logs that have 100+ lines. I confirmed that logs from a new build work fine.
https://cd.screwdriver.cd/pipelines/1/builds/44579
Is this a breaking change? #1043

vengopal:
@minz1027 @jithin1987   can we please reply for @catto .

tkyi:
@catto Sorry, I think @minz1027 commented in one of her PRs saying that the new API needed to be pulled in before the new launcher (https://github.com/screwdriver-cd/screwdriver/pull/1134#issue-196302740).

minz1027:
### Updates 07/23

Will do the optimization for finished build step first. We will update number of pages for a step log when step finishes. For finished step, UI will call API to fetch logs from the last page. And when user scroll up then go fetch the previous page.

- [ ] `data-schema`: add `pages` in build step meta
- [ ]  `log-service`: take in api url to update build step meta
- [ ] `api`: support load logs from the end
- [ ] `ui`: fetch logs from the end for finished steps

catto:
@tkyi Thanks for your comment, but I meant to say it seems that there is no backward compatibility.
Please see your cluster: https://cd.screwdriver.cd/pipelines/1/builds/44579 
This is an old build before the upgrade. `test` step shows only first 100 lines even if both API and launcher has been upgraded.
[...]

PR #1215, opened 2018-07-25
feat(1114): Allow for loading logs in descending order

tkyi:
## Context

Log loading is slow and people usually want to see the end of their logs. We think it can improve performance if we load logs from the end to the beginning.

## Objective

This PR adds support for loading logs in descending order. You can add an extra query param to `/builds/{id}/steps/{name}/logs` called `sort`, which accepts `ascending` or `descending` as an argument.

Note: I also did some refactoring, not sure if it makes sense. Please help.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/1114


plugins/builds/steps/logs.js
```diff
@@ -4,62 +4,109 @@ const boom = require('boom');
...
             })
+            .on('error', e => reject(e))
             // Parse the ndjson
             .pipe(ndjson.parse({
                 strict: false
             }))
             // Filter down to the lines we care about
             .on('data', (line) => {
-                if (line.n >= linesFrom) {
+                if (sort === 'descending' || line.n >= linesFrom) {
```
minz1027: To make this easier to understand, can we do something like this,
```
const shouldInclude = sort === 'ascending' ? line.n >= linesFrom : line.n <= linesFrom
if (shoudlnclude) {}
```

PR #1221, opened 2018-07-26
fix: fix log fetching for old logs

minz1027:
Log loading randomly failed to load old logs with more than 100 lines. 
After investigation, i think the cause is maxLines is a global variable and can be shared by all requests. It's possible when trying to get big logs, it's set to 1000 and requests after that cannot get logs with 100 lines.


catto:
ref: #1114

Issue #1211, opened 2018-07-24
Display build aborted reason when a PR is closed


jithin1987:
**What happened**:

If a pull request is closed, SD aborts the running build but does not explain why in the UI. SD UI should display a status message like it display when a user aborts the build.

**What you expected to happen**:

When a pull request is closed, we should display a status message in build detail page explaining why build was aborted.

**How to reproduce it**:

1. Close a PR for when SD PR build is running.
2. Build is aborted.
3. No information is displayed in UI

tkyi:
This feature is working:

When a new commit is pushed to PR:
<img width="1348" alt="screen shot 2018-08-06 at 10 17 22 am" src="https://user-images.githubusercontent.com/3230529/43730983-fd22f44c-9961-11e8-85c4-cea98ff5e9f5.png">

When a PR is closed:
<img width="1348" alt="screen shot 2018-08-06 at 10 18 42 am" src="https://user-images.githubusercontent.com/3230529/43731042-25a4ca44-9962-11e8-915a-6c07291bff4c.png">

PR #1233, opened 2018-07-31
fix(1211): Display build aborted reason when a PR is closed

tkyi:
## Context

We set the build status message when a user aborts a build in the UI. It would be nice if we had an equivalent status message when the build is aborted because a PR is closed.

## Objective

This PR adds a status message for aborted PR builds.

## References

Issue: https://github.com/screwdriver-cd/screwdriver/issues/1211


minz1027:
Can be aborted in PR sync event, e.g. when a new commit came it. Should customize message for that

PR #1236, opened 2018-08-06
feat: [small] able to add sequelize retry

yuichi10:
## Context

When using active/active database cluster and do multi-node writing, dead lock will happen. 

So I want to make it possible to use retry.

## Objective

Add retry settings to a config file.

## References
- [Multi-node writing and Unexpected deadlocks](https://www.percona.com/blog/2012/08/17/percona-xtradb-cluster-multi-node-writing-and-unexpected-deadlocks/)

### DIFF
--- a/config/custom-environment-variables.yaml
+++ b/config/custom-environment-variables.yaml
@@ -73,6 +73,9 @@ datastore:
         pool:
           __name: DATASTORE_SEQUELIZE_POOL
           __format: json
+        retry:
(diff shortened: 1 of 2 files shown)


tkyi:
Not sure if you also want to add a key in the `default.yaml` for `retry` like `pool` https://github.com/screwdriver-cd/screwdriver/blob/master/config/default.yaml#L63

yuichi10:
@tkyi I Add a key in the default.yaml!

PR #1238, opened 2018-08-06
fix(1114): Use step model for build step update

tkyi:
### Context
To avoid race conditions when updating step meta when a build is running, we should separate steps into a separate model from the build. The API should update the step model directly rather than overriding the entire `steps` field of a build model.

The model will handle calling the step factory in a build GET request to populate the step data appropriately.

### Objectives
This PR contains the following changes: 

- add step factory
- check for step in build update and update step model directly

### Related links
This bug is related to changes made for https://github.com/screwdriver-cd/screwdriver/issues/1114
Blocked by: https://github.com/screwdriver-cd/models/pull/280


plugins/builds/steps/get.js
```diff
@@ -21,15 +21,15 @@ module.exports = () => ({
...
-                .then((model) => {
-                    if (!model) {
+            buildFactory.get(request.params.id)
+                .then((buildModel) => {
+                    if (!buildModel) {
                         throw boom.notFound('Build does not exist');
                     }
 
-                    const stepModel = model.steps.filter(step => (
+                    const stepModel = buildModel.steps.filter(step => (
```
jithin1987: won't this only fetch the initial step data which `buildModel` has ?

plugins/builds/steps/update.js
```diff
@@ -20,43 +20,66 @@ module.exports = () => ({
...
                         throw boom.notFound('Build does not exist');
                     }
-                    const steps = build.steps;
                     const now = (new Date()).toISOString();
 
-                    stepIndex = steps.findIndex(step => step.name === request.params.name);
+                    // Check if step model exists
+                    return stepFactory.get({ buildId, name: stepName })
+                        .then((step) => {
+                            if (!step) {
```
jithin1987: will this happen, once new code goes in ? Is this specifically for builds which were created before this release and still running after release ?

plugins/builds/steps/get.js
tkyi: No, Min's model changes will call the step model and set the step information in a build when this get is called.

plugins/builds/steps/update.js
tkyi: Yes, this is for backwards compatibility. I think it should only apply if a build is in the middle of running when this change is pulled in.

PR #1252, opened 2018-08-14
fix: [1] able to promote tag by tag

yuichi10:
## Context

Now sd-cmd cannot promote tag by tag. 

## Objective

Be able to promote tag with tag.
(e.g. 
```
sd-cmd promote hoge/huga latest stable
```

todo:
- [x]  add test

### DIFF
--- a/test/plugins/commands.test.js
+++ b/test/plugins/commands.test.js
@@ -993,7 +993,6 @@ describe('command plugin test', () => {
             version: 'latest'
         };
         const testCommandTag = decorateObj(hoek.merge({ id: 1 }, payload));
-        const testCommandTagWithTag = decorateObj(hoek.merge({ id: 1 }, payloadTag));
 
         beforeEach(() => {
             options = {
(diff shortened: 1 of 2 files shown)


test/plugins/commands.test.js
```diff
@@ -1091,6 +1095,70 @@ describe('command plugin test', () => {
...
+                    name: 'test',
+                    tag: 'stable',
+                    version: '1.2.0'
+                });
+                assert.calledTwice(commandTagFactoryMock.get);
+                assert.equal(reply.statusCode, 201);
+            });
+        });
+
+        it('update command tag if has good permission and tag exists with tag', () => {
```
tk3fftk: update**s**
if has -> if it has
what do you mean by `tag exists with tag` ? I think tags are attached to a version.

test/plugins/commands.test.js
```diff
@@ -1091,6 +1095,70 @@ describe('command plugin test', () => {
             });
         });
 
+        it('creates commands tag if has good permission and tag does not exist with tag', () => {
```
tk3fftk: command tags?

test/plugins/commands.test.js
sakka2: > updates

fixed

> if has -> if it has

fixed

> what do you mean by tag exists with tag ? I think tags are attached to a version.

I changed the sentence a bit.
I write `with tag` because this test send tag information to API instead of version.

test/plugins/commands.test.js
sakka2: fixed to `a command tag` because this test creates just one tag.

yuichi10:
I am going to update README!

Issue #880, opened 2018-02-13
getting slower to show pipeline list


catto:
Currently the pipeline list page doesn't have pagination capability. If I have thousands of pipelines in my cluster, it takes very long to load pipeline list and render them.
I tested with my cluster which has 2000 pipelines, it takes 1 second to load pipeline JSON from API and 5 seconds to render on a browser. It will be still getting slower if I add more pipelines.

It would be nice to have pagination or number limit on pipeline list page.

tkyi:
Related to https://github.com/screwdriver-cd/screwdriver/issues/233

tkyi:
This story would greatly speed up interaction with the pipeline search/list page. It should be pretty similar to the events list pagination story.

## Implementation

- [x] ~[models] Add `list` method with optional pagination for pipelineFactory~ handled by baseFactory `list` (https://github.com/screwdriver-cd/models/blob/master/lib/baseFactory.js#L126-L133)
- [x] ~[screwdriver] Pass pagination from GET `pipelines` endpoint to models `list` method~ already exists (https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/pipelines/list.js#L40-L45)
- [ ] [UI] Update UI to handle pipeline list pagination

tkyi:
**Update 9/17:**
The pagination is almost complete, but there’s an issue with sorting the pipelines alphabetically by name. Since we’ve added pagination, we can no longer do the sorting in the UI, it must be done when making the call to the datastore.

The issue is that the information needed, `name`, is stored in a JSON string in `scmRepo` in the DB:
```
{"branch":"tkyi-patch-1","name":"tkyi/mytest","url":"https://git.ouroath.com/tkyi/mytest/tree/tkyi-patch-1"}
```
Initially, we tried to use change the `sortKey` to `scmRepo.name`(https://github.com/screwdriver-cd/datastore-sequelize/blob/master/index.js#L367-L372), but parsing the JSON string is difficult.

The alternative solution we're thinking about is adding `name` as its own field at the root level in the pipeline object, but this change will require a data migration.

tkyi:
## Manual Steps for Existing Databases
In order to use these latest changes, you will need to do some SQL queries to migrate data from `scmRepo` to the new `name` field. This `name` field will be used for sorting and searching in the Search UI. 

### Without migrating
[...]

PR #1276, opened 2018-09-06
fix(880): Pass search field and term to datastore for pipeline list [4]

tkyi:
## Context
Since we're adding pagination for pipelines, we can no longer have search functionality stored in the UI. 

## Objective
This PR passes search field and term to the datastore from the `GET /pipelines` endpoint. It will search the `scmRepo` text field for a term matching `%name%{pipeline_name}%`.

scmRepo field example:
```
{"branch":"tkyi-patch-1","name":"tkyi/mytest","url":"https://git.ouroath.com/tkyi/mytest/tree/tkyi-patch-1"}
```

To do a search, it expects a param like this:
```
params: {
  search: {
    field: "scmRepo",
    keyword: "%name%tkyi/mytest%"
  }
}
```

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/880
Blocked by https://github.com/screwdriver-cd/models/pull/290, https://github.com/screwdriver-cd/datastore-sequelize/pull/32, https://github.com/screwdriver-cd/data-schema/pull/278

### DIFF
--- a/plugins/pipelines/list.js
+++ b/plugins/pipelines/list.js
@@ -44,7 +44,9 @@ module.exports = () => ({
                 if (request.query.search) {
                     config.search = {
                         field: 'scmRepo',
-                        term: `%name%${request.query.search}%` // Do a fuzzy search for name: screwdriver-cd/ui
+                        // Do a fuzzy search for name: screwdriver-cd/ui
+                        // See https://www.w3schools.com/sql/sql_like.asp for syntax
+                        keyword: `%name%${request.query.search}%`
                     };
                 }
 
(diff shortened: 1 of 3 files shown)


plugins/pipelines/list.js
```diff
@@ -41,6 +41,13 @@ module.exports = () => ({
                     config.sortBy = request.query.sortBy;
                 }
 
+                if (request.query.search) {
+                    config.search = {
+                        field: 'scmRepo',
+                        term: `%name%${request.query.search}%` // Do a fuzzy search for name: screwdriver-cd/ui
```
minz1027: Let's add a reference link here to indicate that this `%` syntax is from SQL.

Issue #1085, opened 2018-05-23
Allow job dependencies in PR builds


edudar:
**What happened**:
A pipeline that has job dependencies will only run jobs that require `~pr` for a pull request but won't run jobs that depend on these with `~pr`. Configuration example:
```
jobs:
  main:
    requires: [~pr, ~commit]
  client:
    requires: [main]
```
PR build runs `main` but completely ignores `client` job.

**What you expected to happen**:
We don't want to have different builds depending on whether it's pull request build or not. Some deviations here are OK like we won't likely publish binaries from PR build but when jobs run things like additional set of tests, let's sey integration, and we want to make sure they pass in PR build. Overall I'd expect that master and branch/PR builds would run identically unless it's explicitly configured.

petey:
Yeah, I think this is kind of an arbitrary limit on on the workflow configuration right now. It would be important to have successive jobs retain the $SD_PULL_REQUEST value so that users could continue to gate script activity based on whether or not a job started as a result of a PR or not.

The UX surrounding how PRs are handled currently is a bit misleading when compared to the yaml config. See #909, #1010

stolfi:
I agree on the UX. The fact that the flow diagrams show the chaining of the jobs is very misleading.

jithin1987:
This will be solved with https://github.com/screwdriver-cd/screwdriver/issues/823

Issue #1242, opened 2018-08-09
specific branch job test 


yuichi10:
Now there is no test for specific branch job function.
To release Screwdriver with confidence, need to add function test for the fucntion.

related to https://github.com/screwdriver-cd/screwdriver/issues/1082

PR #1286, opened 2018-09-11
fix(1242): add scenario of branch filtering

tk3fftk:
## Context
There is not feature test for branch filtering workflow.

## Objective
This PR adds scenarios for branch filtering workflow.
Also updates pipeline definition of other scenarios to using `requires`.

## References
design doc: https://github.com/screwdriver-cd/screwdriver/blob/ba6478f5d43acae3fa1d5396cc2c417a591cee7a/design/workflow.md
guide: https://docs.screwdriver.cd/user-guide/configuration/workflow#branch-filtering
related issue: https://github.com/screwdriver-cd/screwdriver/issues/1242


features/workflow.feature
```diff
@@ -58,3 +64,55 @@ Feature: Workflow
...
+            | FOO   | ~pr             |
+            | BAR   | ~pr:staging     |
+            | BAZ   | ~pr:/^.*$/      |
+            | QUX   | ~pr:/master     |
+        When a pull request is opened
+        And it is targeting the pipeline's branch
+        Then the "FOO" job is started
+        And the "BAR" job is not started
+        And the "BAZ" job is started
+        And the "QUX" job is started
```
tk3fftk: "QUX" job is not started in current implementation.
ref: https://github.com/screwdriver-cd/screwdriver/blob/ba6478f5d43acae3fa1d5396cc2c417a591cee7a/plugins/webhooks/index.js#L176

Do we need to allow and work `~pr:<pipeline_branch>` ?

yuichi10:
I have questions about the specific PR workflow.
1) Should specific PR run when the specific branch is same with the pipeline branch? (Same with https://github.com/screwdriver-cd/screwdriver/pull/1286#discussion_r216532684 )
2) Should specific PR allow to trigger?
3) Should specific PR allow to publish or deploy?

I want to know how the spcific PR workflow work.

tk3fftk:
In my opinion, 1-3 should be worked. This feature will fix https://github.com/screwdriver-cd/screwdriver/issues/1085 by another implementation.

yuichi10:
I see. I also think it is a good solution that specific PR allow 1-3 for the issue!
Then do we need the test about specific PR triggers? Also, do we need to test that ~PR will not trigger?

tk3fftk:
> Then do we need the test about specific PR triggers? Also, do we need to test that ~PR will not trigger?

I'll add trigger test cases.

features/workflow.feature
```diff
@@ -58,3 +64,71 @@ Feature: Workflow
...
+        Given an existing pipeline with the workflow:
+            | job   | requires        |
+            | FOO   | ~commit         |
+            | BAR   | ~commit:staging |
+            | BAZ   | ~commit:/^.*$/  |
+        When a new commit is pushed
+        And it is against the pipeline's branch
+        Then the "FOO" job is started
+        And the "BAR" job is not started
+        And the "BAZ" job is started
```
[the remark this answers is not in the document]
tk3fftk: Other cases check SHA with triggered job. This case a job is triggered by commit so I think it does not need in this case. But other scenario may need SHA check. Thanks.

features/workflow.feature
```diff
@@ -58,3 +64,71 @@ Feature: Workflow
...
+            | FOO2  | FOO             |
+            | QUX   | ~pr:/master     |
+            | QUX2  | QUX             |
+        When a pull request is opened
+        And it is targeting the pipeline's branch
+        Then the "FOO" job is started
+        And the "QUX" job is started
+        Then the "FOO" job is succeeded
+        And the "QUX" job is succeeded
+        Then the "FOO2" job is not started
```
yoshwata: should FOO2 be started?

tk3fftk: This example is a little strange but `FOO2` is not started. `FOO2` job requires `FOO` and `FOO` requires `~pr`.  `~pr` will not trigger subsequence jobs.

features/workflow.feature
```diff
@@ -46,15 +50,91 @@ Feature: Workflow
...
+        And that "QUX" build uses the same SHA as the "FOO" build
+
+    Scenario: Branch filtering (the pipeline's branch is committed)
+        Given an existing pipeline with the workflow:
+            | job   | requires        |
+            | FOO   | ~commit         |
+            | BAR   | ~commit:staging |
+            | BAZ   | ~commit:/^.*$/  |
+        When a new commit is pushed
+        And it is against the pipeline's branch
```
tkyi: I think `pipeline's branch` wording is confusing for me
Can we call out the branch specifically by saying something like `And it is on the "master" branch`

And is there a difference between `it is against` vs `it is targeting`?

features/workflow.feature
```diff
@@ -18,25 +18,29 @@ Feature: Workflow
...
         When the "FOO" job is started
         And the build succeeded
         Then the "BAR" job is started
         And that "BAR" build uses the same SHA as the "FOO" build
 
     Scenario: Parallel
         Given an existing pipeline with the workflow:
-            | job | triggers |
-            | FOO | BAR+BAZ  |
+            | job | requires |
```
tkyi: 👍 thumbs up to changing from `job/triggers` to `job/requires`

features/workflow.feature
tk3fftk: `pipeline's branch` is reused from git-flow feature:
https://github.com/screwdriver-cd/screwdriver/blob/a37c2485340e84240bef8147d2f9bdd337c6f082/features/git-flow.feature#L37
But I agree with you that it's confusing so I fixed.

> And is there a difference between it is against vs it is targeting?

unified with `it is targeting`

tk3fftk: @tkyi fixed!

PR #1294, opened 2018-09-14
fix(757): allow update build status msg only

minz1027:
If payload contains only statusMsg, only updates that.

Blocked by: https://github.com/screwdriver-cd/data-schema/pull/281/files

### DIFF
--- a/plugins/builds/update.js
+++ b/plugins/builds/update.js
@@ -85,6 +85,13 @@ module.exports = () => ({
 
                     return eventFactory.get(build.eventId).then(event => ({ build, event }));
                 }).then(({ build, event }) => {
+                    // If payload only has statusMessage
+                    if (!desiredStatus && statusMessage) {
+                        build.statusMessage = statusMessage;
+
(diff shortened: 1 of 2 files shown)


plugins/builds/update.js
```diff
@@ -85,6 +85,11 @@ module.exports = () => ({
 
                     return eventFactory.get(build.eventId).then(event => ({ build, event }));
                 }).then(({ build, event }) => {
+                    // If payload only has statusMessage
+                    if (!desiredStatus && statusMessage) {
```
tkyi: is this `build.statusMessage`?

minz1027: this statusMessage is from payload: https://github.com/screwdriver-cd/screwdriver/blob/fef6c46497a34b651b2b09868f2d66e32a57e666/plugins/builds/update.js#L30

minz1027: ah, missing the assignment!  fixed

Issue #1271, opened 2018-09-05
Improvements for templates and commands list page


catto:
## Context
Template list page with lots of templates are slower to load. SD Command list page also.
And it is hard for users to find template/command since search/filter feature doesn't exist if pagination is implemented.

## Objective
* Templates and Commands API should support pagination.
* UI should provide search feature for these pages.

tkyi:
## Design Mocks
### Template landing page
![screen shot 2018-10-02 at 11 25 01 am](https://user-images.githubusercontent.com/3230529/46368801-00473380-c636-11e8-9141-c75acc1d0ae2.png)

- search will search through all fields: `namespace`, `name`, `description`
- results will return sorted by `name`
- pagination will look similar to pipeline events and search page

### Template Namespace page
![screen shot 2018-10-02 at 11 27 30 am](https://user-images.githubusercontent.com/3230529/46368890-400e1b00-c636-11e8-83c9-8f8328dc0840.png)
_Note: Not sure if we still need this or we can just reuse the landing page._

tkyi:
## Implementation

#### Backend Tasks
- [ ] List namespaces (e.g.: `GET /templates/namespaces`)
- [ ] Multiple field search support
- [ ] Pagination support
- [ ] "Trusted" namespace support

tkyi:
## "Trusted" feature

Some points of discussion when it comes to the "trusted" feature:
### 1. How will it be managed/maintained?
If we have a list of "trusted" elements, it will most likely be manually maintained by the Screwdriver admins.
Alternatively, we could have users "star" their favorite or trusted templates and display the "most starred" templates accordingly.

### 2. How will that "trusted" tag/flag be maintained?
Templates can only be created, not updated, so it's unclear how we would track the "trusted" flag (or even "most starred"). It doesn't make sense to have owners add the "trusted" flag themselves when creating the template and we would have to manually alter the template in the datastore if we wanted to add the "trusted" flag after creation. If we add an UPDATE endpoint for templates, the "trusted" flag would be tied to a specific template version, not the general template itself. 
We would definitely need to keep track of this "trusted" information in a separate table.

### 3. Should we have "trusted" template or namespace?
[...]

PR #1322, opened 2018-10-05
feat(1271): Template list should handle extra query options [4]

tkyi:
## Context

In order to support the new template design, we'll need these additional endpoints:
1. Searching through multiple fields for templates: `GET /templates?search={keyword}&sortBy=name&sort=ascending&page={page}&count={count}`
2. Listing all distinct namespaces in templates: `GET /templates?distinct=namespace&sort=ascending&sortBy=namespace`

## Objective

This PR allows additional query options to be passed in to handle those new endpoints.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/1271
Blocked by https://github.com/screwdriver-cd/datastore-sequelize/pull/35, https://github.com/screwdriver-cd/models/pull/293


plugins/templates/list.js
```diff
@@ -4,6 +4,9 @@ const boom = require('boom');
 const joi = require('joi');
 const schema = require('screwdriver-data-schema');
 const listSchema = joi.array().items(schema.models.template.get).label('List of templates');
+const distinctSchema = joi.string().label('Field to return unique results by');
```
minz1027: this can be `joi.string().valid(Object.keys(schema.models.template.base))` so that it's always valid field

PR #1327, opened 2018-10-08
fix(1271): Distinct returns array of objects

tkyi:
## Context

Datastore calls `decodeFromDialect` https://github.com/screwdriver-cd/datastore-sequelize/blob/master/index.js#L19, converting the result to an array of objects.

## Objective

This PR changes validation to handle the result properly.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/1271


DekusDenial:
the test should have passed.....retrying.....

Issue #1117, opened 2018-06-12
Periodic build does not start if a user suspended the scm permission


yoshwata:
**What happened**:
Periodic build does not start because the redis's information of admins is not correct.

**What you expected to happen**:
Periodic build starts even if an admin user is suspended.

**How to reproduce it**:
1. `UserA` makes a periodic build pipeline.
2. Suspend `UserA`'s permission of the repository.
3. Wait the periodic build starts, (but never starts)

**Idea to solve it**:
Add api to get calculated `admins`, and call it at [here](https://github.com/screwdriver-cd/executor-queue/blob/v2.2.3/index.js#L166)

PR #1330, opened 2018-10-10
[2]fix(1117): Pass pipelineFactory to executor when setup executor

kumada626:
## Context

The periodic build does not start in the case of the pipeline admin user was removed after periodic build made.
To fix this problem, executor-queue use pipelineFactory. Therefore, API fixed to pass pipelineFactory to executor when setup executor.

## Objective

Fixed to pass pipelineFactory to executor when setup executor.

## References

issue: #1117 
executor-queue PR: https://github.com/screwdriver-cd/executor-queue/pull/33


catto:
Note: coveralls never updated the status even if there are some builds for this PR. I'm going to merge this PR because I've confirmed that the coverage remains the same value as before.
https://coveralls.io/builds/19462780

Issue #1079, opened 2018-05-22
Generate and Use Deploy Keys


stjohnjohnson:
Currently, most CI/CD services today will automatically add a R/O Deploy key to your repository and use those on checkout.

Screwdriver does not.  Would it be possible to support this?

--
UPDATE: https://github.com/screwdriver-cd/screwdriver/issues/1079#issuecomment-658256382

jithin1987:
Doesn't user has to generate the Deploy key  and make it available to the CI/CD tool ?

What Screwdriver could do is - 
1. Define a secret KEY for the Deploy key. User must add this in secret for the pipeline,
2. Modify `scm-github` to use that KEY if available, and use it for ssh based checkout.

stjohnjohnson:
Nah, if you use anything like CircleCI, Werker, Codeship, etc. it will generate and add the deploy key for you via [API](https://developer.github.com/v3/repos/keys/#add-a-new-deploy-key).

jithin1987:
We could use this to support private repositories without having wide Oath access. Ask user to provide a Deploy Key and use that for checking out Private repository. So flow will be.

1. Use generates Deploy key at Github.
1. As part of create pipeline take this input and store the KEY in Secrets and then use this for git operations.
1. If user wants to move to a Deploy key after creating pipeline, then add the key at Pipeline -> Options
1. Modify `scm-github` to use that KEY if available, and use it for ssh based checkout.

jithin1987:
We can automate generation part later on, but I guess for a CI/CD tool to do that, we should  be asking Oauth scope for adding keys.

stjohnjohnson:
You could skip 1-3 and just take it as a secret.

jithin1987:
> You could skip 1-3 and just take it as a secret. For the short-term

That's my first suggestion https://github.com/screwdriver-cd/screwdriver/issues/1079#issuecomment-393631839

gthosaniNetskope:
@stjohnjohnson, @jithin1987 wondering if there was any ETA for this issue ?

supra08:
Hi @jithine @tkyi ! I would like to work on this as my GSoC 2020 project. I am able to comprehend the goal from this issue as well as #1415. I would like to know if there are some more points to it that I should consider before drafting the proposal.

jithine:
[...]

PR #1331, opened 2018-10-10
Private repositories: add design document

russdeluca:
## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/1079

### DIFF
--- a/design/private-repos.md
+++ b/design/private-repos.md
@@ -11,30 +11,55 @@ This documentation covers the design proposal for that new feature.
 
 ## The Problem
 
-The current authentication methods do not provide access to private repositories.  
-Also, there are several stages of a pipeline in which different access keys are needed 
-when dealing with private repositories.
-This document will outline each stage and how it will be addressed.
+The current authentication methods do not provide access to private repositories.  This includes
+GitHub API calls as well as git clones via SSH.
+
+The OAuth token provided today for GitHub API calls is limited in scope to only public repositories.  If this scope were to 
+be expanded, then all users (whether they need this expanded scope or not) would be granting Screwdriver
+a security scope larger than required.  The screwdriver app should only be granted scope
+that is required.
+There are several stages of a Screwdriver pipeline in which this expanded scope is required
+when dealing with private repositories.  This document will outline each stage and how it will be addressed.
+
+Additionally, the Screwdriver app only has access to clone public repositories with the ssh key
+currently being used by Screwdriver.
 
 ## Background
 
 Reference [Issue 1079][issue1079]
 
 ## Proposal
+Any user intending to use Screwdriver with private repositories must perform the following prerequisites:
+- create a Personal Access Token within the Developer Settings of their Github user profile
+and include that token into their Screwdriver user profile.
+	- Github Personal Access Token is created at: https://github.com/settings/tokens/new
+	- Screwdriver user profile is edited at:  /user-settings
+- create/obtain an ssh keypair 
+	- store the public key within the Deploy Keys section of the private repository settings 
+	- associate the private key with the Screwdriver pipeline associated with this private repository
 
-Pipeline creation:
-- User must provide a GitHub Deploy Key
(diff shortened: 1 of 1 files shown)


design/private-repos.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Background
+
+Reference [Issue 1079][issue1079]
+
+## Proposal
+
+Pipeline creation:
+- User must provide a GitHub Deploy Key
+- User must provide a GitHub Personal Access Token
```
minz1027: Can this be the github Oauth token that we already store in the user table?

jimgrund: This will need to be a token that has elevated privileges for access to private repos.  The current token only allows public repos.

design/private-repos.md
```diff
@@ -0,0 +1,72 @@
...
+- create a Personal Access Token within the Developer Settings of their Github user profile
+and include that token into their Screwdriver user profile.
+	- Github Personal Access Token is created at: https://github.com/settings/tokens/new
+	- Screwdriver user profile is edited at:  /user-settings
+- create/obtain an ssh keypair 
+	- store the public key within the Deploy Keys section of the private repository settings 
+	- associate the private key with the Screwdriver pipeline associated with this private repository
+
+Private Repository Pipeline Creation:
+1 A GitHub Personal Access Token that allows control of private repositories must be associated to the user profile within Screwdriver.
```
jithin1987: One concern here is that this Personal Access Token has wide permission and can do anything including delete user stuff. We are back with original concern of preventing SD having access to a user token which can do pretty much anything on behalf of user.

design/private-repos.md
```diff
@@ -0,0 +1,72 @@
...
+1 User indicates on the Create form that the repository is a private repository which expands the web form to ask for additional information.
+1 User adds the private ssh key associated with the private repository to the Pipeline Create webform.
+
+The private ssh key will be stored as a Secret on the pipeline at pipeline creation time.  This Secret will be named [TBD]. 
+The existence of the secret with name [TBD] will essentially be a flag that indicates that this is a
+private repository requiring special handling.
+- When Screwdriver attempts to perform API calls for a private repository pipeline, it will use this flag to know to use the 
+GitHub Personal Access Token associated with the user logged in.
```
jithin1987: Can we replace this access check with something else ? Rather then current check against a repository is it possible to replace this check with say isUserAdmin() at a GIT org level ? or even skip it if deploy keys are available.

This can be straight forward if it's a users private repo Eg: `jithine/private-repo` we can just check if logged in user is same as the SCM org.
But tricky when it's a teams organization `team/private`. Can we explore if there is any hook which GIT offers which ties permission check at an organization level ?

Alternatively if user does provide a deploy key, we could skip all checks which ties back to SCM and let it pass or fail at code checkout.

Idea is to see if we can get away without doing Personal Access Token.

design/private-repos.md
```diff
@@ -0,0 +1,72 @@
...
+- When Screwdriver attempts to clone the repository for a private repository pipeline, it will use this flag to know to use the
+private ssh key that was associated to the pipeline.
+
+Screwdriver features that are affected by these changes are:
+- Creating/updating the pipeline (API calls)
```
jithin1987: We should spike if we can get away with having to call github at all for permission if we have deploy keys.

design/private-repos.md
```diff
@@ -0,0 +1,72 @@
...
+- Creating/updating the webhook (API calls)
+- Managing secrets (API calls)
+- Executing a build job (API calls and Cloning repo)
+
+Anticipated problem not resolved by this implementation:
+- A Pipeline leveraging Repo and multiple private repositories would not work with the above design.
+GitHub requires a unique public SSH key per repository and the above design only supports defining one private key per Pipeline.
```
jithin1987: this is the intended behavior, user should generate deploy keys for each pipeline.

tkyi:
Closing this PR for now since we aren't planning to work on it anytime soon.

PR #1343, opened 2018-10-16
fix: Initialize cache bookend

pranavrc:
Related PR: https://github.com/screwdriver-cd/cache-bookend/pull/1


bin/server
```diff
@@ -43,6 +43,10 @@ authConfig.scm = scm;
 const ArtifactPlugin = require('screwdriver-artifact-bookend');
 const artifact = new ArtifactPlugin(ecosystem.store);
 
+// Build cache plugin
+const CachePlugin = require('screwdriver-cache-bookend');
+const cache = new CachePlugin(ecosystem.api);
```
d2lam: There is no `ecosystem.api`, do you mean `httpd.uri`

pranavrc: This one? https://github.com/screwdriver-cd/screwdriver/blob/e0e2b8a998abc949585fee4f3f073f181bf80c83/bin/server#L24

PR #1328, opened 2018-10-09
Branch specific test

yokawara:
## Context
Step definitions for workflow is not implemented yet.

## Objective
Implement step definitions for workflow features.
As we should make spec of PR branch filtering features more clarify, they are now disabled.

## Reference
Before merging this PR, https://github.com/screwdriver-cd-test/functional-workflow/pull/1 have to be merged.

PR #1348, opened 2018-10-18
fix: Fix workflow test

kumada626:
## Context

Functional test of workflow still frequently fails after this PR #1345 .
To get build info certainly, I use `searchForBuild` instead of `findEventBuilds`.
## Objective

Use `searchForBuild` instead of `findEventBuilds` in `the "(.*)" job is triggered` and `that "(.*)" build uses the same SHA as the "(.*)" build` steps.

## References

relates: #1328 #1345


yoshwata:
lint error?

PR #1376, opened 2018-11-02
feat: expose SameSite Cookie option

jithin1987:
## Context

`SameSite` cookie config provides strict CSRF protection, but it breaks CORS interaction for our internal usecases. Hence providing an option for cluster admins to set this value.

## Objective
1. Upgrade `hapi-auth-cookie` to v7  which supports `isSameSite` setting.
2. Provide a config to set this value.

## References

1. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#Directives See `SameSite` section

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -34,6 +34,8 @@ auth:
   admins: []
   # Default session timeout (in minutes)
   sessionTimeout: 120
+  # SameSite Cookie Option
+  sameSite: false
 
 httpd:
   # Port to listen on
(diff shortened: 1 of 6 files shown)


config/default.yaml
```diff
@@ -34,6 +34,8 @@ auth:
   admins: []
   # Default session timeout (in minutes)
   sessionTimeout: 120
+  # SameSite Cookie Option
+  sameSite: false
```
minz1027: should this be the default option? we can set the default as `strict` and relax it by COOKIE_SAME_SITE.

jithin1987: Done

Issue #1319, opened 2018-10-02
Multiple Build Clusters


jithin1987:
## Context

This is applicable for a Scalable Screwdriver instance which makes use of [executor-queue](https://github.com/screwdriver-cd/executor-queue/)

Screwdriver builds are executed via a single queue worker which reads the job configurations from Queue and uses appropriate executor to run execute the build.

This model can be expanded such that Screwdriver can send the build information to multiple build clusters which provides following capabilities.

1. Running at least two build clusters which allows us to take down a build cluster for critical infrastructure updates.
1. Enables Screwdriver users to potentially bring their own build clusters so that they can have specialized hardware for specific needs.
1. Enables Screwdriver to support a [build cluster in AWS which can leverage power of EC2 ](https://github.com/screwdriver-cd/screwdriver/issues/1243)cloud compute resources

## Objective

1. Create a Scheduler service which fronts Screwdriver build Queue
    1. Workers will no longer connect directly to Queue to get new builds.
    1. Scheduler Service is responsible for distributing builds to Workers running in different build clusters.
    1. Scheduler service should be able to authorize Worker service and expose only the Job information which are applicable to the build cluster.
    1. Scheduler should create as many build queues as the number of registered Build Clusters. 
1. Register a build cluster
    1. Cluster admin should be able to register their build cluster with Screwdriver. Registration should take in information such as cluster name, queue name which should be created for the cluster, Authorization means (Eg: JWT Public key for the cluster) and SCM Organizations where pipelines must belong to in order to use this cluster .
1. Implicit and Explicit build clusters
   1. Users may or may not specify a specific build cluster in their Screwdriver pipeline config
   1. If user has not specified any build cluster information (implicit), Screwdriver API should assign any of the generic build clusters applicable to the build. (Default supported build cluster(s))
   1. If user has specified a build cluster in the Screwdriver pipeline config, then Screwdriver API should enqueue the job in the build queue applicable to the build cluster
      1. If it's a specialized build cluster, (Screwdriver customer provided) then Screwdriver API must validate if the owner of the pipeline is Authorized to use the cluster for user builds. We can tie this Authorization to the SCM Organization which is captured as part of registering a cluster.  
1. Document design at https://github.com/screwdriver-cd/screwdriver/tree/master/design

## References

1. Worker - https://github.com/screwdriver-cd/queue-worker

minz1027:
### Implementations
[...]

PR #1380, opened 2018-11-10
fix(1319): Fix build clusters endpoints

tkyi:
## Context

Since we're calling a list() instead of a get(), it returns and array instead of a since build cluster.

## Objective

This PR accounts for that array.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/1319

### DIFF
--- a/plugins/buildClusters/get.js
+++ b/plugins/buildClusters/get.js
@@ -23,20 +23,24 @@ module.exports = () => ({
             }
         },
         handler: (request, reply) => {
+            const name = request.params.name;
             const factory = request.server.app.buildClusterFactory;
             const config = {
                 params: {
-                    name: request.params.name
+                    name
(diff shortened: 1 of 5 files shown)


plugins/buildClusters/get.js
```diff
@@ -23,20 +23,24 @@ module.exports = () => ({
...
-                    name: request.params.name
+                    name
                 }
             };
 
             return factory.list(config)
                 .then((model) => {
-                    if (!model) {
-                        return reply(boom.notFound('Build cluster does not exist'));
+                    if (!Array.isArray(model)) {
```
minz1027: No need for this check, we have it in model already.
https://github.com/screwdriver-cd/models/blob/master/lib/baseFactory.js#L154

Issue #1385, opened 2018-11-15
Build Cache Invalidation


d2lam:
In certain cases, users want to invalidate the cache such as when secrets are accidentally leaked. 
This feature should address two use cases:
- Users are able to invalidate the cache from the UI. The cache will be removed from store. 
- Users are able to start builds with a clean cache. 

Related: https://github.com/screwdriver-cd/screwdriver/issues/1257

d2lam:
Back end work is done. Cache can now be cleaned using user's token, as long as user has access to the pipeline. Available endpoints:
DELETE ${STORE_API}/v4/caches/pipelines/${pipelineID}
DELETE ${STORE_API}/v4/caches/events/${eventID}
DELETE ${STORE_API}/v4/caches/jobs/${jobID}

Need work on UI part, maybe @DekusDenial can help with the UI when you have time?

PR #1387, opened 2018-11-16
feat: check if a user is admin of a pipeline

d2lam:
Need this endpoint so that we can call from the `store` to invalidate the cache.

Related:
https://github.com/screwdriver-cd/screwdriver/issues/1385
https://github.com/screwdriver-cd/store/pull/67


plugins/pipelines/isAdmin.js
```diff
@@ -0,0 +1,54 @@
+'use strict';
+
+const boom = require('boom');
+const joi = require('joi');
+const schema = require('screwdriver-data-schema');
+const idSchema = joi.reach(schema.models.pipeline.base, 'id');
+
+module.exports = () => ({
+    method: 'GET',
+    path: '/pipelines/{id}/isAdmin',
```
tkyi: wonder if `isAdmin` is confusing, since you can be a Screwdriver admin and/or a pipeline admin
should be okay since the route is prefaced by `/pipeline`

d2lam: Changing the endpoint since we need it for jobs and events as well

PR #1395, opened 2018-11-26
fix: fix banner tests

tk3fftk:
Some banner tests are failed but these are uncaught.
This PR catches and passes them.


test/plugins/banner.test.js
```diff
@@ -141,29 +141,24 @@ describe('banner plugin test', () => {
...
             server.inject(options).then((reply) => {
-                const expected = Object.assign({}, testBanner);
+                const expected = Object.assign({}, options.payload);
 
-                delete expected.id;
-                delete expected.createTime;
+                expected.createdBy = options.credentials.username;
                 assert.calledWith(bannerFactoryMock.create, expected);
                 assert.equal(reply.statusCode, 201);
                 assert.equal(reply.result, testBanner);
```
tkyi: is it better to use `assert.deepEqual` here?

PR #1397, opened 2018-11-28
fix: Use toolbox script for deploying

tkyi:
## Context

The beta deploy job keeps failing even though the deployment is successful. This is because the script we are using is grepping for `200 OK` in the response, which is compatible when using HTTP/1.1. However, HTTP/2 only returns `HTTP/2 200`, which fails the script.

## Objective

This PR switches to using the toolbox script for deploying, which solves this issue already.

## References
- Toolbox: https://github.com/screwdriver-cd/toolbox/blob/master/k8s-deploy.sh#L30
- Original script: https://gist.github.com/stjohnjohnson/3d2388b2a7ba658cdcdaffa8cd874e50#file-k8s-deploy-sh-L30

### DIFF
--- a/screwdriver.yaml
+++ b/screwdriver.yaml
@@ -41,7 +41,7 @@ jobs:
     beta:
         requires: [publish]
         steps:
-            - setup-ci: git clone https://gist.github.com/3d2388b2a7ba658cdcdaffa8cd874e50.git ci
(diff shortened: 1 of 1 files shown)


tkyi:
@DekusDenial That might be a good improvement for the toolbox

PR #1401, opened 2018-11-30
feat(1271): add compact query to templates and commands

DekusDenial:
## Context

API call to `/templates` is way too slow for the node.js layer to serialize all the bloated json payload

## Objective

- Apply the same compact query optimization to `templates` and `commands` to speed up the API performance by over 95%
- Trimmed the fat off the response payload
- Dedupe multi versions by grouping on `namespace` and `name`

## References

related to: screwdriver-cd/screwdriver#1271


plugins/templates/list.js
```diff
@@ -58,18 +67,23 @@ module.exports = () => ({
...
                 };
             }
 
-            if (request.query.page || request.query.count) {
-                config.paginate = {
-                    page: request.query.page,
-                    count: request.query.count
-                };
+            // check if the call wants compact data
+            if (compact === '1') {
```
tkyi: why `1` vs `true`/`false`?

tkyi:
Some banner test failures probably due to https://github.com/screwdriver-cd/data-schema/pull/295/files

plugins/templates/list.js
tkyi: I don't have a strong preference either way
I think we usually stick to `true`/`false`

DekusDenial: i can change

DekusDenial: kk, will change

DekusDenial:
@tkyi i can remove those tests since they are no longer relevant (due to the change that removed the defaults from the schema)

Issue #1246, opened 2018-08-09
PR authors should be able to restart their own PR builds


anilkumarmyla:
**What happened**:
PR authors who do not have push permissions to repo cant restart PR builds

**What you expected to happen**:
PR authors should be able to restart their own PR builds

tkyi:
Was able to rerun my PR build against Dao's repo: https://cd.screwdriver.cd/pipelines/173/builds/83878

jithin1987:
This now works as expected.

PR #1414, opened 2018-12-14
feat(1246): Allow PR author to run their own PR build

tkyi:
## Context

PR authors should be able to rerun their own builds, even if they don't have access to the repo.

## Objective

This PR allows PR authors to start events on their own PR.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/1246
Blocked by bug fixed in https://github.com/screwdriver-cd/scm-base/pull/64


plugins/events/create.js
```diff
@@ -88,54 +117,75 @@ module.exports = () => ({
...
-                                // This is needed to make admins dirty and update db
-                                pipeline.admins = newAdmins;
-
-                                return pipeline.update()
-                                    .then(() => {
-                                        throw boom.unauthorized(`User ${username} `
-                                        + 'does not have push permission for this repo');
-                                    });
+                            if (!permissions.push && !prNum) {
+                                return updateAdmins({ pipeline, isAdmin: false, username });
```
minz1027: feel the `updateAdmins` can just take in the `permissions` object. And for the non-pr case, it can be all done here in this block. then you don't need the part in the end: https://github.com/screwdriver-cd/screwdriver/blob/b81432caaae7b4f467ccf3cb0c4086cc6d29ba08/plugins/events/create.js#L180

```js
if (!prNum) {
     return updateAdmins({ pipeline, permissions, username });
}
```

minz1027: same for the pr case

plugins/events/create.js
```diff
@@ -88,54 +123,63 @@ module.exports = () => ({
                         throw boom.unauthorized('Token does not have permission to this pipeline');
                     }
 
+                    let token;
+                    let scmConfig;
+                    let permissions;
+                    let prInfo;
```
minz1027: token and prInfo shouldn't need to be let and defined here

jithin1987:
Does [this PR ](https://github.com/screwdriver-cd/screwdriver/pull/1420) of pinning scm-github affect this one ?

tkyi:
@jithin1987 yes, need to unpin `scm-github` version in this PR

Issue #1355, opened 2018-10-18
Feature to comment back to Git Pull Request from PR build


tijuj:
**meta_set summary.{key} doesn't automatically add PR comments**:

**Earlier**:

setting meta.summary.{key} as below would add a PR comment 

```
pull-request:
        settings:
            DEPLOY: k8s
            sd_pr_comments: 'true'
        steps:
            postdeploy:
              description: "add instance to PR"
              command: |
                meta_set meta.summary.Instance `meta_get meta.instances.k8s`
                meta_set meta.summary.Build `meta_get meta.build`
                meta_set meta.summary.SHA "$_pull_request_sha"
```

would automatically add PR comments like -
<img width="786" alt="screen shot 2018-10-18 at 3 21 48 pm" src="https://user-images.githubusercontent.com/3414143/47187599-a73ff680-d2e9-11e8-99df-55dccbcbbdec.png">

It doesn't work anymore. 
Is there another way to do this? This feature was really helpful for code reviews and QA.

tkyi:
I think this feature is feasible.
---
In V4, we'll skip needing to set `sd_pr_comments: 'true'`. To use the meta summary feature, you just need to `meta set` to the `summary` key like this:
```
main:
        steps:
            - postdeploy: |
                meta set summary.Instance `meta get instances.k8s`
                meta set summary.Build `meta get meta.build`
                meta set summary.SHA "$_pull_request_sha"
```

jithin1987:
@tijuj This feature is live now.

tijuj:
thank you @jithin1987 @tkyi

PR #1433, opened 2019-01-07
fix(1355): Add bot token comment for scm PR comments

tkyi:
Adding comment in config yamls for bot token for PR commenting.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/1355

### DIFF
--- a/config/custom-environment-variables.yaml
+++ b/config/custom-environment-variables.yaml
@@ -265,6 +265,8 @@ scms:
   #     # The username and email used for checkout with github
   #     username: SCM_USERNAME
   #     email: SCM_EMAIL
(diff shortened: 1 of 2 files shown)


config/custom-environment-variables.yaml
```diff
@@ -265,6 +265,8 @@ scms:
   #     # The username and email used for checkout with github
   #     username: SCM_USERNAME
   #     email: SCM_EMAIL
+  #     # Token for writing PR comments in Github, needs public_repo scope
+  #     token: A-BOT-GITHUB-PERSONAL-ACCESS-TOKEN
```
tk3fftk: I think this may be using `_` instead of `-`.

Issue #1412, opened 2018-12-10
Build Metrics


jithin1987:
## Context

A Screwdriver build generates metrics pertaining to a build time and status of it's individual steps. This information is presented in build detail page where information on total build time and how much time each individual step took is displayed in Build detail page of SD UI. Pipeline detail page of SD UI  shows information on total build time for an event by aggregating build times across all the jobs which was executed for that event. However SD lacks an aggregated build metrics view at a pipeline level. This Pipeline Metrics view should help a user to look at a pipeline and quickly get views on various build trends.

### Metrics at Pipeline scope

1. Trend of commit/pr events at pipeline scope.
    1. Time series graph where each point represents time taken for the event as a whole.
    1. Data points should show information about event and a link to the events page which loads that specific event.
1. Trend of builds at pipeline scope.
    1. Time series graph where each point represents time taken for a specific build across all events of the pipeline.
   1. Data points should show information about the build (eg: Job) and a link to the specific build. 
1. Trend of build steps at pipeline scope.
    1. Time series graph where each point represents time taken for a specific step across all the builds for the pipeline.
    1. Data points should show information about the build to which the step belongs to and a link to the specific build's step. 

### Metrics at Pipeline -> Job scope

This is the view user will see after filtering out a specific job in the pipeline

1. Trend of builds at job scope
    1. Time series graph where each point represents time taken for a specific build of the job in the pipeline.
   1. Data points should show information about the build (eg: Job, Event) and a link to the specific build. 
1. Trend of build steps at job scope.
    1. Time series graph where each point represents time taken for a step  across all the builds of the scoped job.
    1. Data points should show information about the build to which the step belongs to and a link to the specific build's step. 

### Metrics at Pipeline -> Job -> Step scope

This is the view user will see after filtering out a specific step in a job in the pipeline

1. Trend of build steps at job scope.
    1. Time series graph where each point represents time taken for a specific step of the build across the same job.
    1. Data points should show information about the build to which the step belongs to and a link to the specific build's step. 

<del>
### Metrics at Collection scope

1. View should be similar to pipeline scoped metrics.
1. View will involve events, builds and steps spread across multiple pipelines.
1. Data points should show information about Pipeline, event and build to which it belongs to.
</del>

## Objective

1. Design for a data store for storing aggregated views.
1. Store and aggregate build metric data for building Pipeline, Job and Collection scoped views.
1. UI to display the time series data about build metrics.
[...]

PR #1466, opened 2019-01-31
fix(1412): return type for metrics

d2lam:
No need to make it toJSon again. 

Blocked by: https://github.com/screwdriver-cd/models/pull/323
Related: screwdriver-cd/screwdriver#1412


plugins/pipelines/metrics.js
```diff
@@ -35,13 +35,13 @@ module.exports = () => ({
...
                 })
-                .then(metrics => reply(metrics.map(m => m.toJson())))
+                .then(metrics => reply(metrics))
                 .catch(err => reply(boom.boomify(err)));
         },
         validate: {
             query: joi.object({
-                startTime: joi.date().iso(),
-                endTime: joi.date().iso()
+                startTime: joi.string().isoDate(),
```
tkyi: are these both required?

Issue #1468, opened 2019-02-03
Performance issue of webhook


catto:
**What happened**:
I have a big github repository that has thousands of pull requests in it. When I open a new pull request in that repository, it takes tens of seconds to process webhook and sometimes ends with timeout by github webhook service (30s).

**What you expected to happen**:
Each webhook is processed in a few seconds.

**How to reproduce it**:
1. Create so many PRs in your repository. It doesn't matter keep them open or close
1. Open new PR or push new commit to the PR

**References**
1. https://github.com/screwdriver-cd/screwdriver/issues/1769
1. https://github.com/screwdriver-cd/screwdriver/issues/1737

catto:
We should check if a result of listFiles API includes the desired SHA or not instead of putting 1.8sec wait so that webhook endpoint will be faster for most cases and every PR build ensures the latest changed files are reflected to github.
https://developer.github.com/v3/pulls/#list-pull-requests-files

catto:
No.2 is being fixed via #1603

jithin1987:
@catto one other issue I see is wrt `scm.getBranchList()` https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/webhooks/index.js#L174

We have that for branch filtering feature for identifying triggered pipelines. However we could refactor it such that it does not reach out to Git and instead query our database for triggered pipelines.

jithin1987:
https://github.com/screwdriver-cd/screwdriver/pull/1643/files addresses my previous comment

jithin1987:
Last work has been done at https://github.com/screwdriver-cd/screwdriver/pull/1653/files which addresses https://github.com/screwdriver-cd/screwdriver/issues/1468#issuecomment-460035553

catto:
Reopening the issue because "3. magic wait for 1.8 secs every time" I wrote is probably fixable now. Also sometimes I see issues like screwdriver.yaml not found because of magic branch is not updated. #593
Let's replace 1.8 sec wait with polling pull_request API.
https://developer.github.com/v3/git/#checking-mergeability-of-pull-requests

tanoda:
according to GitHub doc,
* if `mergeable` is `null`, wait a minute and retry
* if `mergeable` is `true`, we can use `merge` refs to get yaml.
* if `mergeable` is `false`, ... maybe we want to use pipeline branch to get yaml?

jithin1987:
[...]

PR #1487, opened 2019-02-09
fix(1468): improve performance of webhook plugin

catto:
## Context

Webhook plugin processes Pull Request events inefficiently. #1468 
It causes timeout of webhook from github
image

## Objective

* Removes unnecessary `pipeline.sync` operation from webhook plugin.
  * `pipeline.sync` takes over 10 seconds to complete depending on pipeline. It runs twice currently. Second one runs by `eventFactory`, so the first one should be removed.
* Reduces duplicated data fetches.
  * `parseUrl` and `getChangedFiles` should be called once and use the result in later process.
  * `pullRequestEvent` and `pushEvent` will receive them as `scmConfig`
* Calls `promiseToWait` and `obtainScmToken` in parallel
* Rewrite of `pullRequestEvent` with async/await

## References

* Original issue: #1468
* eventFactory calls pipeline.sync: https://github.com/screwdriver-cd/models/blob/d294a9e0d013e4ffcfcac7a80e701449843009f9/lib/eventFactory.js#L356-L383

### DIFF
--- a/plugins/webhooks/index.js
+++ b/plugins/webhooks/index.js
@@ -203,9 +203,8 @@ async function pullRequestOpened(options, request, reply) {
     const { hookId, prSource, pipeline } = options;
 
     if (pipeline) {
-        const p = await pipeline.sync();
         // @TODO Check for cluster-level default
-        const restriction = p.annotations['beta.screwdriver.cd/restrict-pr'] || 'none';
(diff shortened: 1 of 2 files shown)


catto:
Decreasing 0.002% is caused by removing a few lines. Covered lines are not changed.

jithin1987:
@catto we missed following up on this PR. What's the status of this ?

catto:
Closing due to out of date

PR #1494, opened 2019-02-14
fix: allow update build status to COLLAPSED and FROZEN

minz1027:
-  allow update build status to COLLAPSED and COLLAPSED
- set init step endTime and exit code for COLLAPSED and COLLAPSED, otherwise from the UI it may seem like the init step is still going


minz1027:
@pranavrc no need to add it there, those are to restrict that, you can only change status if current status are one of `'RUNNING', 'QUEUED', 'BLOCKED', 'UNSTABLE'`. `COLLAPSED` and `FROZEN` will be changed from `QUEUED` and won't change afterwards.

Issue #1447, opened 2019-01-23
`[skip ci]` should create an empty event


jithin1987:
**What happened**:

If user uses `[skip ci]` in commit title, screwdriver ignores the webbook completely. This leads to a situation where events list does not reflect latest commit. And if a user manually starts a job from latest event, it's not evident that it's from a previous commit. 

**What you expected to happen**:

1. `[skip ci]` should create an event with out running any jobs
1. User should then be able to interact with the empty pipeline graph to start jobs with the latest commit SHA.

**How to reproduce it**:

1. Make a commit with `[skip ci]`
1. Run a job from latest commit.

PR #1512, opened 2019-02-21
feat(1447): skip ci creates empty event

d2lam:
Pass `skipMessage` to models if commit message includes `skip ci`
This only applies to push events, pull request events still work like normal

Related: https://github.com/screwdriver-cd/screwdriver/issues/1447
Blocked by: https://github.com/screwdriver-cd/models/pull/331

### DIFF
--- a/plugins/webhooks/index.js
+++ b/plugins/webhooks/index.js
@@ -134,12 +134,13 @@ async function triggeredPipelines(pipelineFactory, scmConfig, branch, type) {
  * @param  {Array}        options.changedFiles  List of changed files
  * @param  {String}       options.branch        The branch against which pr is opened
  * @param  {String}       options.action        Event action
+ * @param  {String}       [options.skipMessage] Skip message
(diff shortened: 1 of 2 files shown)


plugins/webhooks/index.js
```diff
@@ -595,10 +604,7 @@ exports.register = (server, options, next) => {
                     request.log(['webhook', hookId], `Received event type ${type}`);
 
                     if (/\[(skip ci|ci skip)\]/.test(parsed.lastCommitMessage)) {
-                        message = 'Skipping due to the commit message';
-                        request.log(['webhook', hookId], message);
```
tkyi: are we getting rid of this log statement?

d2lam: hmm since we are not skipping this webhook path anymore, we should skip it right?

tkyi: oh I see
I thought the message would be lost
nvm!

PR #1514, opened 2019-02-22
test pr

minz1027:


minz1027:
github sends 2 request to each webhook probably it took us too long to return.

PR #1525, opened 2019-02-28
Update admin on webhook

yoshwata:
## Context

Our cluster has many request by no admin exists recently. 

Admin add/delete function is implemented on start or sync from UI.  Webhook is also a function that user frequently use, so I'd like to port it on webhook.

## Objective

Users are registered as pipeline's admin when they trigger from github push or open PR.

## References

https://github.com/screwdriver-cd/screwdriver/pull/1106

### DIFF
--- a/plugins/webhooks/index.js
+++ b/plugins/webhooks/index.js
@@ -53,10 +53,6 @@ async function checkAndUpdateAdmins(userFactory, username, scmContext, pipeline)
     try {
         await userFactory.get({ username, scmContext })
             .then((user) => {
-                if (!pipeline) {
-                    throw boom.notFound('Pipeline does not exist');
-                }
-
                 if (!user) {
                     throw boom.notFound('User does not exist');
                 }
--- a/plugins/webhooks/index.js
+++ b/plugins/webhooks/index.js
@@ -14,33 +14,29 @@ const WAIT_FOR_CHANGEDFILES = 1.8;
  * @param  {String}    username     Username of user
  * @return {Promise}                Updates the pipeline admins and throws an error if not an admin
  */
-function updateAdmins(permissions, pipeline, username) {
-    const newAdmins = pipeline.admins;
-
-    // Delete user from admin list if bad permissions
-    if (!permissions.push) {
-        delete newAdmins[username];
-        // This is needed to make admins dirty and update db
-        pipeline.admins = newAdmins;
-
-        return pipeline.update()
-            .then(() => {
-                throw boom.forbidden(`User ${username} `
-                + 'does not have push permission for this repo');
-            });
-    }
-
-    // Add user as admin if permissions good and does not already exist
-    if (!pipeline.admins[username]) {
-        newAdmins[username] = true;
-        // This is needed to make admins dirty and update db
-        pipeline.admins = newAdmins;
-
-        return pipeline.update();
-    }
-
-    return Promise.resolve();
-}
+// function updateAdmins(permissions, pipeline, username) {
+//     const newAdmins = pipeline.admins;
+//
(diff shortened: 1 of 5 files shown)


plugins/webhooks/index.js
```diff
@@ -2,10 +2,76 @@
...
+ * @param  {UserFactory}    userFactory     UserFactory object
+ * @param  {String}         username        Username of user
+ * @param  {String}         scmContext      Scm which pipeline's repository exists in
+ * @param  {Pipeline}       pipeline        Pipeline object
+ */
+async function checkAndUpdateAdmins(userFactory, username, scmContext, pipeline) {
+    try {
+        await userFactory.get({ username, scmContext })
+            .then((user) => {
+                if (!pipeline) {
```
tk3fftk: Is this check needed?

plugins/webhooks/index.js
```diff
@@ -2,10 +2,76 @@
...
+/**
+ * Update admins array after check existance of pipeline, user and permission
+ * @param  {UserFactory}    userFactory     UserFactory object
+ * @param  {String}         username        Username of user
+ * @param  {String}         scmContext      Scm which pipeline's repository exists in
+ * @param  {Pipeline}       pipeline        Pipeline object
+ */
+async function checkAndUpdateAdmins(userFactory, username, scmContext, pipeline) {
+    try {
+        await userFactory.get({ username, scmContext })
```
tk3fftk: I think the above example is more simple when we use `await`.
```
try{
  const user = await userFactory.get({ username, scmContext });
  const userPermissions = await user.getPermissions(pipeline.scmUri);

  return updateAdmins(
                userPermissions,
                pipeline,
                username));

} catch (err) {...}
```

test/plugins/webhooks.test.js
```diff
@@ -502,15 +509,75 @@ describe('github plugin test', () => {
                 });
             });
 
-            it('returns 500 when failed', () => {
-                eventFactoryMock.create.rejects(new Error('Failed to start'));
+            it('reuturns 200 when user is registered as admin', () => {
```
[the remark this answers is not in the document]
yoshwata: I fixed, thank you.

plugins/webhooks/index.js
yoshwata: No need to do that, I fixed. Thank you.

plugins/webhooks/index.js
yoshwata: It's nice idea. I fixed. Thank you.

plugins/webhooks/index.js
```diff
@@ -2,10 +2,48 @@
...
+    try {
+        const user = await userFactory.get({ username, scmContext });
+        const userPermissions = await user.getPermissions(pipeline.scmUri);
+        const newAdmins = pipeline.admins;
+
+        // Delete user from admin list if bad permissions
+        if (!userPermissions.push) {
+            delete newAdmins[username];
+            // This is needed to make admins dirty and update db
+            pipeline.admins = newAdmins;
```
tk3fftk: `pipeline.admins = newAdmins;` and `return pipeline.update();` are duplicated with the below `if` statement. We can aggregate these.

plugins/webhooks/index.js
```diff
@@ -421,14 +459,14 @@ function pullRequestEvent(pluginOptions, request, reply, parsed) {
                         restrictPR
                     };
 
+                    updateAdmins(userFactory, username, scmContext, pipeline);
```
wahapo: We do not need await here? I am worried about when `updateAdmins` will be completed.

yoshwata: You are right. I add `await`. Thank you.

PR #1532, opened 2019-03-07
fix(1010): [1][medium] add chainPR annotation.

wdstar:
## Context

prChain flag is moved from top-level setting to pipeline-level annotation.
- Current
```
prChain: true
jobs:
   main:
       ....
```
- After
```
annotations:
  screwdriver.cd/chainPR: true 
jobs:
   main:
       ....
```

## Objective

This PR adds value resolution of the `screwdriver.cd/chainPR` annotation and the plugin-level setting, which is added for maintenance of similarity with `restrictPR`.

## References

- https://github.com/screwdriver-cd/screwdriver/issues/1010#issuecomment-466188824

### DIFF
--- a/plugins/webhooks/index.js
+++ b/plugins/webhooks/index.js
@@ -5,6 +5,9 @@ const joi = require('joi');
 const winston = require('winston');
 const workflowParser = require('screwdriver-workflow-parser');
 
+const ANNOT_NS = 'screwdriver.cd';
+const ANNOT_CHAIN_PR = `${ANNOT_NS}/chainPR`;
+const ANNOT_RESTRICT_PR = `${ANNOT_NS}/restrictPR`;
 const WAIT_FOR_CHANGEDFILES = 1.8;
 
 /**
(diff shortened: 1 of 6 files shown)


plugins/webhooks/index.js
```diff
@@ -207,7 +211,8 @@ async function createPREvents(options, request) {
             configPipelineSha,
             startFrom: `~pr:${branch}`,
             changedFiles,
-            causeMessage: `${action} by ${userDisplayName}`
+            causeMessage: `${action} by ${userDisplayName}`,
+            chainPR: resolvedChainPR
```
wdstar: This `eventConfig.chainPR` flag is finally passed to `eventFactory.create()`.

wdstar:
The test about addition of `~release` and `~tag` nodes to workflowGraph fails.
https://cd.screwdriver.cd/pipelines/1/builds/98121

plugins/webhooks/index.js
```diff
@@ -617,7 +645,8 @@ exports.register = (server, options, next) => {
     const pluginOptions = joi.attempt(options, joi.object().keys({
         username: joi.string().required(),
         ignoreCommitsBy: joi.array().items(joi.string()).optional(),
-        restrictPR: joi.string().valid('all', 'none', 'branch', 'fork').optional()
+        restrictPR: joi.string().valid('all', 'none', 'branch', 'fork').optional(),
+        chainPR: joi.boolean().optional()
```
sakka2: should add param `chainPR` to doc comment.

plugins/webhooks/index.js
```diff
@@ -232,6 +237,21 @@ async function createPREvents(options, request) {
...
+ * @param  {Boolean}  chainPR              Plugin Chain PR flag
+ * @param  {Pipeline} pipeline             Pipeline
+ * @param  {Object}   pipeline.annotations Pipeline-level annotations
+ * @return {Boolean}
+ */
+function resolveChainPR(chainPR, pipeline) {
+    const defaultChainPR = typeof chainPR === 'undefined' ? false : chainPR;
+    const annotChainPR = pipeline.annotations[ANNOT_CHAIN_PR];
+
+    return typeof annotChainPR === 'undefined' ? defaultChainPR : annotChainPR;
```
sakka2: [ask] Is there any case that value of chainPR becomes `undefined`?

plugins/webhooks/index.js
wdstar: I added it. Thanks.

plugins/webhooks/index.js
wdstar: Yes, when there is no `screwdriver.cd/chainPR` annotation in a `screwdriver.yaml`.

plugins/webhooks/index.js
```diff
@@ -232,6 +237,21 @@ async function createPREvents(options, request) {
...
+/**
+ * Resolve ChainPR flag
+ * @method resolveChainPR
+ * @param  {Boolean}  chainPR              Plugin Chain PR flag
+ * @param  {Pipeline} pipeline             Pipeline
+ * @param  {Object}   pipeline.annotations Pipeline-level annotations
+ * @return {Boolean}
+ */
+function resolveChainPR(chainPR, pipeline) {
+    const defaultChainPR = typeof chainPR === 'undefined' ? false : chainPR;
```
tkyi: does this work too?
```suggestion
    const defaultChainPR = chainPR || false;
```

plugins/webhooks/index.js
tkyi: Does this work here?
```suggestion
    return annotChainPR || defaultChainPR;
```

wdstar: No. If `annotChainPR` is false and `defaultChainPR` is true, it returns true, but we expect false.

plugins/webhooks/index.js
wdstar: Yes, but we should not use `||` short circuit evaluation about **Boolean** type.
Please refer to the following comments too.

plugins/webhooks/index.js
tkyi: Ah thanks for explaining, makes sense

PR #1539, opened 2019-03-13
fix(1010): [small] add chainPR value to getNextJobs function

sakka2:
## Context
Now, Jobs does not chain even if chainPR is true.
We need to pass `pipeline.prChain` flag to getNextJobs function, when checking `triggerNextJobs`.

`getNextJobs` function has fixed before.
https://github.com/screwdriver-cd/workflow-parser/blob/master/lib/getNextJobs.js

## References
screwdriver-cd/screwdriver#1010 (comment)


jithin1987:
@sakka2 tests are failing after this PR was merged. I will have to add a revert PR to let the pipeline function for now.

Issue #1521, opened 2019-02-27
Pipeline status become unknown after [skip ci]


tk3fftk:
**What happened**:
After #1447 fixed, `[skip ci]` creates an empty event. It might make a pipeline status to unknown. It also brakes a badge status.

**What you expected to happen**:
- Set another status for `[skip ci]` event.
- Or get a status before `[skip ci]` event.

**How to reproduce it**:
-  commit with `[skip ci]`
- show a badge of the pipeline

Related: https://github.com/screwdriver-cd/screwdriver/issues/1447

d2lam:
@tk3fftk This also applies to `sourcePaths`. 

I think the correct behavior would be get the status before `[skip ci]` event.

wahapo:
It can be realized by executing `lastEvent = events.pop()` until `lastEvent.getBuilds()` is not empty.
https://github.com/screwdriver-cd/screwdriver/blob/6cdfabb996d895f8eeb64602f09ab9e3f4d7a0f5/plugins/pipelines/badge.js#L108

wahapo:
The badge that was the number one issue has been corrected.

The status of `Event` itself is not owned anywhere. For now, as far as I investigated, non-empty event ID is not required anywhere except for api of badge. I think that there is no need to fix to possess the ID of empty event if it is useless. We should think when we need it. I think it is good to close the issue ticket, how is it?

The response of `GET /v4/pipelines/<id>` is as follows.
```
{
  id: 1,
  name: 'wahapo/wahapo-repository',
  scmUri: 'github.url:1:master',
  scmContext: 'github:github.url',
  scmRepo:
  {
    branch: 'master',
    name: 'wahapo/wahapo-repository',
    url: 'https://github.url/wahapo/wahapo-repository/tree/master'
  },
  createTime: '2019-03-18T04:00:00.000Z',
  admins: { wahapo: true },
  workflowGraph: { nodes: [Array], edges: [Array] },
  annotations: {},
  lastEventId: 4,
  configPipelineId: undefined,
  childPipelines: undefined,
  prChain: false
}
```

And the response of `GET /v4/events/<id>` is as follows.
```
{
  id: 4,
  causeMessage": 'Merged by wahapo',
  commit: {
    author: {
      avatar: 'avatar.url',
      name: 'wahapo',
      username: 'wahapo',
      url: 'https://github.url/wahapo'
    },
    message: '[skip ci]test',
[...]

PR #1543, opened 2019-03-14
fix(1521): badge use last effective event status

wahapo:
## Context
https://github.com/screwdriver-cd/screwdriver/issues/1521#issuecomment-472771605

## Objective
When `[skip ci]` created an empty event, the badge was displayed as unknown. Fixed to create badges as far back as effective build.
ui

■Before
before

■After
after

### DIFF
--- a/plugins/pipelines/badge.js
+++ b/plugins/pipelines/badge.js
@@ -104,8 +104,8 @@ module.exports = () => ({
                         return reply.redirect(getUrl(badgeService));
                     }
 
-                    return pipeline.getEvents({ sort: 'ascending' }).then((events) => {
-                        const getLastEffectiveEvent = async (events) => { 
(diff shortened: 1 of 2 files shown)


plugins/pipelines/badge.js
```diff
@@ -104,38 +104,42 @@ module.exports = () => ({
                         return reply.redirect(getUrl(badgeService));
                     }
 
-                    return pipeline.getEvents({ sort: 'ascending' }).then((events) => {
-                        const lastEvent = events.pop();
+                    return pipeline.getEvents({ sort: 'ascending' }).then((allEvents) => {
+                        const getLastEffectiveEvent = async (events) => {
```
catto: There is no `await` in this function, so this `async` shouldn't be needed.

wahapo: fixed

PR #1548, opened 2019-03-14
feat(1412): Add default time range & time range check

d2lam:
We don't want to allow getting metrics for all time since that might be a lot of data. 
Limiting it to 6 months for now. If user requests for more than 6 months, it will give `400` with appropriate message. If the time is not passed in, API will default to `6 months to now`.  


plugins/builds/metrics.js
```diff
@@ -2,6 +2,8 @@
 
 const boom = require('boom');
 const joi = require('joi');
+const { setDefaultTimeRange, validTimeRange } = require('../helper.js');
+const MAX_DAYS = 180; // 6 months
```
jithin1987: can this is made a k8s env variable so that we could relax on fly ?

plugins/helper.js
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Set default start time and end time
+ * @method setDefaultStartEnd
+ * @param  {String}         start     start time
+ * @param  {String}         end       end time
+ * @param  {Number}         maxDay    max day range
+ * @return {Object}                   Default start time and end time
+ */
+function setDefaultTimeRange(start, end, maxDay) {
+    const endTime = end || new Date(Date.now()).toISOString();
```
jithin1987: we don't need Date.now() here `(new Date()).toISOString()` should do

plugins/helper.js
```diff
@@ -0,0 +1,37 @@
...
+    const startTime = start || new Date(new Date().setDate(new Date(endTime).getDate() - maxDay))
```
jithin1987: I would recommend using `moment.js` for this. This can be concisely expressed using https://momentjs.com/docs/#/manipulating/subtract/

d2lam: I saw `moment.js` but not sure if we want to pull a new dependency since the calculation is quite simple.
I can do it if that's what we prefer.

jithin1987: we are creating 6 date objects in this one function call. `moment` will abstract all this logic. and is well tested.

DekusDenial:
please dont use `moment` unless it's necessary.
https://github.com/you-dont-need/You-Dont-Need-Momentjs

plugins/helper.js
d2lam: @DekusDenial @jithin1987   you two can fight :p

jithin1987: I agree on the performance aspect. If we are to replace with some alternative libraries from the link @DekusDenial posted, I am okay with it.

However some of the concerns are not applicable. This does not go into browser bundle, but on API server.

d2lam: I prefer to not pull in another dependency since the calculation isn't overly complicated. 
But I can switch to `dayjs` or `date-fns`, it looks more lightweight.

plugins/builds/metrics.js
d2lam: what would it be under?

jithin1987: API deployment config ? Not a big deal, but if we choose later to relax this env will be better

package.json
```diff
@@ -66,6 +66,8 @@
     "boom": "^7.3.0",
     "config": "^1.31.0",
     "crumb": "^6.1.0",
+    "date-fns": "^1.30.1",
```
xxyuk: Seems `date-fns` is never used.

DekusDenial: I would have rather use it instead, lol

plugins/helper.js
DekusDenial: 1) no time zone exchange here
2) it’s ISO, u want one with faster parsing. This again brings up my question about why almost all our times are in ISO and not epoch which is more of the norm and no brainer in parsing

DekusDenial: Using `moment` blindly is like importing the whole `lodash/underscore` but u only use one function from it.

PR #1554, opened 2019-03-18
fix(1010): add chainPR flag to eventConfig

sakka2:
## Context
chainPR flag is not passed when event is created.
This API will be called when `Restart` button of the Job is pushed for example.

## Objective
Add chainPR flag of base pipeline to payloads when event is created.

## References
Issue: https://github.com/screwdriver-cd/screwdriver/issues/1010


plugins/events/create.js
```diff
@@ -182,6 +182,7 @@ module.exports = () => ({
 
                                         payload.prInfo = prInfo;
                                         payload.prRef = prInfo.ref;
+                                        payload.chainPR = pipeline.prChain;
```
d2lam: probably too late to change this since it's already in the `models`, but not sure why we name they key `chainPR` instead of `prChain`

tkyi: I think it was to be consistent with `restrictPR`

PR #1569, opened 2019-03-27
fix(1412): allow aggregation

d2lam:
Allows aggregation option and combine steps in build metrics

Blocked by: https://github.com/screwdriver-cd/models/pull/355


d2lam:
Coverage decreased because `step metrics` is removed, so it brings down overall coverage. But the two files being changed are still 100%

plugins/jobs/metrics.js
```diff
@@ -7,7 +7,7 @@ const MAX_DAYS = 180; // 6 months
 
 module.exports = () => ({
     method: 'GET',
-    path: '/jobs/{id}/metrics/builds',
```
jithin1987: does UI need to change ?

d2lam: UI doesn't call this yet.

PR #1573, opened 2019-03-29
fix: display disabled state for job badge

minz1027:
Currently, the job badge shows build status of the last build. This PR will show the state of the job if it is `DISABLED`.

### DIFF
--- a/plugins/pipelines/jobBadge.js
+++ b/plugins/pipelines/jobBadge.js
@@ -62,6 +62,17 @@ module.exports = config => ({
                     return reply.redirect(getUrl(badgeConfig));
                 }
 
+                if (job.state === 'DISABLED') {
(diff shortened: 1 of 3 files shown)


plugins/pipelines/jobBadge.js
```diff
@@ -62,6 +62,17 @@ module.exports = config => ({
                     return reply.redirect(getUrl(badgeConfig));
                 }
 
+                if (job.state === 'DISABLED') {
+                    return reply.redirect(getUrl(Object.assign(
```
jithin1987: can we still fetch the last build status and display it

`pipeline/job success` in grey color

minz1027: i thinks this is confusing by mixing the color and words

PR #1580, opened 2019-04-08
fix(1010): [3] Fix the error from forked repository

sakka2:
## Context
Now, job error happens in PRs from forked repository, because `sd-scm-github` step tries to `git reset` with the commit SHA of forked repository which dose not exist in original one.
This is because there is no `config.prRef` value in argument below.
https://github.com/screwdriver-cd/scm-github/blob/master/index.js#L316

## Objective
In chainPR workflow, we should pass the `prRef` value to the buildFactory.create() method.
To achieve this, 
1. We need to set `prRef` into the events model  when event is created.
2. Then, we can get `prRef` from events model like `event.pr.ref`;

## References
Issue:
https://github.com/screwdriver-cd/screwdriver/issues/1010#issuecomment-478911280

Other PRs:
[1]: data-schema: https://github.com/screwdriver-cd/data-schema/pull/327
[2]: models: https://github.com/screwdriver-cd/models/pull/359

### DIFF
--- a/plugins/builds/index.js
+++ b/plugins/builds/index.js
@@ -27,30 +27,46 @@ const workflowParser = require('screwdriver-workflow-parser');
  * @param  {Boolean}  [config.start]        Whether to start the build or not
  * @return {Promise}
  */
-function createBuild({ jobFactory, buildFactory, eventFactory, pipelineId, jobName, username,
-    scmContext, build, start }) {
-    return Promise.all([
-        eventFactory.get(build.eventId),
-        jobFactory.get({
-            name: jobName,
-            pipelineId
-        })
-    ]).then(([event, job]) => {
-        if (job.state === 'ENABLED') {
-            return buildFactory.create({
-                jobId: job.id,
-                sha: build.sha,
-                parentBuildId: build.id,
-                eventId: build.eventId,
-                username,
-                configPipelineSha: event.configPipelineSha,
-                scmContext,
-                start: start !== false
-            });
-        }
-
-        return null;
+async function createBuild({ jobFactory, buildFactory, eventFactory, pipelineFactory,
+    pipelineId, jobName, username, scmContext, build, start }) {
+    const event = await eventFactory.get(build.eventId);
+    const job = await jobFactory.get({
+        name: jobName,
+        pipelineId
     });
+
+    let prRef = '';
+
+    if (event.type === 'pr') {
+        const scm = buildFactory.scm;
+        const pipeline = await pipelineFactory.get(pipelineId);
+        const token = await pipeline.token;
+        const scmConfig = {
+            scmContext: pipeline.scmContext,
+            scmUri: pipeline.scmUri,
+            token,
+            prNum: event.prNum
+        };
+        const prInfo = await scm.getPrInfo(scmConfig);
+
+        prRef = prInfo.ref;
+    }
+
+    if (job.state === 'ENABLED') {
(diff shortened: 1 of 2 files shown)


plugins/builds/index.js
```diff
@@ -27,30 +27,46 @@ const workflowParser = require('screwdriver-workflow-parser');
...
+        pipelineId
     });
+
+    let prRef = '';
+
+    if (event.type === 'pr') {
+        const scm = buildFactory.scm;
+        const pipeline = await pipelineFactory.get(pipelineId);
+        const token = await pipeline.token;
+        const scmConfig = {
```
jithin1987: do we have to query `scm` each time ? Can `prRef` be stored in PR event ?

sakka2: Oh, I think it's better, thanks!

plugins/builds/index.js
```diff
@@ -27,30 +27,46 @@ const workflowParser = require('screwdriver-workflow-parser');
...
-                configPipelineSha: event.configPipelineSha,
-                scmContext,
-                start: start !== false
-            });
-        }
-
-        return null;
+async function createBuild({ jobFactory, buildFactory, eventFactory, pipelineFactory,
+    pipelineId, jobName, username, scmContext, build, start }) {
+    const event = await eventFactory.get(build.eventId);
```
tk3fftk: Do we need some error handling for these `await` statements?

plugins/builds/index.js
```diff
@@ -27,30 +27,46 @@ const workflowParser = require('screwdriver-workflow-parser');
  * @param  {Boolean}  [config.start]        Whether to start the build or not
  * @return {Promise}
  */
```
tk3fftk: Need to add `pipelineFactory` to JSDoc

sakka2: We didn't need pipelineFactory, so I deleted.

plugins/builds/index.js
sakka2: @jithin1987 
I fixed to store the `prRef` into the event model.

Please look these.
https://github.com/screwdriver-cd/data-schema/pull/327
https://github.com/screwdriver-cd/models/pull/359

plugins/builds/index.js
sakka2: May be error will be catched here?
`createBuild` method is called from `triggerNextJobs`, and `triggerNextJobs` is called here.
https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/builds/update.js#L208

plugins/builds/index.js
```diff
@@ -27,30 +27,30 @@ const workflowParser = require('screwdriver-workflow-parser');
...
-
-        return null;
+async function createBuild({ jobFactory, buildFactory, eventFactory, pipelineId,
+    jobName, username, scmContext, build, start }) {
+    const event = await eventFactory.get(build.eventId);
+    const job = await jobFactory.get({
+        name: jobName,
+        pipelineId
     });
+    const prRef = event.pr.ref ? event.pr.ref : '';
```
jithin1987: is `pr` guaranteed to be an object here ? else we should do a check on `event.pr` as well.

sakka2: @jithin1987 
yes, `pr` is guaranteed to ben an object here. Please see this link.
https://github.com/screwdriver-cd/models/blob/master/lib/eventFactory.js#L416

PR #1582, opened 2019-04-10
fix(1412): allow other aggregation options

d2lam:
Allows other aggregation options

Blocked by: https://github.com/screwdriver-cd/models/pull/361


plugins/pipelines/metrics.js
```diff
@@ -54,7 +56,7 @@ module.exports = () => ({
             query: joi.object({
                 startTime: joi.string().isoDate(),
                 endTime: joi.string().isoDate(),
-                aggregate: joi.boolean().default(false)
+                aggregate: joi.string().valid('none', 'day', 'week', 'month', 'year')
```
DekusDenial: honestly i hate to leave this kinda comment but i would suggest u change the name to something like `aggregateInterval` I know it's long but it doesn't have to be exact wording

d2lam: Changed to aggregateInterval

Issue #1285, opened 2018-09-11
Stop parallel builds require a hard task


yuichi10:
**What happened**:
Can not stop parallel builds by step.
When I try to stop parallel builds, I need to stop one by one. It very tough things for me.

**What you expected to happen**:
I want to stop all parallel builds by one step.

**How to reproduce it**:
Running parallel builds and try to stop every these builds.

yoshwata:
Some new UIs are needed to deal with this issue.

### For normal events
My proposal is like below. A new `Stop` button is on UI, next of `DURATION`.
<img width="1287" alt="collage2" src="https://user-images.githubusercontent.com/32473622/53146950-60407580-35e9-11e9-8ade-d62a9c972bc7.png">

After the `Stop` is pushed, all `QUEUED` or `RUNNING` jobs are stopped. The stop button is not shown at this situation because there are no jobs to stop.
![screen shot 2019-02-21 at 14 01 02](https://user-images.githubusercontent.com/32473622/53147228-3f2c5480-35ea-11e9-9602-c32d63d0e665.jpg)

### For pr events
For PR events, I propose UI like below. PR jobs are surrounded by frame. It gathers PR events by each PR. A stop button is put on frames of non stopped events.
![collage5](https://user-images.githubusercontent.com/32473622/53147239-4b181680-35ea-11e9-8638-eb1939c1bb50.jpg)

Is it reasonable? Are there better idea?

wahapo:
I have another idea. It is a proposal that put the link of `Stop all builds in this event` in the tooltip. In addition, we are also good to fill the tooltip in the swirl on the right side.
<img width="1215" src="https://user-images.githubusercontent.com/24538326/53157887-caffaa00-3605-11e9-9334-51b8c93aad51.png">

yoshwata:
Thanks @wahapo! I tried to write image of PR pattern in your idea. Is it make your sence?
<img width="1306" alt="collage7" src="https://user-images.githubusercontent.com/32473622/53219820-e968b280-36a4-11e9-944f-b6163e308d7a.png">

jithin1987:
The Pull Request section will have to have a re-design when we PR chain feature is functional. It will have to behave similar to how events, does. Infact nothing prevents us from implementing the same today, ie Pull Requests section should show pull request events instead of jobs. So there won't be multiple cards for each parallel jobs but one card for one event which shows all parallel jobs. 

[...]

PR #1592, opened 2019-04-19
feat(1285): Add endpoint for stopping all builds in an event

tkyi:
## Context

It can be annoying for users to have to manually stop each build in an event. It would be nice if they could stop all running/queued/blocked builds with one click or API call.

## Objective

This PR adds an endpoint that checks/updates user permissions and gets all running/queued/blocked builds for an event and updates the status to `ABORTED`.

_Note: The endpoint is currently a `PUT /events/{id}/stop`, but open to any ideas.
Wondering if `/events/{id}/builds/stop` or `/events/{id}/builds/stopall` would be better. Or a `POST`?_

This PR also does some light refactoring, moving the `updateAdmins` function into `events/index.js`.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/1285


plugins/events/stopBuilds.js
```diff
@@ -0,0 +1,135 @@
...
+                                    // Add good user to admins
+                                    return updateAdmins({
+                                        permissions,
+                                        pipeline,
+                                        username
+                                    });
+                                }
+
+                                // PR author should be able to stop their own PR build
+                                if (event.prNum && isPrOwner) {
```
d2lam: we can combine `|| adminDetails.isAdmin` here and resolve

plugins/events/stopBuilds.js
```diff
@@ -0,0 +1,135 @@
...
+                        // Check permissions
+                        return user.getPermissions(pipeline.scmUri)
+                            .then((userPermissions) => {
+                                const isPrOwner = hoek.reach(event,
+                                    'commit.author.username') === username;
+
+                                permissions = userPermissions;
+
+                                // Check if user has push access or is a Screwdriver admin
+                                if (permissions.push || adminDetails.isAdmin) {
```
d2lam: I think we can remove this block `updateAdmins` will do the adding/removing good/bad admins and resolves/throw

Issue #1070, opened 2018-05-17
Environment variables are not evaluated


jithin1987:
**What happened**:

While SD allows configuration of environment variables via [`environment` section](https://docs.screwdriver.cd/user-guide/configuration/environment), they are not evaluated, instead it is taken as it is.

Lets say `$SD_ARTIFACTS_DIR=/tmp`
```
shared:
  environment:
    TEST_PATH: $SD_ARTIFACTS_DIR/test
```

Inside a build 
```
echo $TEST_PATH
$SD_ARTIFACTS_DIR/test
```

**What you expected to happen**:
```
echo $TEST_PATH
/tmp/test
```

**How to reproduce it**:

1. Set an environment variable with refers another one. Eg: `TEST_PATH: $SD_ARTIFACTS_DIR/test`
2. Using that environment variable in a build does not give the expected value.

Zhongtang:
One thing we need to pay attention to is the order of evaluation. e.g. assuming `$SD_SOURCE_DIR` is `/tmp`

```
shared:
  environment:
    PATH1: $PATH2/test
    PATH2: $SD_SOURCE_DIR/sunfire_reports
```

In this case, do we evaluate `$PATH1` as `/test` (evaluate based on the order of definition) or `/tmp/sunfire_reports/test` ? If we do the later, we need to be careful about potential loop in evaluation, e.g.

```
shared:
  environment:
    PATH1: $PATH2/test
    PATH2: $PATH1/sunfire_reports
```

minz1027:
this is because we set the initial env via exec.Command.Env
https://github.com/screwdriver-cd/launcher/blob/master/executor/executor.go#L217

exec pkg intentionally does not expand env: https://golang.org/pkg/os/exec/

To make it evaluate in order, we can put export env cmd in a file and source it

d2lam:
### Update 4/19/2019
Opened PRs to write the base ENV into a file, and inject the source file command into the `sd-setup-launcher` step.

d2lam:
### Update 4/25/2019
There is a problem:
`environment` is an object, which does not guarantee order. So if user puts:
```
A="hello"
B="${A}world"
```
It's possible B is evaluated first. 

The best and least-hacky way to solve this is making `environment` an array. We initially made it an `object` since the order didn't matter then (no eval env). 
Will keep this issue opened so that we can continue discussing.

d2lam:
### Update 5/6/2019
[...]

PR #1600, opened 2019-04-26
fix(1070): convert env to array

d2lam:
Launcher calls this endpoint to get the environment variables. Since it's an object, golang randomizes the order, which might gives issues once we start supporting evaluate ENV. 

This PR converts the `environment`, and launcher will interpret `environment` as an array. 

**Note**: this change is backward compatible for the users. However, when we deploy, we need to deploy both API and launcher together. There might be a short time (during deployment) that builds might break. 

**Blocked by:**
- https://github.com/screwdriver-cd/data-schema/pull/334
- launcher changes

**Related**: https://github.com/screwdriver-cd/screwdriver/issues/1070


test/plugins/data/build.json
```diff
@@ -9,6 +9,9 @@
     "endTime": "2038-01-19T03:16:10.131Z",
     "parameters": {},
     "stats": {},
+    "environment": {
```
jithin1987: can we have a test case with multiple values ?

d2lam: actually added this here to only show we have `environment` in build object, but the test cannot cover the converting because the `update` function is from model and it's mocked.

Issue #1455, opened 2019-01-25
screwdriver yaml for mono repo can grow big and hard to maintain


jithin1987:
**What happened**:

`sourcePaths` feature which  is used to construct monorepo based workflow can result in a large `screwdriver.yaml` file depending on the number of sub projects in the repository. In order to better manage workflow of individual sub projects in a monorepo, separating out `screwdriver.yaml` configurations for the sub projects will result in smaller and maintainable workflow configuration files

**What you expected to happen**:

**How to reproduce it**:

tkyi:
Would this feature basically be a way to specify where source code is?
Currently, `screwdriver.yaml` and all other files are relative to root. We could allow users to specify `root` in their screwdriver.yaml file. All files and things would be taken relative to that root.

```
jobs:
    main:
        image: node:6
        requires: [~pr, ~commit]
        root: "src/app/"
        steps:
            - echo: echo hi
```

jithin1987:
I am thinking of supporting multiple `screwdriver.yaml` configurations at sub directories which leads to separate different pipelines. There are two use cases here.

1. With source paths.
    With source paths, multiple `screwdriver.yaml` files will still have a single pipeline, but config for different `sourcePaths` will come from separate `screwdriver.yaml` files. This approach will have  only one pipeline.
1. Without source paths.
    This will be the simplest to implement. Uniqueness of the pipeline now won't be just repo + branch. But repo + branch + path to `screwdriver.yaml` file. This approach will lead to multiple pipelines.

minz1027:
I think combined it with sourcePaths is making this very complicated. Cuz when user has a yaml under a subdir, they probably expect cmd ran relative to that subdir. If that subdir doesn't match the sourcePath, we do we do? When we create the workspace, where should we point to? Also to parse the yaml, if you want to have a sub yaml with multiple jobs, how would that look like? Since from the root yaml, you need to defined A job with with sourcePaths first.

e.g.
```jobs:
    main:
        image: node:6
        requires: [~pr, ~commit]
        sourcePaths: ["src/app/", "test/abc"]
        configPath: someDir
```

minz1027:
I think the second way is neat and simpler. But that root / sourceDir config cannot come from the yaml. We need to have that when user create the pipeline since that will be part of the unique constraints.

jithin1987:
[...]

PR #1601, opened 2019-04-29
feat(1455): Allow create or update pipeline with rootDir (5)

tkyi:
## Context
Would be nice if users could put screwdriver.yaml/source dir not at root.

## Objective
This PR adds rootDir to pipeline create and update. Also adds helper function to make sure rootDir does not start with or end with a `/`.

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/1455


plugins/pipelines/helper.js
```diff
@@ -26,6 +26,27 @@ const formatCheckoutUrl = (checkoutUrl) => {
...
+/**
+ * Get rid of leading/trailing slashes in rootDir, return empty string as default
+ * @method sanitizeRootDir
+ * @param  {String}     rootDir     Root directory (ex: /src/component/app/ or /)
+ * @return {String}                 Root dir with no leading/trailing slashes
+ */
+const sanitizeRootDir = (rootDir = '') => {
+    let result = rootDir;
+
+    if (rootDir.startsWith('/')) {
```
d2lam: shouldn't we just get rid of all extra `/`. 
For example: `//src` --> `src`

plugins/pipelines/helper.js
```diff
@@ -26,6 +26,27 @@ const formatCheckoutUrl = (checkoutUrl) => {
...
+        result = result.slice(1);
```
d2lam: shouldn't we just get rid of all extra /.
For example: //src --> src

plugins/pipelines/helper.js
```diff
@@ -26,6 +26,27 @@ const formatCheckoutUrl = (checkoutUrl) => {
...
+    }
+
+    if (rootDir.endsWith('/')) {
+        result = result.slice(0, -1);
```
d2lam: same comment as above

plugins/pipelines/helper.js
tkyi: I think that's fair
updated method

plugins/pipelines/helper.js
```diff
@@ -26,6 +26,15 @@ const formatCheckoutUrl = (checkoutUrl) => {
     return result;
 };
 
+/**
+ * Get rid of leading/trailing slashes in rootDir, return empty string as default
+ * @method sanitizeRootDir
+ * @param  {String}     rootDir     Root directory (ex: /src/component/app/ or /)
+ * @return {String}                 Root dir with no leading/trailing slashes
+ */
+const sanitizeRootDir = (rootDir = '') => rootDir.replace(/^\/+|\/+$/g, '');
```
jithin1987: can we also validate if it's an actual path ?

minz1027: you mean check if it exists in the repo? that's going to be hard, need to make git calls

jithin1987: Not checking if it actually exits in repo, just patterns for valid file path. Eg: /x/y/z or ./s or s/y

tkyi: I think that would be hard to do correctly. Here are some of the top results based on a quick search online:
- https://stackoverflow.com/questions/16231210/file-path-validation-in-javascript
- https://www.codeproject.com/Tips/216238/Regular-Expression-to-Validate-File-Path-and-Exten
- https://www.npmjs.com/package/is-valid-path (this package was last published 4 years ago)

If you really want it, I can try to adapt one of those regex patterns.

jithin1987: we need to look into only valid unix file paths.  

https://stackoverflow.com/questions/28989672/regex-to-tell-if-a-string-contains-a-linux-file-path-or-if-a-linux-file-path-as
https://stackoverflow.com/questions/55069650/regex-pattern-to-validate-linux-folder-path

tkyi: added

tkyi:
@jithin1987 Updated regex to strip `./` from path

Issue #823, opened 2017-12-14
Trigger Pipeline on GitHub Release


tk3fftk:
User wants to start jobs triggered [release event](https://developer.github.com/v3/activity/events/types/#releaseevent) same like `~pr` or `~commit`.
This feature image is like following:
![image](https://user-images.githubusercontent.com/3445553/33981247-64af3304-e0ef-11e7-8226-36eaac9f37f9.png)
If a user wants to trigger another pipeline's jobs, the user should set `~sd@1234:release` in this case.

I could not think about the cases of Bitbucket and Gitlab...

(updated) concourse has this feature: https://github.com/concourse/github-release-resource

minz1027:
a work around can be `~sd@1234:job_that_does_release` and we support that already. But you guys actually want it to check against github release instead of a screwdriver job?

tk3fftk:
Maybe yes. This feature request was mentioned our coworker.
Anyway, I'll tell him about the work around. Thanks!

tk3fftk:
I asked the requester. Actually, he wants to trigger same pipeline's jobs (e.g., production deployment job) by creating "Release" manually on GitHub UI.

petey:
This feels like an overly specific tie in to Github features. I do not think that Bitbucket or Gitlab have a "release" entity or event.

lgfausak:
I took a quick look at the webhooks on github.  IMO there are a number of things that would make for nice triggers/entry points, not just _release_, but also _issues_, _pull request_, _label_, _delete_, etc. If the specificity is an issue perhaps the entrypoint could be given some sort of namespace; e.g. **github**-_release_, **github**-_delete_, etc to make it clear that the feature is not going to work on bitbucket.

yuichi10:
@minz1027 @jithin1987 
Many users want to use this feature in my working place. Is this feature in progress?
Thank you.

tk3fftk:
`~tag:branch` feature has a bug. It watched `master_branch` as `branch` but it doesn't work when `master_branch` is different from a tagged branch.
https://github.com/screwdriver-cd/scm-github/blob/5530085d6a7d4e487e66d8c59d6b5c6c2a617431/index.js#L1073

In addition, `~tag:/<tag_name>` feature will be supported for more use case instead of `~tag:branch`.
So we should think about `~tag:<something>` feature in the second pass.

jithin1987:
@catto awaiting an example pipeline https://github.com/screwdriver-cd-test

[...]

PR #1621, opened 2019-05-10
fix(823): [2] fixed not to create event when there is no job triggered by '~tag' or '~release'

sakka2:
## Context
Now, tag or release events are created even if user does not have '~tag' or '~release' definition in screwdriver.yaml.

## Objective
These are empty events so we should not create them.

## References
https://github.com/screwdriver-cd/screwdriver/issues/823#issuecomment-489999734

## OtherPRs
[1] https://github.com/screwdriver-cd/data-schema/pull/338


plugins/webhooks/index.js
```diff
@@ -600,6 +600,15 @@ async function createEvents(eventFactory, userFactory, pipelineFactory,
...
+        const checkStartFroms = ['~tag', '~release'];
+
+        if (checkStartFroms.includes(startFrom) && !hasTriggeredJob(p, startFrom)) {
+            winston.info(
+                `event is not created because there are no jobs triggered by ${startFrom}`
+            );
+            continue; // eslint-disable-line no-continue
+        }
+
         events.push(eventFactory.create(eventConfig));
```
wdstar: This `evens.push()` should be included in the last `if ... else ...` block.

plugins/webhooks/index.js
```diff
@@ -600,6 +600,15 @@ async function createEvents(eventFactory, userFactory, pipelineFactory,
         await updateAdmins(userFactory, username, scmContext, p);
         /* eslint-enable no-await-in-loop */
 
+        const checkStartFroms = ['~tag', '~release'];
```
tk3fftk: I think this information should not be in this line. This should be constant as a global variable or should be in data-schema.

tk3fftk: Also this param name should have more information. How about `excludedStartFroms` or `excludableStartFroms`?

plugins/webhooks/index.js
sakka2: Thanks! I fixed to use `if .. else ..` block.

plugins/webhooks/index.js
sakka2: Thanks! I fixed to use data-schema.
https://github.com/screwdriver-cd/data-schema/pull/338

plugins/webhooks/index.js
```diff
@@ -4,11 +4,14 @@ const boom = require('boom');
...
 const winston = require('winston');
 const workflowParser = require('screwdriver-workflow-parser');
+const schema = require('screwdriver-data-schema');
 
 const ANNOT_NS = 'screwdriver.cd';
 const ANNOT_CHAIN_PR = `${ANNOT_NS}/chainPR`;
 const ANNOT_RESTRICT_PR = `${ANNOT_NS}/restrictPR`;
-const CHECKOUT_URL_SCHEMA = require('screwdriver-data-schema').config.regex.CHECKOUT_URL;
+const EXTRA_TRIGGERS = schema.config.regex.EXTRA_TRIGGERS;
+const EXTRA_TRIGGERS_REGEXP = new RegExp(EXTRA_TRIGGERS);
```
wdstar: `schema.config.regex.EXTRA_TRIGGERS` is already a regex object.
This line's  instantiation is not necessary?

sakka2: Thanks! I fixed.

plugins/webhooks/index.js
tk3fftk: `EXTRA_TRIGGERS` looks good 👍

sakka2:
Sorry, I fixed just the typo. Could you merge this please?

PR #1643, opened 2019-05-21
fix(1455): add webhook trigger for pipelines with rootDir

minz1027:
- add webhook trigger for pipelines with rootDir
- do not create event at all if no changeFiles go under rootDir
- refactor the triggeredPipelines logic for branch filtering, instead of getting a list of branches from git and construct the `scmUris` accordingly, we will just search in the pipeline tables for anything starts with the same scm host and repoId,  e.g.`github.com:12345`

Related to: https://github.com/screwdriver-cd/screwdriver/issues/1455

### DIFF
--- a/plugins/webhooks/index.js
+++ b/plugins/webhooks/index.js
@@ -163,6 +163,25 @@ function hasTriggeredJob(pipeline, startFrom) {
 }
 
 /**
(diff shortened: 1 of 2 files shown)


minz1027:
@jithin1987 branch filtering is part of our function tests. Why would this affect external triggers? how external triggers next job is not part of the webhook logic

PR #1653, opened 2019-05-23
fix: refactor webhook PR events handler

minz1027:
There are a couple issues with current webhook PR logic. This PR will fix these issues.
- duplicate calls to get `triggeredPipelines` to get relevant pipelines for a given webhook event
- PR restriction check, admin updates, abort running builds etc. only apply to a single pipeline not all the relevant ones
- unnecessary pipeline sync for all the PR events before event creation. when we create the event, we will do a pipeline sync over there.

### DIFF
--- a/plugins/webhooks/index.js
+++ b/plugins/webhooks/index.js
@@ -83,6 +83,18 @@ async function updateAdmins(userFactory, username, scmContext, pipeline) {
 }
 
 /**
+ * Update admins for an array of pipelines
+ * @param  {Object}     config.userFactory      UserFactory
+ * @param  {Array}      config.pipelines        An array of pipelines
+ * @param  {String}     config.username         Username
+ * @param  {String}     config.scmContext       ScmContext
+ */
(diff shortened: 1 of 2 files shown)


plugins/webhooks/index.js
```diff
@@ -294,18 +351,19 @@ async function createPREvents(options, request) {
 }
 
 /**
- * Resolve ChainPR flag
- * @method resolveChainPR
- * @param  {Boolean}  chainPR              Plugin Chain PR flag
- * @param  {Pipeline} pipeline             Pipeline
- * @param  {Object}   pipeline.annotations Pipeline-level annotations
- * @return {Boolean}
+ * Stop all the relavant PR jobs for an array of pipelines
```
tkyi:
```suggestion
 * Stop all the relevant PR jobs for an array of pipelines
```

minz1027: lol my spelling!

minz1027: addressed

PR #1671, opened 2019-06-06
fix: tag/release are triggered by every branch pipelines

yuichi10:
## Context
`~tag` trigger is not triggered, if you use a pipeline that targets to the non master branch. Only these pipeline have `~tag:master`, the trigger will fire. 
When user creates a tag, all branch pipeline should trigger `~tag` trigger.

The same is the case with `~release` trigger

## Objective
All branch pipeline job search jobs which trigger from `~tag` and `~release`, when tag or release trigger was fired.

## Reference
#823 
On the second pass, tag and release trigger will have tag name filter function.

### DIFF
--- a/plugins/webhooks/index.js
+++ b/plugins/webhooks/index.js
@@ -31,8 +31,7 @@ function determineStartFrom(action, type, targetBranch, pipelineBranch) {
     } else {
         switch (action) {
         case 'release':
-            startFrom = '~release';
-            break;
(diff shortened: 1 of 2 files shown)


tk3fftk:
Do we need same fix for a release trigger?

yuichi10:
@tk3fftk 
That's right. I also fix the release one.

Issue #1427, opened 2019-01-03
Trusted templates & commands


jithin1987:
## Context

Trust implementation for Templates & Commands is a continuation of work from https://github.com/screwdriver-cd/screwdriver/issues/1271

## Objective
1. Design backend for marking a template/command as trusted.
1. UI to highlight trusted templates/commands . Refer [mocks](https://github.com/screwdriver-cd/screwdriver/issues/1271#issuecomment-426421541)

## References

1. Feature discussion - https://github.com/screwdriver-cd/screwdriver/issues/1271#issuecomment-430828998 
2. Namespace auth - https://github.com/screwdriver-cd/screwdriver/issues/1121

frankthrock:
It might be interesting to include a count of the number of times each template was used in the n days to give a sense of stability...

d2lam:
So there are 2 ways to approach this:

### **Screwdriver admins decide what is trusted**
- Pros: simple implementation. We have a separate table with template namespace, name, and a flag to mark it's trusted or not. 
- Cons: maintenance will get out of hand. Cluster admins will be the bottleneck. When template owners request to mark their template as trusted, how do you decide? Do you need to go through a set of steps to verify it? Or is it merely based on the trusting relationship with the template owner? Also, this action is done manually. 

### **Screwdriver users decide what is trusted** 
We can use a combination of how github & npm does it. Users can **star** a template. Templates with more stars can be interpreted as more trusted. Additionally, we can have fields such as **last published** and **download counts**. These fields need to be auto updated when a template is created, and when a template is used. 
- Pros: There will be no maintenance required by the cluster admins since the users decide what is trusted or not. Also, not every template needs to be "official" to be useful. This approach can let the users decide what's useful rather than us deciding what's good for them.  
- Cons: Implementation will be trickier. 

Having something "officially approved by the Screwdriver team" (approach #1) sounds assuring, but most of the time the Screwdriver team *doesn't* have the knowledge of what all these templates do. We cannot decide what is good/bad. I think letting the users decide would be more appropriate.

tkyi:
I agree and I like the `Screwdriver users decide what is trusted` option better.

The first option seems to have a lot of overhead cost and is unsustainable. The second option has a higher initial cost but I think it will be better in the long run.

jithin1987:
[...]

PR #1681, opened 2019-06-20
fix: add admin only endpoint to mark "trusted"

DekusDenial:
## Context

To provide admins access to mark certain template & command as trusted

## Objective

To add separate endpoints for template & command for marking trusted

## References

https://github.com/screwdriver-cd/screwdriver/issues/1427

blocked by: 
https://github.com/screwdriver-cd/data-schema/pull/340

## License

### DIFF
--- a/plugins/commands/updateTrusted.js
+++ b/plugins/commands/updateTrusted.js
@@ -1,0 +1,57 @@
+'use strict';
+
+const boom = require('boom');
+const joi = require('joi');
+const schema = require('screwdriver-data-schema');
+const baseSchema = schema.models.command.base;
+
+module.exports = () => ({
+    method: 'PUT',
+    path: '/commands/{namespace}/{name}/trusted',
+    config: {
+        description: "Update a command's trusted property",
+        notes: 'Returns null if successful',
+        tags: ['api', 'commands', 'trusted'],
+        auth: {
+            strategies: ['token'],
+            scope: ['admin', '!guest']
+        },
+        plugins: {
+            'hapi-swagger': {
+                security: [{ token: [] }]
+            }
+        },
+        handler: async (request, reply) => {
+            const { name, namespace } = request.params;
+            const { commandFactory } = request.server.app;
+            const { trusted } = request.payload;
+
+            // get the earliest entry
+            const commands = await commandFactory.list({
+                params: { namespace, name },
+                paginate: { count: 1 },
+                sortBy: 'id',
+                sort: 'ascending'
+            });
+
+            if (commands.length === 0) {
+                throw boom.notFound(`Command ${namespace}/${name} does not exist`);
+            }
+
+            const command = commands[0];
+
+            command.trusted = trusted;
+
+            return command.update().then(() => reply().code(204), err => reply(boom.boomify(err)));
+        },
+        validate: {
+            params: {
+                namespace: joi.reach(baseSchema, 'namespace'),
+                name: joi.reach(baseSchema, 'name')
+            },
+            payload: {
+                trusted: joi.reach(baseSchema, 'trusted')
+            }
+        }
+    }
+});
--- a/plugins/templates/updateTrusted.js
+++ b/plugins/templates/updateTrusted.js
@@ -1,0 +1,56 @@
+'use strict';
+
+const boom = require('boom');
+const joi = require('joi');
+const schema = require('screwdriver-data-schema');
+const baseSchema = schema.models.template.base;
+
+module.exports = () => ({
+    method: 'PUT',
+    path: '/templates/{name}/trusted',
+    config: {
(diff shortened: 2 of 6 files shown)


DekusDenial:
test failed due to depending PR https://github.com/screwdriver-cd/data-schema/pull/340

tkyi:
@jithin1987 see https://github.com/screwdriver-cd/datastore-sequelize/pull/52

plugins/commands/updateTrusted.js
```diff
@@ -0,0 +1,53 @@
...
+const boom = require('boom');
+const joi = require('joi');
+const schema = require('screwdriver-data-schema');
+const baseSchema = schema.models.command.base;
+
+module.exports = () => ({
+    method: 'PUT',
+    path: '/commands/{namespace}/{name}/trusted',
+    config: {
+        description: "Update a command's trusted property",
```
tkyi: single quotes?

DekusDenial: eslint not complaining so i guess it’s ok with single q inside double q? unless u want me to backlash escape it?

DekusDenial:
@jithin1987 as per our discussion yesterday, once a template/command got marked trusted it remains that for life unless we explicitly remove that certification. The GET query is basically a group by query after all, as addressed by the other PR

plugins/commands/updateTrusted.js
tkyi: I don't mind but just thought it looked odd haha
So used to eslint complaining.

plugins/commands/updateTrusted.js
```diff
@@ -0,0 +1,53 @@
...
+            const { trusted } = request.payload;
+
+            const commands = await commandFactory.list({ params: { namespace, name } });
+
+            if (commands.length === 0) {
+                throw boom.notFound(`Command ${namespace}/${name} does not exist`);
+            }
+
+            return Promise.all(
+                commands.map((command) => {
```
tkyi: If I'm reading this correctly, looks like we plan to get **all** the commands with the name/namespace and we're going to go and set each command with the `trusted` value?
If we're checking the MAX value with datastore-sequelize, I think we can just set one of the command versions, right?

plugins/commands/updateTrusted.js
DekusDenial: haha, u should try coding in java, again, if never…

plugins/commands/updateTrusted.js
DekusDenial: but that mean u will need to find that one version when u gonna remove it which i rarely think that would ever happen once something got trusted

DekusDenial:
@tkyi not really famililar with the admin scope, but is `admin` not enough? The code seems like if you’re a sys admin, you’re also sd admin

plugins/commands/updateTrusted.js
tkyi: Well after calling this endpoint, any newly published command versions will not have `trusted: true`, anyways, right?

And for removing it, are we trying to use the same endpoint? Or will we make a separate one? Or just rely on a DB call?

I think it should be pretty easy to find the one version. We never delete commands/templates. You can pass in another param to `factory.list` to find it or construct a pretty simple SQL command I think.

tkyi:
I think the `scope` is for tokens. But I don't think we have `admin` scope in any tokens atm 😞 
That's another thing we need to fix

tkyi:
Okay actually we do push the `admin` scope: https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/buildClusters/create.js#L44-L52
Shouldn't need those other checks then

plugins/commands/updateTrusted.js
DekusDenial: ok, since u spotted my laziness, i guess i surrender….

DekusDenial:
just to clarify, SD admin will use these endpoints for (un)marking templates & commands. Even though they are shown in swagger doc but won't matter since they are protected by scope `admin`

plugins/templates/updateTrusted.js
```diff
@@ -0,0 +1,56 @@
...
+                security: [{ token: [] }]
+            }
+        },
+        handler: async (request, reply) => {
+            const { name } = request.params;
+            const { templateFactory } = request.server.app;
+            const { trusted } = request.payload;
+
+            // get the earliest entry
+            const templates = await templateFactory.list({
```
tkyi: not sure if this is any better (probably not): https://github.com/screwdriver-cd/models/blob/master/lib/templateFactory.js#L222

DekusDenial: i think the sole purpose here is to get the earliest one, dont really need a flexibility like that function there right?

PR #1695, opened 2019-07-08
fix: setup tokengen to executor

s-yoshika:
## Context
It's reported that sometimes periodicBulids aren't fired.
And we found following errors in sdapi logs.
```
error: failed to post build event for job xxx: TypeError: this.userTokenGen is not a function
```

The executor-queue sets up `userTokenGen` in `_startPeriodic` method,
https://github.com/screwdriver-cd/executor-queue/blob/master/index.js#L287-L289
and which is called by `job.update` ( and which is called by pipeline.sync etc..).

So if sdapi is restarted, the `userTokenGen` is null until the sdapi is requested to update any jobs, and also periodicBuilds don't work.
https://github.com/screwdriver-cd/executor-queue/blob/master/index.js#L48
https://github.com/screwdriver-cd/executor-queue/blob/master/index.js#L190

tokenGen was splited for purposes in executor-queue, `tokenGen` is for temporal scope and `userTokenGen` is for user scope but these property setup are delayed until `_start` and `_startPeriodic` is called. This change ensures the setup is done just when api is started.
## Objective
This ensures that the executor has `tokenGen` and `userTokengen` when sdapi is started.
I think these should be passed into the constructor of executor in the future, but it needs big refactoring.

If it is effective we remove following codes.
https://github.com/screwdriver-cd/executor-queue/blob/master/index.js#L287-L289
https://github.com/screwdriver-cd/executor-queue/blob/master/index.js#L431-L433

## License

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -130,20 +130,23 @@ module.exports = (config) => {
     return registrationMan(server, config)
         .then(() => {
             // Initialize common data in buildFactory and jobFactory
-            if (server.app.buildFactory) {
-                server.app.buildFactory.apiUri = server.info.uri;
-                server.app.buildFactory.tokenGen = (buildId, metadata, scmContext, expiresIn) =>
-                    server.plugins.auth.generateToken(server.plugins.auth
-                        .generateProfile(buildId, scmContext, ['temporal'], metadata), expiresIn
-                    );
-            }
-            if (server.app.jobFactory) {
-                server.app.jobFactory.apiUri = server.info.uri;
-                server.app.jobFactory.tokenGen = (username, metadata, scmContext) =>
-                    server.plugins.auth.generateToken(server.plugins.auth
-                        .generateProfile(username, scmContext, ['user'], metadata)
-                    );
-            }
+
+            server.app.buildFactory.apiUri = server.info.uri;
+            server.app.buildFactory.executor.tokenGen = server.app.buildFactory.tokenGen =
+                (buildId, metadata, scmContext, expiresIn) => server.plugins.auth.generateToken(
+                    server.plugins.auth.generateProfile(
+                      buildId, scmContext, ['temporal'], metadata
+                    ),
+                    expiresIn
+                );
+
+            server.app.jobFactory.apiUri = server.info.uri;
+            server.app.jobFactory.executor.userTokenGen = server.app.jobFactory.tokenGen =
+                (username, metadata, scmContext) => server.plugins.auth.generateToken(
+                    server.plugins.auth.generateProfile(
+                        username, scmContext, ['user'], metadata
+                    )
+                );
 
             // Start the server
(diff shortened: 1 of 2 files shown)


jithin1987:
https://github.com/screwdriver-cd/executor-queue/pull/55/files is where we introduced `userTokenGen`

s-yoshika:
Description is not enough. I added.

> tokenGen is splited for purposes in executor-queue, `tokenGen` is for temporal scope and `userTokenGen` is for user scope but these property setup are delayed when `_start` and `_startPeriodic` is called. This change ensures the setup is done just when api is started.

test/lib/server.test.js
```diff
@@ -185,7 +198,18 @@ describe('server case', () => {
         });
 
         it('doesnt affect non-errors', () => (
-            hapiEngine({ ecosystem }).then(server => (
+            hapiEngine({
+                ecosystem,
```
tk3fftk: I think this parameter object should be `const` or `let` and should be reusable.

s-yoshika: OK, I'll fix it.

lib/server.js
```diff
@@ -130,20 +130,20 @@ module.exports = (config) => {
     return registrationMan(server, config)
         .then(() => {
             // Initialize common data in buildFactory and jobFactory
-            if (server.app.buildFactory) {
```
d2lam: how come we no longer need if statement?

s-yoshika: `buildFactory` and `jobFactory` are always required for Screwdriver to work, so I think these if statements are no longer need. Please teach me if these statements are still need.

catto:
note: 
* [tokenGen and userTokenGen are null initially when the executor is setup.](https://github.com/screwdriver-cd/executor-queue/blob/66beb2430e62999c6a03fef8baf02a31f9945219/index.js#L47-L48)
* [`startPeriodic` expects tokenGen parameter to be passed and store it to `this.userTokenGen` if it's null.](https://github.com/screwdriver-cd/executor-queue/blob/66beb2430e62999c6a03fef8baf02a31f9945219/index.js#L286-L289)
* [`startPeriodic` actually receives `tokenGen` parameter from only models (job, jobFactory).](https://github.com/screwdriver-cd/models/blob/master/lib/job.js#L214)
* So, after api restarts, userTokenGen is null until `job.update` or `jobFactory.create` is called. These are called by pipline.sync for example.
* And [periodic builds are triggered by resque delayed job (startDelayed) but it doesn't pass tokenGen parameter to startPeriodic](https://github.com/screwdriver-cd/executor-queue/blob/66beb2430e62999c6a03fef8baf02a31f9945219/index.js#L84-L88)
* Therefore, any periodic builds, invoked from a API that haven't run normal builds or synced a pipeline since its start, aren't run with the error described in this PR.

PR #1700, opened 2019-07-16
fix: update github to octokit for functional test

kkokufud:
## Context
Github module is deprecated, but functest is used it.
Update github module to @octokit/rest for functest.

## Objective
- Update `github.js` and related tests for functest.
- Add `@octokit/rest` and delete `github` to `package.json`.

## License


kkokufud:
Thank you!

Issue #1393, opened 2018-11-21
Need more functional tests


tk3fftk:
Some functional tests are not implemented yet. If some tests are not important, it may be deletable.
Ignored or not existing tests are following:
- [ ] [some test cases of events](https://github.com/screwdriver-cd/screwdriver/blob/9f9f743327eb4b4956c16b1ccee3a8623eae9de2/features/events.feature#L41-L60) (these test cases may be able to move to workflow tests)
- [ ] [executor-queue](https://github.com/screwdriver-cd/screwdriver/blob/9f9f743327eb4b4956c16b1ccee3a8623eae9de2/features/executor-queue.feature)
- [ ] metadata (external meta feature)
- [x] [trigger](https://github.com/screwdriver-cd/screwdriver/blob/9f9f743327eb4b4956c16b1ccee3a8623eae9de2/features/trigger.feature)
- [ ] remote trigger, remote join
- [x] [workflow](https://github.com/screwdriver-cd/screwdriver/blob/9f9f743327eb4b4956c16b1ccee3a8623eae9de2/features/workflow.feature#L95-L142) (these will be implemented in https://github.com/screwdriver-cd/screwdriver/issues/1010)
- [ ] [template](https://github.com/screwdriver-cd/screwdriver/blob/9f9f743327eb4b4956c16b1ccee3a8623eae9de2/features/templates.feature#L51-L81)
- [ ] build cache
- [ ] build cluster
- [ ] [coverage-sonar](https://github.com/screwdriver-cd/screwdriver/issues/2547)

PR #1702, opened 2019-07-17
fix(1393): Add functional test for triggers

tkyi:
## Context

Missing trigger functional tests.

## Objective

This PR adds some triggers functional tests. Also updates to comply with deprecation: https://github.com/cucumber/cucumber-js/blob/master/CHANGELOG.md#deprecations

Issues:
- ~new parallel fork join feature breaks rule: external parallel triggered builds should be a part of the same event if triggered by the same build~ _fixed in https://github.com/screwdriver-cd/screwdriver/pull/1986_
- cannot delete and recreate pipeline as the "requires" triggers need to be hardcoded in the screwdriver.yaml

TODO:
- refactor to clean up code
- add parallel fork join tests?

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/1393

## License


features/step_definitions/trigger.js
```diff
@@ -0,0 +1,182 @@
...
+        shouldNotDeletePipeline: true });
+});
+
+When(/^the "(.*)" job on branch "(?:.*)" is started$/,
+    { timeout: TIMEOUT }, function step(jobName) {
+        let jobId;
+        let buildVarName;
+
+        switch (jobName) {
+        case 'success_A':
```
klu909: can we use hashmaps here instead so we don't need to have long switch like this

klu909: one hashmap to map`fail_A` to `successAJobId` and `successABuildId`?

tkyi: updated

tkyi:
@jithine https://github.com/screwdriver-cd-test/functional-trigger/blob/pipelineB/screwdriver.yaml

Issue #947, opened 2018-03-27
Add user-created pipelines into a user's collection automatically


tkyi:
## Context
If a user creates a pipeline, they probably want an easy way to keep track of it or navigate to it.

## Proposal
Users should all have a "My Pipelines" collection (or "Favorites") by default. When a user creates a pipeline, that pipeline should be automatically added to their "My Pipelines" collection.

drewfish:
... or perhaps "My Creations"

edudar:
... or just a (multi-?)selector for collections on a pipeline creation screen. There's plenty of space to fill.

PR #1733, opened 2019-08-30
feat(947): Implement user default collection

Code-Beast:
## Context

Would be nice if users can find all pipeline create by themselves in a default collection.

## Objective

This PR creates a default collection ("My Pipelines") where all pipelines a user creates will be stored when the user create the first pipeline and prevents the creation or renaming of default collection using API. (This PR comes with a change in the schema of collection => new field: type).

## References

Related to [947](https://github.com/screwdriver-cd/screwdriver/issues/947)

## License


features/collections.feature
```diff
@@ -18,6 +18,11 @@ Feature: User Collections
 #            | miss wormwood | no access   |
         And an existing pipeline with that repository
 
+    Scenario: Check Default Collection
+        And "calvin" is logged in
+        When they check the default collection
+        Then they can see the default collection contains that pipeline
```
tkyi: what pipeline?

should the scenario include something like 
```
And "calvin" is logged in
When they create a new pipeline
Then they can see the default collection contains that pipeline
```

plugins/auth/login.js
```diff
@@ -101,11 +102,13 @@ function addOAuthRoutes(server, config) {
                 );
                 request.cookieAuth.set(profile);
 
-                return factory.get({ username, scmContext })
+                return userFactory.get({ username, scmContext })
                     // get success, so user exists
                     .then((model) => {
                         if (!model) {
-                            return factory.create({
+                            // TODO: Move default collection creation here after database migration
```
tkyi: not sure if this is needed since the only time a default collection is really useful is when the user's pipelines are autopopulated in it

plugins/pipelines/create.js
```diff
@@ -69,22 +69,51 @@ module.exports = () => ({
...
-                            protocol: request.server.info.protocol,
-                            pathname: `${request.path}/${pipeline.id}`
-                        });
+                        })
+                        // get the default collection for current user
+                        .then(pipeline =>
+                            collectionFactory.get({
+                                userId: user.id,
+                                type: 'default',
+                                name: 'My Pipelines'
```
tkyi: don't need to have this param

features/collections.feature
tkyi: Nevermind, I see that the ensurePipelineExists method creates a pipeline

plugins/auth/login.js
tkyi: I see the design calls for a default collection to exist no matter what

plugins/pipelines/create.js
Code-Beast: Deleted the name param.

plugins/auth/index.js
```diff
@@ -174,6 +175,25 @@ exports.register = (server, options, next) => {
                                             return Promise.reject();
                                         }
 
+                                        const description =
+                                            `The default collection for ${user.username}`;
+
+                                        collectionFactory.get({
+                                            userId: user.id,
+                                            type: 'default',
+                                            name: 'My Pipelines'
```
tkyi: hm this is still here

plugins/pipelines/create.js
tkyi: still seeing `name` param here

Issue #1747, opened 2019-09-16
sd incorrectly excludes commit from build


klu909:
**What happened**:
build not trigger when there's commit made and sent by automated system
**What you expected to happen**:
build should run for committer that's not in the ignoreList
**How to reproduce it**:

PR #1753, opened 2019-09-18
fix(1747): add a new condition to skip ci check

klu909:
… build

## Context

sd incorrectly excludes commit from build when commit is made but is sent by user in ignoreUser

## Objective

add a new condition to ci skip check to prevent SD from skipping legitimate build

## References

https://github.com/screwdriver-cd/screwdriver/issues/1747

## License

### DIFF
--- a/plugins/webhooks/index.js
+++ b/plugins/webhooks/index.js
@@ -837,11 +837,18 @@ exports.register = (server, options, next) => {
                     }
 
                     // if skip ci then don't return
-                    if (ignoreUser && ignoreUser.includes(username) && !skipMessage) {
-                        message = `Skipping because user ${username} is ignored`;
-                        request.log(['webhook', hookId], message);
+                    if (ignoreUser && !skipMessage) {
+                        const commitAuthors = Array.isArray(parsed.commitAuthors) ?
+                            parsed.commitAuthors : [username];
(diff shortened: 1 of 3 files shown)


plugins/webhooks/index.js
```diff
@@ -836,8 +837,14 @@ exports.register = (server, options, next) => {
                         skipMessage = 'Skipping due to the commit message: [skip ci]';
                     }
 
+                    if (action === 'push' && !skipMessage && parsed.commitAuthors) {
```
tkyi: can probably get `commitAuthors` var on line 829 along with all the other keys in `parsed`

plugins/webhooks/index.js
```diff
@@ -836,8 +838,14 @@ exports.register = (server, options, next) => {
                         skipMessage = 'Skipping due to the commit message: [skip ci]';
                     }
 
+                    if (action === 'push' && !skipMessage && commitAuthors) {
```
jithin1987: this should be something like 
```
commitAuthors = Array.isArray(commitAuthors) ? commitAuthors : [username]
validCommitAuthors = commitAuthors.filter(author => !ignoreUser.includes(author));

if (!validCommitAuthors.length && !skipMessage) {
   return ...
}
```

Issue #1664, opened 2019-06-03
sequalize sync with ALTER blocks all MySQL operations


tk3fftk:
**What happened**:
We use MySQL as a sequelize dialect. 
`datastore-sequalize` setups with [`sync({ alter: true })`](https://github.com/screwdriver-cd/datastore-sequelize/blob/d15c8b1925c6b3284974fbcd8332e7a43285ccaf/index.js#L220) so all ALTER queries are executed in every SD API Pod setup including Pod Restart, Rolling Update for Deployment without schema change. 
Then our MySQL sometimes become disk busy and all operation are blocked by ALTER queries.

**What you expected to happen**:
`alter` parameter of sequalize sync should be optional.

**References:**
- `alter: true` option did not exist at first ( https://github.com/screwdriver-cd/datastore-sequelize/commit/4e9768e6c258ce7b0e5f839ab50b500240715f1b ) 
- [sequalize docs](http://docs.sequelizejs.com/class/lib/sequelize.js~Sequelize.html#instance-method-sync)
    - > Alters tables to fit models. Not recommended for production use.
- https://github.com/screwdriver-cd/screwdriver/issues/1298

jithin1987:
https://github.com/screwdriver-cd/data-schema/pull/352

PR #1756, opened 2019-09-19
feat(1664): make sequalize ddl sync optional

parthasl:
## Context

Screwdriver database DDL changes are applied during Screwdriver api startup, which leads to api experiencing slowness or goes to the unreachable state. This does table-level exclusive locks because of 1. DDL statements are executed in parallel from multiple pods and 2. ALTER statements are executed even though no changes made to the database schema. Root-cause for this issue is calling Sequelize sync method with { alter: true } parameter.

## Objective

This PR addresses sequelize ddl sync required or optional based on an environment variable setting.

## References

https://github.com/screwdriver-cd/screwdriver/issues/1664

## License


config/default.yaml
```diff
@@ -56,6 +56,7 @@ httpd:
 
 datastore:
   plugin: sequelize
+  ddlSyncEnabled: "true"
```
tk3fftk: Why are we using `string`?

parthasl: followed buildMetricsEnabled. I think boolean will not work in k8s env variables

Issue #1754, opened 2019-09-19
Enable to on/off coverage plugin in each job


sakka2:
**What happened**:
When we use coverage-bookend, coverage scanning is executed all jobs, including jobs which is not necessary to scan.
It must be leads to longer build time, or surging database size.

**What you expected to happen**:
- Add cluster config `coverage.default`. If this config value is true, coverage-bookend will execute coverage scanning as default. If it is false, no scanning as default.
- Default value of `coverage.default` is `true`, so it will not be affected present system.
- Enable SD users to set `SD_COVERAGE_PLUGIN_ENABLED` environment variable (`true` / `false`) in screwdriver.yaml to execute coverage-bookend or not in each job.
- If `SD_COVERAGE_PLUGIN_ENABLED` environment variable is set, follow its setting and execute bookend or not.
- If `SD_COVERAGE_PLUGIN_ENABLED` environment variable is not set, follow cluster level config `coverage.default`.

PR #1758, opened 2019-09-20
fix(1754): [1] Add cluster level coverage plugin setting (enable or not)

sakka2:
## Context
When we use coverage-bookend, coverage scanning is executed all jobs, including jobs which is not necessary to scan.
It must be leads to longer build time, or surging database size.

## Objective
- Add cluster level config of coverage plugin (`default`: `true` / `false`）
- Default value is `true`.

## References
Issue: https://github.com/screwdriver-cd/screwdriver/issues/1754
Other PRs
[2] https://github.com/screwdriver-cd/coverage-bookend/pull/4
[3] https://github.com/screwdriver-cd/coverage-base/pull/11

## License
I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

### DIFF
--- a/config/custom-environment-variables.yaml
+++ b/config/custom-environment-variables.yaml
@@ -340,7 +340,7 @@ notifications:
 
 coverage:
   plugin: COVERAGE_PLUGIN
-  enabled: COVERAGE_PLUGIN_DEFAULT_ENABLED
+  default: COVERAGE_PLUGIN_DEFAULT_ENABLED
   sonar:
     # Screwdriver API url
     sdApiUrl: URI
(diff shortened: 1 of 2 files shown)


config/custom-environment-variables.yaml
```diff
@@ -340,6 +340,7 @@ notifications:
 
 coverage:
   plugin: COVERAGE_PLUGIN
+  enabled: COVERAGE_PLUGIN_ENABLED
```
tk3fftk: I think `COVERAGE_PLUGIN_ENABLED` looks "we can use coverage plugin or not". 
Do we use `default: COVERAGE_PLUGIN_DEFAULT_ENABLED` or another one?

PR #1759, opened 2019-09-20
fix: Webhook 500 error when nothing branch

sugarnaoming:
## Context
- When the branch is got, so it might be unexpected branch because it was got by forward matching.
```
  e.g.
  expected branch is 'branch-01', but also get 'branch-011'
    branch-01
    branch-011
```
- fixed if `getCommitSha` function throws error when pipeline exists and scm branch of pipeline not exists, all events is not created and not work.
- fixed no-await-in-loop in ESLint rule to improve processing performance by parallel processing.

## Objective
- fix forward matching to perfect matching for `triggeredPipelines`
- fix catch error of `getCommitSha`
- fix no-await-in-loop for `createPREvents` and `createEvents`

## License

### DIFF
--- a/plugins/webhooks/index.js
+++ b/plugins/webhooks/index.js
@@ -299,17 +299,13 @@ async function createPREvents(options, request) {
     const pipelineFactory = request.server.app.pipelineFactory;
     const scmDisplayName = scm.getDisplayName({ scmContext: scmConfig.scmContext });
     const userDisplayName = `${scmDisplayName}:${username}`;
-    const events = [];
 
     scmConfig.prNum = prNum;
 
-    for (let i = 0; i < pipelines.length; i += 1) {
-        const p = pipelines[i];
-        /* eslint-disable no-await-in-loop */
+    return Promise.all(pipelines.map(async (p) => {
         const b = await p.branch;
         // obtain pipeline's latest commit sha for branch specific job
         const configPipelineSha = await pipelineFactory.scm.getCommitSha(scmConfig);
-        /* eslint-enable no-await-in-loop */
 
         const { skipMessage, resolvedChainPR } = getSkipMessageAndChainPR({
             pipeline: p,
@@ -333,7 +329,6 @@ async function createPREvents(options, request) {
             prRef,
             prNum,
             prTitle,
-            // eslint-disable-next-line no-await-in-loop
             prInfo: await eventFactory.scm.getPrInfo(scmConfig)
         };
 
@@ -345,10 +340,8 @@ async function createPREvents(options, request) {
             eventConfig.startFrom = '~pr';
         }
 
-        events.push(eventFactory.create(eventConfig));
-    }
-
-    return Promise.all(events);
+        return eventFactory.create(eventConfig);
+    }));
 }
 
 /**
@@ -637,53 +630,64 @@ async function createEvents(eventFactory, userFactory, pipelineFactory,
     const events = [];
     const meta = createMeta(parsed);
 
-    for (let i = 0; i < pipelines.length; i += 1) {
-        const p = pipelines[i];
-        /* eslint-disable no-await-in-loop */
-        const pipelineBranch = await p.branch;
-        /* eslint-enable no-await-in-loop */
-        const startFrom = determineStartFrom(action, type, branch, pipelineBranch);
+    const pipelineTaple = await Promise.all(pipelines.map(async (p) => {
+        const taple = { branch: await p.branch, pipeline: p };
+
+        return taple;
+    }));
+
+    const ignoreExtraTriggeredPipelines = pipelineTaple.filter((taple) => {
+        const startFrom = determineStartFrom(action, type, branch, taple.branch);
 
         // empty event is not created when it is triggered by extra triggers (e.g. ~tag, ~release)
-        if (EXTRA_TRIGGERS.test(startFrom) && !hasTriggeredJob(p, startFrom)) {
+        if (EXTRA_TRIGGERS.test(startFrom) && !hasTriggeredJob(taple.pipeline, startFrom)) {
             winston.info(`Event not created: there are no jobs triggered by ${startFrom}`);
-        } else {
-            /* eslint-disable no-await-in-loop */
-            const token = await p.token;
-            const scmConfig = {
-                scmUri: p.scmUri,
-                token,
-                scmContext
-            };
-            // obtain pipeline's latest commit sha for branch specific job
-            const configPipelineSha = await pipelineFactory.scm.getCommitSha(scmConfig);
-            /* eslint-enable no-await-in-loop */
-            const eventConfig = {
-                pipelineId: p.id,
-                type: 'pipeline',
-                webhooks: true,
-                username,
-                scmContext,
-                startFrom,
-                sha,
-                configPipelineSha,
-                changedFiles,
-                commitBranch: branch,
-                causeMessage: `Merged by ${username}`,
-                meta
-            };
 
-            if (skipMessage) {
-                eventConfig.skipMessage = skipMessage;
-            }
-
-            /* eslint-disable no-await-in-loop */
-            await updateAdmins(userFactory, username, scmContext, p);
-            /* eslint-enable no-await-in-loop */
-
-            events.push(eventFactory.create(eventConfig));
+            return false;
         }
-    }
+
+        return true;
+    });
+
+    const eventConfigs = await Promise.all(ignoreExtraTriggeredPipelines.map(async (igpset) => {
+        const resoleved = await igpset;
+        const pipelineBranch = resoleved.branch;
+        const startFrom = determineStartFrom(action, type, branch, pipelineBranch);
+        const token = await resoleved.pipeline.token;
+        const scmConfig = {
+            scmUri: resoleved.pipeline.scmUri,
+            token,
+            scmContext
+        };
+        // obtain pipeline's latest commit sha for branch specific job
+        const configPipelineSha = await pipelineFactory.scm.getCommitSha(scmConfig);
+        const eventConfig = {
+            pipelineId: resoleved.pipeline.id,
+            type: 'pipeline',
+            webhooks: true,
+            username,
+            scmContext,
+            startFrom,
+            sha,
+            configPipelineSha,
+            changedFiles,
+            commitBranch: branch,
+            causeMessage: `Merged by ${username}`,
+            meta
+        };
+
+        if (skipMessage) {
+            eventConfig.skipMessage = skipMessage;
+        }
+
+        await updateAdmins(userFactory, username, scmContext, resoleved.pipeline);
+
+        return eventConfig;
+    }));
+
+    eventConfigs.forEach((eventConfig) => {
+        events.push(eventFactory.create(eventConfig));
+    });
 
     return Promise.all(events);
 }
--- a/plugins/webhooks/index.js
+++ b/plugins/webhooks/index.js
@@ -844,11 +844,18 @@ exports.register = (server, options, next) => {
                     }
 
                     // if skip ci then don't return
-                    if (ignoreUser && ignoreUser.includes(username) && !skipMessage) {
-                        message = `Skipping because user ${username} is ignored`;
-                        request.log(['webhook', hookId], message);
+                    if (ignoreUser && !skipMessage) {
+                        const commitAuthors = Array.isArray(parsed.commitAuthors) ?
+                            parsed.commitAuthors : [username];
+                        const validCommitAuthors = commitAuthors.filter(author =>
+                            !ignoreUser.includes(author));
 
-                        return reply({ message }).code(204);
+                        if (!validCommitAuthors.length) {
+                            message = `Skipping because user ${username} is ignored`;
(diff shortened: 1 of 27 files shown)


plugins/webhooks/index.js
```diff
@@ -253,7 +253,10 @@ async function triggeredPipelines(pipelineFactory, scmConfig, branch, type, acti
     let pipelinesOnOtherBranch = [];
 
     pipelines.forEach((p) => {
-        if (p.scmUri.startsWith(scmUri)) {
+        const tmpScmUri = p.scmUri.split(':');
+        const tmpScmBranch = `${tmpScmUri[0]}:${tmpScmUri[1]}:${tmpScmUri[2]}`;
+
+        if (tmpScmBranch === scmUri) {
```
sakka2: [ask] Why cannot we just say `p.scmUri === scmUri` ?
Is there any case `${tmpScmUri[3]}` ?

sugarnaoming: This check is until `branch_name`.
because 'root_dir' suffix may have been appended, so `root_dir` excluded

`scmUri` structure is here 
`scm_uri_domain:repo_id:branch_name:root_dir`

sakka2: I see. You mean that `${tmpScmUri[3]}` will be the `root_dir`?

sugarnaoming: yes

plugins/webhooks/index.js
```diff
@@ -634,53 +630,64 @@ async function createEvents(eventFactory, userFactory, pipelineFactory,
...
-            /* eslint-enable no-await-in-loop */
+            return false;
+        }
+
+        return true;
+    });
 
-            events.push(eventFactory.create(eventConfig));
+    const eventConfigs = await Promise.all(ignoreExtraTriggeredPipelines.map(async (igpset) => {
+        const resoleved = await igpset;
```
tkyi:
```suggestion
        const resolved = await igpset;
```

plugins/webhooks/index.js
```diff
@@ -634,53 +630,64 @@ async function createEvents(eventFactory, userFactory, pipelineFactory,
     const events = [];
     const meta = createMeta(parsed);
 
-    for (let i = 0; i < pipelines.length; i += 1) {
-        const p = pipelines[i];
-        /* eslint-disable no-await-in-loop */
-        const pipelineBranch = await p.branch;
-        /* eslint-enable no-await-in-loop */
-        const startFrom = determineStartFrom(action, type, branch, pipelineBranch);
+    const pipelineTaple = await Promise.all(pipelines.map(async (p) => {
```
tkyi: what is a taple? 🤔

s-yoshika:
Do we need to add test cases for this change?

plugins/webhooks/index.js
```diff
@@ -634,53 +630,64 @@ async function createEvents(eventFactory, userFactory, pipelineFactory,
...
+        const taple = { branch: await p.branch, pipeline: p };
+
+        return taple;
+    }));
+
+    const ignoreExtraTriggeredPipelines = pipelineTaple.filter((taple) => {
```
s-yoshika: `pipelineTable` is needed? Can we combine with L633-639 into one as below?
`const filteredPipelines = await Promise.all(pipelines.filter(async(p) => {...}))`

plugins/webhooks/index.js
```diff
@@ -253,7 +253,10 @@ async function triggeredPipelines(pipelineFactory, scmConfig, branch, type, acti
     let pipelinesOnOtherBranch = [];
 
     pipelines.forEach((p) => {
-        if (p.scmUri.startsWith(scmUri)) {
+        const tmpScmUri = p.scmUri.split(':');
```
tk3fftk: I think we should add a comment why splitting and concatenation of `scmUri`.

tk3fftk:
Do we need to add any test cases for ensuring this fix?

plugins/webhooks/index.js
kkokufud: Thanks, I fixed.

plugins/webhooks/index.js
kkokufud: It was misspeling of `tuple`.
`tuple` means just a combination in this codes.

plugins/webhooks/index.js
kkokufud: Unfortunately, `filter` seems to not supported async. `filter` needs `boolean`, but it returns `Promise`.

plugins/webhooks/index.js
kkokufud: Thanks, I added comment.

plugins/webhooks/index.js
s-yoshika: I got it, thanks!

kkokufud:
> Do we need to add test cases for this change?
> Do we need to add any test cases for ensuring this fix?

Thanks, added test case.

plugins/webhooks/index.js
```diff
@@ -634,53 +631,64 @@ async function createEvents(eventFactory, userFactory, pipelineFactory,
     const events = [];
     const meta = createMeta(parsed);
 
-    for (let i = 0; i < pipelines.length; i += 1) {
-        const p = pipelines[i];
-        /* eslint-disable no-await-in-loop */
-        const pipelineBranch = await p.branch;
```
s-yoshika: It is good to follow eslint, but `pipeline.branch` seems not to use asynchronous processing? I think it just wraps synchronouse processing with Promise(my understanding may be something wrong).
https://github.com/screwdriver-cd/models/blob/master/lib/pipeline.js#L925

So, I think fixing those doesn't improve performance than we think.
[IMO] And I like small changes if it's needless.

s-yoshika: Never mind, I found there are some await statement except of `p.branch` in this loop.

test/plugins/webhooks.test.js
```diff
@@ -970,6 +970,70 @@ describe('webhooks plugin test', () => {
                 });
             });
 
+            it('returns 201 on success for pipelines when mixed forward matching branch', () => {
+                const pMock1 = {
```
tk3fftk: Is this `pMock1` needed? It looks similar with `pipelineMock`.

kkokufud: oh, deleted unnecessary mock.

sugarnaoming:
fixed if getCommitSha function throws error when pipeline exists and scm branch of pipeline not exists, all events is not created and not work.

plugins/webhooks/index.js
```diff
@@ -253,7 +253,11 @@ async function triggeredPipelines(pipelineFactory, scmConfig, branch, type, acti
     let pipelinesOnOtherBranch = [];
 
     pipelines.forEach((p) => {
-        if (p.scmUri.startsWith(scmUri)) {
+        // This uri expects 'scmUriDomain:repoId:branchName:rootDir'. To Compare, rootDir is ignored.
+        const tmpScmUri = p.scmUri.split(':');
+        const tmpScmBranch = `${tmpScmUri[0]}:${tmpScmUri[1]}:${tmpScmUri[2]}`;
```
tk3fftk: I think we should rename this parameter for more information.

plugins/webhooks/index.js
tk3fftk: And this comparing is not enough.
If `scmUri` is `github.com:1234:master:/somedir` and  `p.scmUri` is`github.com:1234:master`,  then this expression will be `false` and trigger with `pipelinesOnOtherBranch`.
So we should compare with `branch`. These pipelines must have same `scmRepoId`.

plugins/webhooks/index.js
```diff
@@ -634,53 +631,64 @@ async function createEvents(eventFactory, userFactory, pipelineFactory,
...
-            /* eslint-enable no-await-in-loop */
+            return false;
+        }
+
+        return true;
+    });
 
-            events.push(eventFactory.create(eventConfig));
+    const eventConfigs = await Promise.all(ignoreExtraTriggeredPipelines.map(async (igpset) => {
+        const resolved = await igpset;
```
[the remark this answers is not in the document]
sugarnaoming: thank you. I renamed and removed unnecessary await.

plugins/webhooks/index.js
```diff
@@ -300,14 +304,16 @@ async function createPREvents(options, request) {
...
         const b = await p.branch;
         // obtain pipeline's latest commit sha for branch specific job
-        const configPipelineSha = await pipelineFactory.scm.getCommitSha(scmConfig);
-        /* eslint-enable no-await-in-loop */
+        let configPipelineSha = '';
 
+        try {
+            configPipelineSha = await pipelineFactory.scm.getCommitSha(scmConfig);
+        } catch (err) {
+            winston.info(`branch: ${b} ${err.message}`);
```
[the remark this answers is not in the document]
sugarnaoming: Thank you. Add 'skip create event' message  instead of error message.

plugins/webhooks/index.js
sugarnaoming: Thank you. I renamed.

sugarnaoming:
> And this comparing is not enough.
If scmUri is github.com: 1234:master:/somedir and p.scmUri isgithub.com: 1234:master, then this expression will be false and trigger with pipelinesOnOtherBranch.
So we should compare with branch. These pipelines must have same scmRepoId.

Thank you. I compared it strictly.

sugarnaoming:
> How about using conditional operator?

Thank you.
I did not adopted suggestion because conditional operator was long, max-len of eslint rule is 100.

plugins/webhooks/index.js
```diff
@@ -300,14 +305,16 @@ async function createPREvents(options, request) {
...
         const b = await p.branch;
         // obtain pipeline's latest commit sha for branch specific job
-        const configPipelineSha = await pipelineFactory.scm.getCommitSha(scmConfig);
-        /* eslint-enable no-await-in-loop */
+        let configPipelineSha = '';
 
+        try {
+            configPipelineSha = await pipelineFactory.scm.getCommitSha(scmConfig);
+        } catch (err) {
+            winston.info(`branch: ${b} skip create event`);
```
tk3fftk: How about the msg? 
```suggestion
            winston.info(`skip create event for branch: ${b}`);
```

kkokufud: Thanks, fixed messages.

plugins/webhooks/index.js
```diff
@@ -300,14 +305,16 @@ async function createPREvents(options, request) {
...
         const b = await p.branch;
         // obtain pipeline's latest commit sha for branch specific job
-        const configPipelineSha = await pipelineFactory.scm.getCommitSha(scmConfig);
-        /* eslint-enable no-await-in-loop */
+        let configPipelineSha = '';
 
+        try {
+            configPipelineSha = await pipelineFactory.scm.getCommitSha(scmConfig);
+        } catch (err) {
+            winston.info(`skip create event for branch: ${b}`);
```
jithin1987: what's the implication of ignoring this error ?, do we need to differentiate between branch does not exist vs any error in talking to git ?

plugins/webhooks/index.js
```diff
@@ -634,53 +646,71 @@ async function createEvents(eventFactory, userFactory, pipelineFactory,
...
-        /* eslint-enable no-await-in-loop */
-        const startFrom = determineStartFrom(action, type, branch, pipelineBranch);
+    const pipelineTuples = await Promise.all(pipelines.map(async (p) => {
+        const tuple = { branch: await p.branch, pipeline: p };
+
+        return tuple;
+    }));
+
+    const ignoreExtraTriggeredPipelines = pipelineTuples.filter((tuple) => {
+        const startFrom = determineStartFrom(action, type, branch, tuple.branch);
```
jithin1987: can you add this `startFrom` to the tupe so that you can avoid computing this again at https://github.com/screwdriver-cd/screwdriver/pull/1759/files#diff-4bff1c7e1da44fb8e755afbf06260240R670

plugins/webhooks/index.js
```diff
@@ -634,53 +646,71 @@ async function createEvents(eventFactory, userFactory, pipelineFactory,
...
+            token,
+            scmContext
+        };
+        // obtain pipeline's latest commit sha for branch specific job
+        let configPipelineSha = '';
+
+        try {
+            configPipelineSha = await pipelineFactory.scm.getCommitSha(scmConfig);
+        } catch (err) {
+            winston.info(`skip create event for branch: ${pipelineBranch}`);
```
jithin1987: what's the implication of ignoring this error ?, do we need to differentiate between branch does not exist vs any error in talking to git ?

plugins/webhooks/index.js
kkokufud: Thank you. I fixed.

plugins/webhooks/index.js
kkokufud: It intend to ignore 'branch does not exist'. And 5xx error should not be ignored, so I fixed.

plugins/webhooks/index.js
kkokufud: It intend to ignore 'branch does not exist'. And 5xx error should not be ignored, so I fixed.

Issue #1701, opened 2019-07-17
PR authors should not be able to restart builds if `restrictPR` is on


jithin1987:
**What happened**:
By [allowing PR authors to restart builds](https://github.com/screwdriver-cd/screwdriver/issues/1246), we weakened security protection offered by `restrictPR`

**What you expected to happen**:

PR author should not be allowed to restart PR if `restrictPR` [annotation](https://docs.screwdriver.cd/user-guide/configuration/annotations.html#pipeline-level-annotations) is turned on. 

**How to reproduce it**:

1. Enable [restrict PR](https://docs.screwdriver.cd/user-guide/configuration/annotations.html#pipeline-level-annotations)
1. Create a PR from forked repository.
1. PR author has permission to start the build

PR #1762, opened 2019-09-23
fix(1701): PR authors should not be able to restart builds if `restrictPR` is on

klu909:
## Context

PR author should not be allowed to restart PR if restrictPR annotation is turned on.

## Objective

Add a check to enforce restrictPR before allowing PR author to restart builds

## References

https://github.com/screwdriver-cd/screwdriver/issues/1701

## License

### DIFF
--- a/plugins/events/create.js
+++ b/plugins/events/create.js
@@ -3,6 +3,7 @@
 const boom = require('boom');
 const urlLib = require('url');
 const validationSchema = require('screwdriver-data-schema');
+const ANNOT_RESTRICT_PR = 'screwdriver.cd/restrictPR';
 
 module.exports = () => ({
     method: 'POST',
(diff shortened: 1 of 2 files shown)


plugins/events/create.js
```diff
@@ -159,9 +162,11 @@ module.exports = () => ({
                                         payload.prInfo = prInfo;
                                         payload.prRef = prInfo.ref;
                                         payload.chainPR = pipeline.chainPR;
+                                        const restrictPR = pipeline.annotations[ANNOT_RESTRICT_PR];
 
-                                        // PR author should be able to rerun their own PR build
-                                        if (prInfo.username === username) {
+                                        // PR author should be able to rerun their own PR build if restrictPR is not on
```
jithin1987: there is a value `fork` also for restrictPR

plugins/events/create.js
```diff
@@ -4,6 +4,9 @@ const boom = require('boom');
 const urlLib = require('url');
 const validationSchema = require('screwdriver-data-schema');
 
+const ANNOT_NS = 'screwdriver.cd';
+const ANNOT_RESTRICT_PR = `${ANNOT_NS}/restrictPR`;
```
jithin1987: we don't need ANNOT_NS to be a separate variable

plugins/events/create.js
klu909: can you elaborate on this?

tkyi: https://docs.screwdriver.cd/user-guide/configuration/annotations#pipeline-level-annotations

Issue #1717, opened 2019-08-09
latest successful build


uevilwizard:
Builds are accessible at `/pipelines/<pipeline_id>/builds/<build_id>`, but often I don't want to see a particular build number: I want to see the most recent successful build.  

Can we have a keyword like 'lastSuccessfulBuild' that could be used in place of `<build-id>` ?

jithin1987:
This is similar to Badges API badges `/pipelines/ID/job_name/badge` so I propose we add an end point like thiss ``/pipelines/ID/job_name/latest?status=SUCCESS` which `status` an optional query parameter

jithin1987:
we not have /jobs/:jobID/latestBuild, how ever it will be useful to have `/pipelines/ID/job_name/latestBuild?status=SUCCESS` as well else users will have to make an extra API call to translate job name to ID, or hardcode the information in their pipeline.

adong:
How about the following urls with tags(revision)?

```
1) /pipelines/1/job/main/latest
2) /pipelines/1/job/main/latest?status=SUCCESS
3) /pipelines/1/job/main/latest?status=FAILURE
4) /pipelines/1/job/main/3?status=SUCCESS
5) /pipelines/1/job/main/4?status=FAILURE
6) /pipelines/1/job/main/-1?status=SUCCESS
```

This helps Users to explore

1) latest job named `main` from pipeline 1, which defaults to status `SUCCESS`
2) latest job named `main` from pipeline 1, which status is `SUCCESS`
3) latest job named `main` from pipeline 1, which status is `FAILURE`
4) job named `main` from pipeline 1, 3rd revision that has status `SUCCESS`
5) job named `main` from pipeline 1, 4th revision that has status `FAILURE`
6) job named `main` from pipeline 1, `second`-to-last revision that has status `SUCCESS`


I'd think 6) is going above and beyond for user experience.

PR #1771, opened 2019-10-01
feat(1717): latest build

klu909:
## Context

user request a new endpoint to see the most recent build

## Objective

add an end point that returns latest build

## References

https://github.com/screwdriver-cd/screwdriver/issues/1717

## License

### DIFF
--- a/plugins/jobs/latestBuild.js
+++ b/plugins/jobs/latestBuild.js
@@ -1,0 +1,58 @@
+'use strict';
+
+const boom = require('boom');
+const joi = require('joi');
+const schema = require('screwdriver-data-schema');
+const idSchema = joi.reach(schema.models.job.base, 'id');
+const statusSchema = joi.reach(schema.models.build.base, 'status');
+
+module.exports = () => ({
+    method: 'GET',
+    path: '/jobs/{id}/latestBuild',
+    config: {
+        description: 'Get latest build for a given job',
+        notes: 'Return latest build of status specified',
+        tags: ['api', 'job', 'build'],
+        auth: {
+            strategies: ['token'],
+            scope: ['user', 'build', 'pipeline']
+        },
+        plugins: {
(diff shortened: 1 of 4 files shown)


plugins/jobs/latestBuild.js
```diff
@@ -0,0 +1,58 @@
...
+                .then((job) => {
+                    if (!job) {
+                        throw boom.notFound('Job does not exist');
+                    }
+
+                    return job.getLatestBuild({ status });
+                })
+                .then((build) => {
+                    if (Object.getOwnPropertyNames(build).length === 0) {
+                        return reply({});
```
d2lam: I'm also unsure whether to return `{}` or `notFound` here. Want to hear others' opinion on this

tkyi: Not sure as well, but I'm okay with `{}`

jithin1987: I think 404 is semantically correct. With empty object, a client now will have to check not only http status but also look into the return data to see if there is actual result  404 eliminates this problem.

jithin1987: I would say return 404 with body saying `{status: "build does not exist"}` (consistent with responses in other places)

plugins/jobs/latestBuild.js
```diff
@@ -0,0 +1,58 @@
...
+            return jobFactory.get(request.params.id)
+                .then((job) => {
+                    if (!job) {
+                        throw boom.notFound('Job does not exist');
+                    }
+
+                    return job.getLatestBuild({ status });
+                })
+                .then((build) => {
+                    if (Object.getOwnPropertyNames(build).length === 0) {
```
adong: `Object.keys()`

klu909: changed done

PR #1775, opened 2019-10-04
feat(1717): get latest build api for pipeline

klu909:
## Context

Previously we added an endpoint for that returns latest build for a given job using job id, but user might not have job id so they would need to make a separate call to get job id. 

## Objective

Add another endpoint that would retrieve latest build using job name and pipeline id

## References

https://github.com/screwdriver-cd/screwdriver/issues/1717

## License

### DIFF
--- a/plugins/pipelines/latestBuild.js
+++ b/plugins/pipelines/latestBuild.js
@@ -9,7 +9,7 @@ const statusSchema = joi.reach(schema.models.build.base, 'status');
 
 module.exports = () => ({
     method: 'GET',
-    path: '/pipelines/{id}/{jobName}/latestBuild',
+    path: '/pipelines/{id}/jobs/{jobName}/latestBuild',
     config: {
         description: 'Get latest build for a given job',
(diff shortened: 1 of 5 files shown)


plugins/pipelines/latestBuild.js
```diff
@@ -0,0 +1,63 @@
...
+const boom = require('boom');
+const joi = require('joi');
+const schema = require('screwdriver-data-schema');
+const idSchema = joi.reach(schema.models.job.base, 'id');
+const nameSchema = joi.reach(schema.models.job.base, 'name');
+const statusSchema = joi.reach(schema.models.build.base, 'status');
+
+module.exports = () => ({
+    method: 'GET',
+    path: '/pipelines/{id}/{jobName}/latestBuild',
```
tkyi: since we already have `GET /pipelines/{id}/jobs?archived={boolean}`, wondering if it would be better to have something like 
```
GET /pipelines/{id}/jobs/{jobName}/latestBuild?status={boolean}
```

PR #1776, opened 2019-10-07
feat: Enable admin users to update build status failure

sakka2:
## Context
Administrators need to stop jobs which looks like running on the UI but actually failed, and show the reason of failure on the UI for the users.

## Objective
Enable administrator to update build status to `FAILURE`.

## License

### DIFF
--- a/plugins/builds/update.js
+++ b/plugins/builds/update.js
@@ -51,8 +51,17 @@ module.exports = () => ({
 
                     // Users can only mark a running or queued build as aborted
                     if (!isBuild) {
+                        // Check if Screwdriver admin
+                        const adminDetails = request.server.plugins.banners
+                            .screwdriverAdminDetails(username, scmContext);
+
                         // Check desired status
-                        if (desiredStatus !== 'ABORTED') {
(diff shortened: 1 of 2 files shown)


plugins/builds/update.js
```diff
@@ -51,8 +51,17 @@ module.exports = () => ({
...
+
                         // Check desired status
-                        if (desiredStatus !== 'ABORTED') {
+                        if (adminDetails.isAdmin) {
+                            if (desiredStatus !== 'ABORTED' && desiredStatus !== 'FAILURE') {
+                                throw boom.badRequest('Admin can only update builds ' +
+                                    'to ABORTED or FAILURE');
+                            }
+                        } else if (desiredStatus !== 'ABORTED') {
                             throw boom.badRequest('Can only update builds to ABORTED');
```
tk3fftk: How about updating the msg to `User can only update builds to ABORTED` same like admin side msg?

sakka2: I fixed message.

tkyi:
Wondering if we should eventually allow admins to do most of the API operations, like remove templates, etc.

Issue #1723, opened 2019-08-22
Template/Command list screen shows wrong VERSION


yoshwata:
**What happened**:
The `VERSION` number on template's list screen is something wrong.

The `VERSION` indicates `0.0.9` on screen.
![YUS7i5d5e3560456479e30b8c48da](https://user-images.githubusercontent.com/32473622/63491144-6ad4bc80-c4f1-11e9-8b54-afd269c86fa1.png)

However, there is no special tags like `stable` or `latest` on `0.0.9`, and it also not to be newest one (`0.0.14`).
![Screen Shot 2019-08-22 at 15 27 28](https://user-images.githubusercontent.com/32473622/63491163-77591500-c4f1-11e9-9cd4-5743e0039681.jpg)

**What you expected to happen**:
`latest`, `stable` tagged version or simply newest one should appear on list screen.
(I prefer to show `stable` version.)

**How to reproduce it**:
1. Register a template and update 10 time to make the version `0.0.10`.
2. Set tags `stable` and `latest` to `0.0.10`
3. Then, the `VERSION` of list screen is still `0.0.9`.

This will occur on command's list screen too.

yoshwata:
The cause of problem is, like I wrote at https://github.com/screwdriver-cd/screwdriver/pull/1778, the columns is aggregated with `MAX`. 

When we hava a template like below.

| namespace | name | description | version
----|---- |---- |---- |
| foo | bar | 2 | 0.0.10
| foo | bar | 1 | 0.0.9
| foo | bar | 9 | 0.0.8

List screen shows like this about `foo/bar` template (version might be already hidden)

| namespace | name | description | version
----|---- |---- |---- |
| foo | bar | 9 | 0.0.9

The expected behavior is like this.

| namespace | name | description | version
----|---- |---- |---- |
| foo | bar | 2 | 0.0.10

yoshwata:
I propose two solution. I'd like to get your opinion.

#### 1. Paginating

Restricting number of DB queries by paginating. New search queries are emitted by scrolling list screen, so DB performance is not so bad.

#### 2. Implement DB query to get the list of newest templates.

This is a sample for mysql to list newest version of templates. 

```sql
select namespace, name, (select version from templates where namespace=t.namespace and name=t.name order by INET_ATON(SUBSTRING_INDEX(CONCAT(version, '.0'),'.',4)) desc limit 1) version from templates t group by namespace, name;
```

[...]

PR #1778, opened 2019-10-07
fix(1723): Get newest templates or commands on list screen

yoshwata:
## Context

The template or command's list screen does not show consistent information.

When we hava a template like below.

| namespace | name | description | version
----|---- |---- |---- |
| foo | bar | 2 | 0.0.10
| foo | bar | 1 | 0.0.9
| foo | bar | 9 | 0.0.8

List screen shows like this about `foo/bar` template (version might be already hidden)

| namespace | name | description | version
----|---- |---- |---- |
| foo | bar | 9 | 0.0.9

That is because the SQL query select with `MAX` about all columns of template/command table.
https://github.com/screwdriver-cd/datastore-sequelize/blob/8ad1c7f280e09569e20fc31be6e52883451b65b6/index.js#L491

I found that when `group by` is included in config of list, `MAX` is not used as the query. 
https://github.com/screwdriver-cd/datastore-sequelize/blob/8ad1c7f280e09569e20fc31be6e52883451b65b6/index.js#L472

This PR is fix the list api of command/template not to use `MAX` to list them.

## Objective

Get newest templates/command by using `getTemplate` or `getCommand` functions which does not emit `MAX` queries.

## References

https://github.com/screwdriver-cd/screwdriver/issues/1723

## License


yoshwata:
@jithin1987 thanks. I close this PR. I will discuss about db query at https://github.com/screwdriver-cd/screwdriver/issues/1723 continuingly.

What I'd like to fix was to showing newest information about any columns in template/command list screen, so hiding  `version` column does not solve the problem, like I wrote at https://github.com/screwdriver-cd/screwdriver/pull/1778#issue-325138643.

PR #1779, opened 2019-10-07
fix: post build api creating build twice

klu909:
## Context

previous commit breaks an functional test where POST /build would actually create build twice

## Objective

add a skipMessage when we create event so it would skip build creation
 
## References

https://github.com/screwdriver-cd/screwdriver/pull/1773

## License

### DIFF
--- a/plugins/builds/create.js
+++ b/plugins/builds/create.js
@@ -112,7 +112,8 @@ module.exports = () => ({
                                         type,
                                         username,
                                         scmContext,
-                                        sha
+                                        sha,
+                                        skipMessage: 'skip build creation'
                                     });
                                 })
                                 .then((event) => {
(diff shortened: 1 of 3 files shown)


plugins/builds/create.js
```diff
@@ -112,7 +112,8 @@ module.exports = () => ({
                                         type,
                                         username,
                                         scmContext,
-                                        sha
+                                        sha,
+                                        skipMessage: 'skip build creation'
```
jithin1987: why can't we just remove lines 119 - 130. ie don't create a build here let eventFactory take care if it. ?

klu909: Because following logic needs a build and eventFactory return an event, thus we would need to do a build look up. There might be a cleaner way but I am looking to fix this ASAP to unblock other people

PR #1784, opened 2019-10-10
fix: Allow pipeline metrics to be fetched by pagination

Code-Beast:
## Context
Currently on the frontend the dashboard/show page fetches metrics of last month, which may retrieve nothing for some pipeline that hasn't been triggered for a while and cause pipeline-card component or collection-table-row component to show nothing (because lastEventInfo will be empty too). If we allow pipeline metrics to be fetched by pagination, we can always get the latest metrics info as long as the pipeline has been triggered before.

## Objective
Pipeline metrics can be fetched by pagination

## License


Code-Beast:
Need some test changes

plugins/pipelines/metrics.js
```diff
@@ -25,23 +26,29 @@ module.exports = () => ({
             const factory = request.server.app.pipelineFactory;
             const { id } = request.params;
             const { aggregateInterval } = request.query;
+            const { page, count } = request.query;
```
tkyi: can combine with above line

plugins/pipelines/metrics.js
```diff
@@ -24,24 +25,33 @@ module.exports = () => ({
...
-                        throw boom.badRequest(`Time range is longer than ${MAX_DAYS} days`);
-                    }
+                    let config;
 
-                    const config = { startTime, endTime };
+                    // Only when either page or count is unavailable
+                    // check whether startTime and endTime are valid
+                    if (!page || !count) {
+                        if (!startTime || !endTime) {
+                            ({ startTime, endTime } =
```
adong: This is a neat way of writing destruction from function return values.

plugins/pipelines/metrics.js
```diff
@@ -24,24 +25,33 @@ module.exports = () => ({
...
 
-                    if (!validTimeRange(startTime, endTime, MAX_DAYS)) {
-                        throw boom.badRequest(`Time range is longer than ${MAX_DAYS} days`);
-                    }
+                    let config;
 
-                    const config = { startTime, endTime };
+                    // Only when either page or count is unavailable
+                    // check whether startTime and endTime are valid
+                    if (!page || !count) {
```
d2lam: Shouldn't this be if (!page && !count) (https://github.com/screwdriver-cd/models/blob/master/lib/baseFactory.js#L172-L178)
Default count is 50, so if only page is set, it will use count = 50
Default page = 1, so if only count is set, it will use page = 1

If BOTH are not available then it will use startTime & endTime

d2lam:
Don't you also need to make changes to the models so that it will take in the page/count: https://github.com/screwdriver-cd/models/blob/master/lib/pipeline.js#L1287-L1290

Code-Beast:
> Don't you also need to make changes to the models so that it will take in the page/count: https://github.com/screwdriver-cd/models/blob/master/lib/pipeline.js#L1287-L1290

I made changes in this pull request https://github.com/screwdriver-cd/models/pull/399

plugins/pipelines/metrics.js
Code-Beast: I think we only allow fetching by pagination when both page and count are available in the config?

jithin1987: well model implementation needs both https://github.com/screwdriver-cd/models/pull/399/files#diff-679e90b5b15ab1b73ddb7ed2377e7413R1295
unless that's changed this change here is good.

plugins/pipelines/metrics.js
```diff
@@ -24,24 +25,33 @@ module.exports = () => ({
...
+                                setDefaultTimeRange(startTime, endTime, MAX_DAYS));
+                        }
+
+                        if (!validTimeRange(startTime, endTime, MAX_DAYS)) {
+                            throw boom.badRequest(`Time range is longer than ${MAX_DAYS} days`);
+                        }
+
+                        config = { startTime, endTime };
+                    } else {
+                        config = { page, count };
```
tkyi: could set config = { page, count } on line 37 where you declare it then you can remove this `else` statement

plugins/pipelines/metrics.js
d2lam: Why do we suddenly require to have BOTH fields? It has always been you only need `page` or `count` (https://github.com/screwdriver-cd/models/blob/master/lib/baseFactory.js#L172-L179). The other will use default value. Why should the behavior be different here?

jithin1987: well now we need to start changing from model. This should have been caught earlier in review.

jithin1987: @Code-Beast could you please handle this scenario

d2lam: Sorry did not review the previous PR. We can merge this one and fix afterward. It's not too critical, but we should fix it so that the behavior is consistent.

Code-Beast: sure

plugins/pipelines/metrics.js
Code-Beast: okay

jithin1987:
@Code-Beast tests are failing.But doesn't seem like the ones you changed.

Code-Beast:
> @Code-Beast tests are failing.But doesn't seem like the ones you changed.

I see that. Seems like they have existed for some time

PR #1786, opened 2019-10-11
feat: [2] Add tag/release triggered test

kkisic:
## Context
recently tag/release triggered feature has released, but there are no functional tests.

## Objective
- add functional tests about tag and release
- fix bugs about the argument of `searchForBuild`. It requires `desiredSha`, but some functions passed `sha`

## References
https://octokit.github.io/rest.js
https://github.com/screwdriver-cd-test/functional-git/pull/6740

## License


features/step_definitions/git-flow.js
```diff
@@ -80,6 +81,21 @@ defineSupportCode(({ Before, Given, When, Then }) => {
             });
     });
 
+    Given(/^a pipeline with all stopped builds$/, {
+        timeout: TIMEOUT
+    }, function step() {
```
tk3fftk: Do we need a return with Promise Object for this `Given` statement ?

features/git-flow.feature
```diff
@@ -35,4 +36,20 @@ Feature: Git Flow
...
         And it is against the pipeline's branch
-        Then a new build from `main` should be created to test that change
+        Then a new build from "main" should be created to test that change
+
+    Scenario: New Tag
+        When a tag is created
+        Then a new build from "tag-triggered" should be created to test that change
+
+    Scenario: New Annotated Tag
+        When a annotated tag is created
```
tk3fftk: a annotated tag -> an annotated tag

features/step_definitions/git-flow.js
```diff
@@ -143,16 +159,77 @@ defineSupportCode(({ Before, Given, When, Then }) => {
...
         timeout: TIMEOUT
     }, function step() {
+        const branch = this.branch;
+        const tag = this.tag;
+
+        return github.createBranch(branch, this.repoOrg, this.repoName)
+            .then(() => github.createFile(branch, this.repoOrg, this.repoName))
+            .then(({ data }) => {
+                this.sha = data.commit.sha;
+                github.createTag(tag, branch, this.repoOrg, this.repoName);
```
tk3fftk: need `return` for returning Promise Object to cucumber?

features/step_definitions/git-flow.js
```diff
@@ -143,16 +159,77 @@ defineSupportCode(({ Before, Given, When, Then }) => {
...
+        timeout: TIMEOUT
+    }, function step() {
+        const branch = this.branch;
+        const tag = this.tag;
+
+        return github.createBranch(branch, this.repoOrg, this.repoName)
+            .then(() => github.createFile(branch, this.repoOrg, this.repoName))
+            .then(({ data }) => {
+                this.sha = data.commit.sha;
+                github.createAnnotatedTag(tag, branch, this.repoOrg, this.repoName);
```
tk3fftk: need `return`?

features/step_definitions/git-flow.js
```diff
@@ -143,16 +159,77 @@ defineSupportCode(({ Before, Given, When, Then }) => {
...
+            .then(() => github.createFile(branch, this.repoOrg, this.repoName))
+            .then(({ data }) => {
+                this.sha = data.commit.sha;
+
+                return github.createAnnotatedTag(tag, branch, this.repoOrg, this.repoName);
+            })
+            .then(() => github.createRelease(tag, this.repoOrg, this.repoName));
+    });
+
+    Then(/^a new build from "([^"]*)" should be created to test that change$/, {
```
sugarnaoming: This Regular expression is matched with `""`.
How about below?
`"([^"]*)"` => `"([^"]+)"`

features/support/github.js
```diff
@@ -142,6 +150,107 @@ function createPullRequest(branch, repoOwner, repoName) {
...
+
+            return octokit.git.createRef({
+                owner,
+                repo,
+                ref: `refs/tags/${tag}`,
+                sha
+            });
+        })
+        .catch((err) => {
+            Assert.strictEqual(err.status, 422);
```
tk3fftk: [ask] Is this error code same with L107?
> // throws an error if a branch already exists, so this is fine

features/step_definitions/git-flow.js
kkisic: Thanks. I fixed it.

tk3fftk:
Do we need to run functional test in PR when codes are changed?

features/support/github.js
```diff
@@ -36,17 +36,25 @@ function randomString(stringLength) {
  * @param  {String}          repoOwner  Owner of the repository
  * @param  {String}          repoName   Name of the repository
  * @param  {String}          branch     Name of the branch to delete
+ * @param  {String}          tag        Name of the tag to delete
  * @return {Promise}
  */
-function cleanUpRepository(branch, repoOwner, repoName) {
+async function cleanUpRepository(branch, tag, repoOwner, repoName) {
```
tk3fftk: This function seems to not return any `Promise`. Is it ok?

features/git-flow.feature
kkisic: Thanks. I fixed it.

features/support/github.js
```diff
@@ -142,6 +150,107 @@ function createPullRequest(branch, repoOwner, repoName) {
...
+                message: 'this is annotated tag',
+                object: sha,
+                type: 'commit',
+                tagger: {
+                    name: 'test',
+                    email: 'test@example.com',
+                    date: '2019-10-09T15:00:00+09:00'
+                }
+            })
+                .then(response => octokit.git.createRef({
```
tk3fftk: This `then()` should be outside of this function.

features/support/github.js
```diff
@@ -142,6 +150,107 @@ function createPullRequest(branch, repoOwner, repoName) {
...
+function createAnnotatedTag(tag, branch, repoOwner, repoName) {
+    const owner = repoOwner || 'screwdriver-cd-test';
+    const repo = repoName || 'functional-git';
+
+    return octokit.git.getRef({
+        owner,
+        repo,
+        ref: `heads/${branch}`
+    })
+        .then((referenceData) => {
```
tk3fftk: These indents look like a strange:thinking:

kkisic: Because of this [PR](https://github.com/screwdriver-cd/screwdriver/pull/683/files)

features/support/github.js
```diff
@@ -142,6 +150,107 @@ function createPullRequest(branch, repoOwner, repoName) {
...
+            return octokit.git.createRef({
+                owner,
+                repo,
+                ref: `refs/tags/${tag}`,
+                sha
+            });
+        })
+        .catch((err) => {
+            Assert.strictEqual(err.status, 422);
+        });
```
sugarnaoming: I think this indent is too deep.
If this indent is an ESLint rule, there is no problem.

features/support/github.js
tk3fftk: oh...

features/support/github.js
kkisic: Why do you think this `then()` should be outside of this function ?
I think this `then()` can be in this function because `createTag()` makes a tag object only and it is difficult to understand for us.

features/support/github.js
kkisic: I had a misunderstanding, and I need to fail tests.

features/step_definitions/git-flow.js
kkisic: That's true. I fixed it.

features/step_definitions/git-flow.js
kkisic: You're right. We need to return Promise Object.

features/step_definitions/git-flow.js
kkisic: Same as above. I fixed it.

features/support/github.js
kkisic: Same as above. I fixed it.

kkisic:
> Do we need to run functional test in PR when codes are changed?
> CC @jithin1987

I think even if functional test is failed in PR when code are changed, we don't know whether functional test or production code is the cause.
In addition, secret is visible to anyone, because it requires `allowInPR` to be true.

features/support/github.js
tk3fftk: My intention is like following:
```
return octokit.git.getRef({
    (snip)
}).then((referenceData) => {
    const sha = referenceData.data.object.sha;

    return octokit.git.createTag({
                (snip)
    })
}).then(response => octokit.git.createRef({
    ...
    }));
}).catch(() => {...});
```

kkisic: I had a misunderstanding. I fixed it.

features/support/github.js
```diff
@@ -36,17 +36,27 @@ function randomString(stringLength) {
...
+    const tagParams = {
+        owner: repoOwner,
+        repo: repoName,
+        ref: `tags/${tag}`
+    };
 
     return octokit.git.getRef(branchParams)
-        .then(() => octokit.git.deleteRef(branchParams), () => {});
+        .then(() => octokit.git.deleteRef(branchParams))
+        .catch(err => Assert.strictEqual(404, err.status))
```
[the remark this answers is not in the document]
kkisic: I think that `.catch()` is necessary, because if the reference of `branchParams` doesn't exist, the reference of `tagParams` is not deleted.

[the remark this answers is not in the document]
tk3fftk: I might get it. Are these descriptions correct?
- First `deleteRef()` is possible to reject when `branchParams` doesn't exist then continue the chain.
-  `getRef()` and second `deleteRef()` are possoble to reject when `tagParams` doesn't exist then return resolved Promise object.
- this function returns only resolved Promise object so `catch()` of `features/step_definitions/git-flow.js` side looks unnecessary

Ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch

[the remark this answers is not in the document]
tk3fftk: L59 throws `reject`, so `catch()` of `features/step_definitions/git-flow.js` side is necessary.

features/support/sdapi.js
```diff
@@ -236,8 +236,49 @@ function cleanupToken(config) {
...
+
+    return findBuilds({
+        instance,
+        pipelineId,
+        jobName,
+        jwt
+    }).then((buildData) => {
+        const result = buildData.body || [];
+
+        result.filter(item => desiredStatus.includes(item.status)).forEach(build =>
```
d2lam: will this wait for all request promises to finish?

kkisic: I fixed it.

tk3fftk:
@d2lam would you review again?

features/git-flow.feature
```diff
@@ -35,4 +36,20 @@ Feature: Git Flow
...
+
+    Scenario: New Annotated Tag
+        When an annotated tag is created
+        Then a new build from "tag-triggered" should be created to test that change
+
+    Scenario: New Release
+        When a release is created
+        Then a new build from "release-triggered" should be created to test that change
+
+    Scenario: New Annotated Release
```
catto: How about `New Release with Annotated Tag`? There is no "annotated release" on GitHub

kkisic: I think so. I completely thought "release with annotated tag" was "annotated release".

PR #1800, opened 2019-10-16
[skip ci] fix: trigger on data schema

jithin1987:
## License


d2lam:
Actually do we want to do this? we might accidentally pull in something that is not ready. For example, it might need to be in this order: schema -> models -> API

Issue #1610, opened 2019-05-01
Periodic build implementation decouple from API


d2lam:
Currently, periodic build logic is in [executor-queue](https://github.com/screwdriver-cd/executor-queue/blob/master/index.js#L269-L280), which is bundled with the API. However, if the API crashes during the same time periodic build is scheduled to run, it might not add itself to the delayed queue: https://github.com/screwdriver-cd/executor-queue/blob/master/index.js#L320, and periodic build will stop being scheduled.

We should decouple the periodic logic from the API so that API crashing will not affect it.

pritamstyz4ever:
Currently when the node process receives a termination request it is abruptly closed and thus in-process messages are lost. The following PR's will add a graceful shutdown logic which will register tasks to clean up queue related tasks.
https://github.com/screwdriver-cd/models/pull/407
https://github.com/screwdriver-cd/executor-queue/pull/75
https://github.com/screwdriver-cd/executor-base/pull/61
https://github.com/screwdriver-cd/screwdriver/pull/1827

PR #1827, opened 2019-11-04
Feature/periodic build

pritamstyz4ever:
## Context

Currently when the node process terminates it happens abruptly and the in process messages are lost

## Objective

This PR adds a graceful shutdown plugin which can handle any task registered to be executed during termination

## References

https://github.com/screwdriver-cd/screwdriver/issues/1610

## License

### DIFF
--- a/lib/registerPlugins.js
+++ b/lib/registerPlugins.js
@@ -20,7 +20,7 @@ function registerDefaultPlugins(server, callback) {
         '../plugins/swagger',
         '../plugins/validator',
         '../plugins/template-validator',
-        '../plugins/command-validator'
+        '../plugins/command-validator',
     ].map(plugin => require(plugin));
 
     async.eachSeries(plugins, (plugin, next) => {
@@ -29,6 +29,15 @@ function registerDefaultPlugins(server, callback) {
                 prefix: '/v4'
             }
         }, next);
+    }, callback);
+}
+
+function registerCustomPlugin(server, callback) {
+    async.eachSeries(['shutdown'], (pluginName, next) => {
+        server.register({
+            register: require(`../plugins/${pluginName}`),
+            options: { terminationGracePeriod: process.env.TERMINATION_GRACE_PERIOD || 30 }
+        }, {}, next);
     }, callback);
 }
 
@@ -94,7 +103,8 @@ module.exports = (server, config) => (
     new Promise((resolve, reject) => {
         async.series([
             async.apply(registerDefaultPlugins, server),
-            async.apply(registerResourcePlugins, server, config)
+            async.apply(registerResourcePlugins, server, config),
+            async.apply(registerCustomPlugin, server)
         ], (err) => {
             if (err) {
                 return reject(err);
--- a/lib/registerPlugins.js
+++ b/lib/registerPlugins.js
@@ -20,7 +20,7 @@ function registerDefaultPlugins(server, callback) {
         '../plugins/swagger',
         '../plugins/validator',
         '../plugins/template-validator',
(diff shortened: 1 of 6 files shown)


lib/server.js
```diff
@@ -144,8 +144,17 @@ module.exports = (config) => {
                     .generateProfile(username, scmContext, ['user'], metadata)
                 );
             server.app.jobFactory.executor.userTokenGen = server.app.jobFactory.tokenGen;
+            server.plugins.shutdown.handler({
+                taskname: 'executor-queue-cleanup',
+                task: async () => {
+                    await server.app.jobFactory.cleanUp();
+                    server.log(['shutdown'], 'completed clean up tasks', new Date().toISOString());
```
DekusDenial: prefer to use `winston` for logging

lib/server.js
```diff
@@ -144,8 +144,17 @@ module.exports = (config) => {
...
+                }
+            });
 
             // Start the server
-            return server.start().then(() => server);
+            return server.start()
+                .then(() => server)
+                .catch(err => console.log('Failed to start server', err));
```
DekusDenial: same as above

lib/registerPlugins.js
```diff
@@ -32,6 +32,20 @@ function registerDefaultPlugins(server, callback) {
     }, callback);
 }
 
+/**
+ *
```
DekusDenial: consider adding little bit more stuff for jsdoc

plugins/shutdown.js
```diff
@@ -0,0 +1,99 @@
...
+    return '';
+}
+
+/**
+ *
+ * @param {function} fn
+ * @param {string} timeout
+ */
+function promiseTimeout(fn, timeout) {
+    return new Promise(((resolve, reject) => {
```
DekusDenial: not sure which is more elegant, this or using `Promise.race`?

plugins/shutdown.js
```diff
@@ -0,0 +1,99 @@
...
+ * Hapi interface for plugin to handle serve graceful shutdown
+ * @method register
+ * @param  {Hapi.Server}    server
+ * @param  {Object}         options
+ */
+exports.register = (server, options, next) => {
+    const taskHandler = async (resolve, reject) => {
+        try {
+            await Promise.all(Object.keys(tasks).map(async (key) => {
+                server.log(['shutdown'], `executing task ${key}`, new Date().toISOString());
```
DekusDenial: same as above

plugins/shutdown.js
```diff
@@ -0,0 +1,99 @@
...
+            resolve();
+        } catch (err) {
+            console.log(err);
+            reject(err);
+        }
+    };
+    const gracefulStop = async () => {
+        try {
+            server.log(['shutdown'], 'gracefully shutting down server', new Date().toISOString());
+            await server.root.stop({
```
DekusDenial: what's the difference btw this and `server.stop()`?

plugins/shutdown.js
```diff
@@ -24,19 +33,21 @@ function register(task) {
...
-        // Set up the timeout
-        setTimeout(() => {
-            resolve(`Promise timed out after ${timeout} ms`);
-        }, timeout);
-    }));
+    return Promise.race([
+        Promise.resolve(fn),
+        new Promise((resolve) => {
+            const id = setTimeout(() => {
+                clearTimeout(id);
```
DekusDenial: It seems unnecessary to call clear timeout

pritamstyz4ever: Yeah, must be a left over

PR #1836, opened 2019-11-13
Feature/periodic build

pritamstyz4ever:
Add parseInt to env variable


pritamstyz4ever:
Created a separate one

PR #1837, opened 2019-11-13
feat(1610): Add parseInt to terminationgraceperiod

pritamstyz4ever:
env var

## Context
The env var from k8s is type string for terminationGracePeriodSceonds

## Objective

This adds a parseInt to the env var

## License

### DIFF
--- a/lib/registerPlugins.js
+++ b/lib/registerPlugins.js
@@ -42,7 +42,7 @@ function registerCustomPlugin(server, callback) {
         server.register({
             register: require(`../plugins/${pluginName}`),
             options: {
-                terminationGracePeriod: parseInt(process.env.TERMINATION_GRACE_PERIOD, 2)
+                terminationGracePeriod: parseInt(process.env.TERMINATION_GRACE_PERIOD, 10)
                     || 30
             }
         }, {}, next);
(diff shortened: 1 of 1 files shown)


lib/registerPlugins.js
```diff
@@ -41,7 +41,10 @@ function registerCustomPlugin(server, callback) {
     async.eachSeries(['shutdown'], (pluginName, next) => {
         server.register({
             register: require(`../plugins/${pluginName}`),
-            options: { terminationGracePeriod: process.env.TERMINATION_GRACE_PERIOD || 30 }
+            options: {
+                terminationGracePeriod: parseInt(process.env.TERMINATION_GRACE_PERIOD, 2)
```
jithine: radix should be 10. Alternatively you can just do `+process.env.TERMINATION_GRACE_PERIOD`

pritamstyz4ever: fixed

parthasl: why not just Number?

jithine: both provide same outcome. https://stackoverflow.com/questions/4090518/what-is-the-difference-between-parseint-and-number

`+` is same as Number()

tk3fftk:
https://github.com/screwdriver-cd/screwdriver/issues/1610

Issue #1844, opened 2019-11-16
some of api are missing validation on parameters


klu909:
**What happened**:
There's inconsistency on our api, where some of them are missing validation on parameter.
For example on /pipelines/{id} there's validation on id but not on /pipelines/{id}/triggers.
This leads to inconsistent behavior when user incorrectly passing invalid parameter,
/pipelines/randomString would results in code 400 but /pipelines/string/triggers results in
code 500. 
**What you expected to happen**:
Add validation on parameter to api that doesn't have it.
**How to reproduce it**:

PR #1845, opened 2019-11-16
fix(1844): enforce validation on parameters

klu909:
## Context

There's inconsistency on our api, where some of them are missing validation on parameter.
For example on /pipelines/{id} there's validation on id but not on /pipelines/{id}/triggers.
This leads to inconsistent behavior when user incorrectly passing invalid parameter,
/pipelines/randomString would results in code 400 but /pipelines/string/triggers results in
code 500.

## Objective

enforce validation on parameters on our api

## References

https://github.com/screwdriver-cd/screwdriver/issues/1844

## License


adong:
Needs to be rebased with latest master. @klu909

PR #1842, opened 2019-11-15
fix: use cucumber 6

tk3fftk:
## Context
Current cucumber.js version is very old (2!) and some feature tests are flakey.

## Objective
- Use the latest cucumber.js 6 to use `--retry` option
- Fix deprecation of cucumber,js 

## References
- [cucumber-js/cli.md at master · cucumber/cucumber-js](https://github.com/cucumber/cucumber-js/blob/9518f1c46de28767d4e19f2faf5c3b5c29936c97/docs/cli.md#parallel-experimental)
- [cucumber-js/CHANGELOG.md at master · cucumber/cucumber-js](https://github.com/cucumber/cucumber-js/blob/master/CHANGELOG.md)

## License
I confirm that this contribution is made under a BSD license and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #1852, opened 2019-11-21
fix: refix some step definitions

tk3fftk:
## Context
https://github.com/screwdriver-cd/screwdriver/pull/1842 had some unexpected changes for feature tests :disappointed_relieved:
it may cause the test crash.

## Objective
Restore these changes.
Let me know if other changes remain.

## License


jithine:
@tk3fftk with cucumber6 are tests hanging on re-try ? For example https://cd.screwdriver.cd/pipelines/1/builds/160026/steps/test suggests it re-tried and passed. However build just hang.

tk3fftk:
@jithine 
hmm, we may need `--exit` option. I'll add it.
ref: https://github.com/cucumber/cucumber-js/blob/master/docs/cli.md#exiting

Issue #1314, opened 2018-09-27
Local build feature


catto:
There are different of build environments between Screwdriver.cd and users' such as environment variables, templates and commands.
It would be great if users can run their build instantly on their local Mac with a mostly the same environment as Screwdriver.cd's.

For example,
1. A user starts a temporal build via CLI
1. CLI parses `screwdriver.yaml` and run docker container on user's development environment
    1. Setup meta, sd-cmd inside container which can talk to Screwdriver.

tk3fftk:
It will possibly be able to implement in [tk3fftk/sdctl: Screwdriver.cd API wrapper CLI](https://github.com/tk3fftk/sdctl) like `$ sdctl build local`

tk3fftk:
Concourse has `one-off build` feature. It runs a task which is executed inside a container in Concourse cluster and logs are sent to local environment. And the `one-off build` only can be shown by direct specification of a build iD from UI.

Pros of this `one-off build` feature are that we can check cluster specific problems like network issue via cli.
Cons of the feature are that it may be slower than a local build.

References:
- [Concourse CI](https://concourse-ci.org/) (`Rapid Local Iteration`)
- [Running Tasks - Concourse CI](https://concourse-ci.org/running-tasks.html#fly-execute)

yuichi10:
## Background

The build environment on screwdriver has difference from a user's local environment.
Such as

- Can use sd-step command
- Can use template in screwdriver.yaml
- Can use metadata
- Guarantee build reproducibillity

## Objective

- Able to check a result of builds quickly and easily
- Make it easy to troubleshoot

## Use cases

- Check the build works correctly
  1. Do steps work as users thought?
  2. Are tests passed before make a Pull-Request?
  3. Do scripts that includes sd-cmd work as users thought?
- Troubleshoot
  1. The builds work correctly in local environment but do not work on screwdriver
  2. Check the behavior which differ depends on version(such as templates, sd-cmd)
  3. Investigate user builds by screwdriver admins
  
## Design

- A user set user token to sdlocal command
- A user can specify a job defined in screwdriver.yaml and start that job
  - A user can configure meta and secrets
- The build works on the environment which is specified in screwdriver.yaml
- The container has `/opt/sd` directory copied from launcher image
- A user can obtain logs and artifacts of the build in local

## Example

```
$ ls
screwdriver.yaml
$ cat screwdriver.yaml
jobs:
  test:
    image: centos7:latest
    steps:
[...]

PR #1853, opened 2019-11-21
design: New sd-local Proposal

kkisic:
## Context
The build environment on screwdriver has difference from a user's local environment.
Such as

- Can use sd-step command
- Can use template in screwdriver.yaml
- Can use metadata
- Guarantee build reproducibillity

## Objective
- Able to check a result of builds quickly and easily
- Make it easy to troubleshoot

## References
https://github.com/screwdriver-cd/screwdriver/issues/1314#issuecomment-551523261

## License


design/sd-local.md
```diff
@@ -0,0 +1,164 @@
...
+   - Artifacts Directory
+2. Run Local Mode Launcher on Build Container
+   - Run Launcher binary with `--local-mode` option in `run.sh`
+   - Config Build Environment Variables and Environment Variables by env option 
+   - Run Steps got in `Initialization` No.3
+   - Run `sd-cmd`
+     1. Get the binaries from Store
+     2. Run it
+3. Write Artifacts to Artifacts Directory
+4. Output logs to stdout
```
jithine: can logs also be written to a path under Artifacts directory ?

design/sd-local.md
```diff
@@ -0,0 +1,164 @@
...
+- `--local` Run command with `.sdlocal/config` file in current directory.
+
+## Design considerations
+
+#### Implement Local Mode Launcher.
+Need to be implement the following:
+- Add `--local-mode` option to run Launcher binary on Local Mode
+- Not to call Screwdriver API on Local Mode
+- Use the steps information from the response of validator API in `Initialization`
+- Output the build logs to stdout instead of `log-service`
```
jithine: one option is to support a `local-mode` in `log-service` and have it write to `console`. Also we can leverage log-service to write step logs under a directory in `artifacts` folder, then `sdlocal` can simply `tail` this file.

design/sd-local.md
kkisic: We can output logs under Artifacts directory.
Would you tell me why?

design/sd-local.md
kkisic: Is it easier to use `log-service`  to write step logs under a directory in `artifacts` folder?
Do you think you don't need to output logs to stdout?

s-yoshika: It's sounds good. It allows sdlocal to decorate logs. For example, it can put the step name between steps.

design/sd-local.md
jithine: console logs might be too long to analyze later. if we persist logs automatically under Artifacts dir, then user can refer back it later, without having to worry about scrolling limitations of terminal.

sakka2: I understand. I think it is better too.

design/sd-local.md
sakka2: Fixed to use `log-service` and output logs to Artifacts Directory.

MysticDoll:
@jithine Please review again. 
`sdlocal` can tail the artifacts directory, so we delete following one-line.
https://github.com/screwdriver-cd/screwdriver/pull/1853/commits/6adb5dff591059e135c08fdcc714d81bc8cc074b#diff-49f99a06253f0a9a292e89f6a469b9e4L166

MysticDoll:
If this PR has no problem, please merge this.

Issue #1339, opened 2018-10-14
Start event with meta data - parameterized builds


jithine:
## Objective

API has the ability to start an event with custom meta data. This can be exposed in UI to support parametrized builds. 

## Objective

1. Clicking `Start` button should have a UI element for user to provide meta data. (key value pairs)
1. Clicking `Start Pipeline from here` on pipeline event graph should have a UI element to provide meta data (key value pairs) 
1. Build to expose the data under key `event-inputs` (perhaps another better name)

## References

https://api.screwdriver.cd/v4/documentation#!/v4/postV4Events

tk3fftk:
This feature would be very useful in some case.
But we concern that the feature may decrease build reproducibility and make hard to track trouble of builds.
How about adding a cluster wide feature flag to toggle this feature enabled/disabled by admins?

jithin1987:
>  But we concern that the feature may decrease build reproducibility and make hard to track trouble of builds.

Since all the information is persisted in event metadata, I believe build reproducibility is guaranteed. If user want to re-run the build with same parameters, then user can follow same steps user does today - select the event and re-run the build. 

What we need to do from a UX perspective it to highlight parameters in a better way than just pass it to event API. For example, `sd-init-step` can log all parameters. Event graph can display them to better highlight the fact that this event was started with parameters. 

I am open to hiding this behind a feature flag. cc @adong

jithin1987:
Current proposal. `parameters` will be a top level config for pipeline.

```
parameters:
   auth_strategy: simple
   user:
     value: jitthin
     description: User running build
jobs:
  build:
...
```

Only pipelines which have `parameters` config in their `screwdriver.yaml` will see option to enter parameters when triggering Start button and will see only the parameters which are in the `screwdriver.yaml`.

`parameters` will be made available in the build via meta data using key  `parameters`. Events API already supports passing in meta data, so meta data will be a good vehicle to pass in the inputs. 

Now if we have pipelines triggering each other, we need a way to triggering pipeline to pass information if the triggered one is parameterized.

PR #1873, opened 2019-12-06
fix(1339): event with parameters and parent event

klu909:
## Context

In my previous fix for restarting event, I made all event inherit parent event's parameters. This is causing issues where user will not be able to set custom parameters for event with parent event.

## Objective

Event won't inherit parent event parameters if custom parameters are supplied

## References

https://github.com/screwdriver-cd/screwdriver/issues/1339

## License

### DIFF
--- a/plugins/events/create.js
+++ b/plugins/events/create.js
@@ -197,7 +197,8 @@ module.exports = () => ({
                                     .then((parentEvent) => {
                                         payload.baseBranch = parentEvent.baseBranch || null;
 
-                                        if (parentEvent.meta && parentEvent.meta.parameters) {
+                                        if ((!payload.meta || !payload.meta.parameters) &&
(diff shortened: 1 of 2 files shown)


plugins/events/create.js
```diff
@@ -197,7 +197,8 @@ module.exports = () => ({
                                     .then((parentEvent) => {
                                         payload.baseBranch = parentEvent.baseBranch || null;
 
-                                        if (parentEvent.meta && parentEvent.meta.parameters) {
+                                        if ((!payload.meta || !payload.meta.parameters) &&
+                                            parentEvent.meta && parentEvent.meta.parameters) {
```
jithine: on restart scenarios does `payload.meta.parameters` include entire parameters ? Do we need to do a `merge` ?

cc @adong

klu909: yeah it includes entire parameters as long as there's change in default parameters.

Issue #1839, opened 2019-11-14
build timeout is not reliable


jithine:
**What happened**:
Build time out which is enforced via [launcher](https://github.com/screwdriver-cd/launcher/blob/master/executor/executor.go#L255) doesn't work on scenarios when build container crashes. This is because it relies on `launcher` process to be running to enforce the time out.

**What you expected to happen**:
SD API should have safeguard against launcher crashes to ensure that build always times out.

**How to reproduce it**:

1. Start a build.
2. Delete build container.
3. Build runs for ever in Screwdriver UI until user manually Stops it.

jithine:
One option we have is to leverage redis-resque for this.

1. API get request to update build status to `RUNNING` from `launcher`.
1. API drops a message in redis queue using [enqueAt](https://github.com/screwdriver-cd/executor-queue/blob/master/index.js#L333-L334) `buildTimeout + delta`
1. Use queue processing to abort build if `launcher` has not already aborted it.

kumada626:
I concern that builds timeout before the real build time is over timeout.
I think that `enqueueTime` is set when the build is added to executor(e.g. k8s, jenkins) from queue.
However, the build is sometimes queued in the executor and not start immediately. For example, the Jenkins itself has own job queue. In such case, the build is killed before the real build time is over timeout.

pritamstyz4ever:
@kumada626 .Thank you for bringing this up, we have addressed this in new PR's by removing enqueue time and adding the build start time when the build status changes to RUNNING. This would be read in the worker and used for timeout considerations.

PR #1908, opened 2020-01-13
fix(1839): Add endpoint to get active build step

pritamstyz4ever:
## Context

To update active build step after timeout we need a new api

## Objective

This PR will add a new API GET /builds/{id}/steps/active for the current running step which will be stopped after build timeout. To update step the from queue-worker we need to allow temporal token.

## References
https://github.com/screwdriver-cd/screwdriver/issues/1839
Other PR's
https://github.com/screwdriver-cd/executor-base/pull/63
https://github.com/screwdriver-cd/executor-queue/pull/79
https://github.com/screwdriver-cd/queue-worker/pull/151
https://github.com/screwdriver-cd/models/pull/425

## License


plugins/builds/steps/active.js
```diff
@@ -0,0 +1,55 @@
+'use strict';
+
+const boom = require('boom');
+const schema = require('screwdriver-data-schema');
+const getSchema = schema.models.build.getStep;
+
+module.exports = () => ({
```
jithine: missing jsdoc. Refer to other plugin code for example.

plugins/builds/steps/active.js
```diff
@@ -0,0 +1,55 @@
...
+                && request.auth.credentials.username.toString();
+            const buildId = request.params.id && request.params.id.toString();
+
+            if (request.auth.credentials.scope.includes('temporal') && buildId !== buildIdCred) {
+                return reply(boom.forbidden(`Credential only valid for ${buildIdCred}`));
+            }
+
+            return buildFactory.get(buildId)
+                .then((buildModel) => {
+                    console.log(buildModel);
```
jithine: remove console logs

pritamstyz4ever: done

plugins/builds/steps/active.js
```diff
@@ -0,0 +1,53 @@
...
+            }
+        },
+        handler: (request, reply) => {
+            const buildFactory = request.server.app.buildFactory;
+            const buildIdCred = request.auth.credentials.username
+                && request.auth.credentials.username.toString();
+            const buildId = request.params.id && request.params.id.toString();
+
+            if (request.auth.credentials.scope.includes('temporal') && buildId !== buildIdCred) {
+                return reply(boom.forbidden(`Credential only valid for ${buildIdCred}`));
```
tkyi:
```suggestion
                return reply(boom.forbidden(`Credential only valid for build ${buildIdCred}`));
```

plugins/builds/README.md
```diff
@@ -40,6 +40,9 @@ Arguments:
 #### Gets a build step
 `GET /builds/{id}/steps/{name}`
 
+#### Gets the current active build step
+`GET /builds/{id}/steps/active`
```
tkyi: what if a step is named `active`?

plugins/builds/steps/active.js
```diff
@@ -0,0 +1,53 @@
...
+            return buildFactory.get(buildId)
+                .then((buildModel) => {
+                    if (!buildModel) {
+                        throw boom.notFound('Build does not exist');
+                    }
+                    const stepModel = buildModel.steps.find(
+                        step => step.startTime && !step.endTime);
+
+                    if (!stepModel) {
+                        throw boom.notFound('Active step does not exist');
```
tkyi: should it throw error or just return empty?

plugins/builds/README.md
pritamstyz4ever: hmm, it could get mixed. Wondering if naming it latest or current could also end up being the same, thoughts?

plugins/builds/steps/active.js
pritamstyz4ever: Well, I think both works, returns a 404, and in worker we are looking for status 200.

plugins/builds/README.md
jithine: we can make status as a query param. It will be difficult to pull this of this off this path param. GET /builds/{id}/steps?status=active

plugins/builds/steps/active.js
```diff
@@ -0,0 +1,58 @@
...
+            }
+        },
+        handler: (request, reply) => {
+            const buildFactory = request.server.app.buildFactory;
+            const buildIdCred = request.auth.credentials.username
+                && request.auth.credentials.username.toString();
+            const buildId = request.params.id && request.params.id.toString();
+            const status = request.query.status;
+
+            if (status !== 'active') {
```
tkyi: if `status` param is not set, I would expect this endpoint to return all steps for a build

plugins/builds/steps/active.js
tkyi: I think this endpoint should return empty if it cannot find anything

plugins/builds/steps/active.js
pritamstyz4ever: Yeah, will add additional logic to the api based on status and step code as a separate feature.

Issue #1865, opened 2019-11-28
DB  size is growing


s-yoshika:
**What happened**:
DB  size is growing in proportion to the build number.
`builds` and `steps` tables occupies about 95% of the total data size in our environment.

`builds.steps` and `steps.command` are both same data.
And these steps are defined by user's yaml and it tends to have a big size.

**What you expected to happen**:
`builds.steps` are needlless anymore and can use  `steps` table insteadly to save DB capacity.

**How to reproduce it**:

PR #1911, opened 2020-01-14
fix(1865): [2] Remove build step

sugarnaoming:
## Context
As mentioned here screwdriver-cd/screwdriver#1865 , `builds.steps` and `steps.command` are both same data.
And these steps are defined by user's yaml and it tends to have a big size.

## Objective
- Use `stepFactory` to get steps instead of `buildFactory`.

## References
#1865 

## License


kkisic:
Need to fix lint

plugins/builds/get.js
```diff
@@ -23,27 +23,32 @@ module.exports = () => ({
...
 
-            return factory.get(request.params.id)
-                .then((model) => {
-                    if (!model) {
+            return Promise.all([buildFactory.get(request.params.id),
+                stepFactory.list({ params: { buildId: request.params.id } })])
+                .then(([buildModel, stepsModel]) => {
+                    if (!buildModel) {
                         throw boom.notFound('Build does not exist');
                     }
```
kkisic: Do we need to check `stepModel` as same as `buildModel` ?

test/plugins/builds.test.js
```diff
@@ -2728,41 +2726,14 @@ describe.only('build plugin test', () => {
...
-                assert.deepProperty(reply.result, 'name', 'test');
-                assert.deepProperty(reply.result, 'code', 0);
-                assert.deepProperty(reply.result, 'endTime', options.payload.endTime);
-                assert.notDeepProperty(reply.result, 'startTime');
-            });
-        });
-
-        it('returns 200 when updating the code/endTime when the step model exists', () => {
+        it('returns 200 when updating the code/endTime when the step model exists', () =>
             server.inject(options).then((reply) => {
```
kkisic: Do we need to add return statement?

plugins/builds/get.js
```diff
@@ -23,27 +23,40 @@ module.exports = () => ({
...
 
-                    if (Array.isArray(model.environment)) {
-                        return reply(model.toJson());
+
+                    // This if statement should be removed after enough time has passed since build.steps removed.
+                    // Make orders of steps in completed builds sure,
+                    // because steps in old builds in DB have the order not sorted.
+                    if (buildModel.endTime) {
+                        stepsModel.sort((a, b) => new Date(a.startTime).getTime() - new Date(b.startTime).getTime());
+                    }
```
kkisic: Do we need to add the test case about these codes?

test/plugins/builds.test.js
sugarnaoming: This function is written in arrow function without braces.

kkisic: I see.

plugins/builds/get.js
sugarnaoming: Thank you.
I fixed

plugins/builds/get.js
sugarnaoming: Thank you.
added test

plugins/builds/get.js
kkisic: Do we need to add test cases for empty `stepsModel`?

plugins/builds/get.js
kkisic: Would you mind adding some test code for the case where `stepsModel` are in the order of `id` when `buildModel.endTime` is undefined?

plugins/builds/steps/logs.js
```diff
@@ -160,21 +160,13 @@ module.exports = config => ({
             }
         },
         handler: (req, reply) => {
-            const factory = req.server.app.buildFactory;
+            const factory = req.server.app.stepFactory;
```
tk3fftk: This should be renamed to `stepFactory` like `plugins/builds/get.js` for refactor.

plugins/builds/get.js
```diff
@@ -23,27 +23,28 @@ module.exports = () => ({
...
+            return buildFactory.get(request.params.id)
+                .then((buildModel) => {
+                    if (!buildModel) {
                         throw boom.notFound('Build does not exist');
                     }
 
-                    if (Array.isArray(model.environment)) {
-                        return reply(model.toJson());
+                    if (Array.isArray(buildModel.environment)) {
+                        return reply(buildModel.toJsonWithSteps());
```
tk3fftk: How about another `build.toJson()` cases like this line? Do we need adding steps same with here after removing `builds.steps`?
https://github.com/screwdriver-cd/screwdriver/blob/b9045682bcc43af1846a9cd9d15400fd79c5bdc5/plugins/builds/update.js#L184

plugins/builds/steps/logs.js
yuichi10: I fixed

plugins/builds/get.js
yuichi10: I fixed.

s-yoshika:
New steps endpoint seems appeared. We should fix it.
https://github.com/sonic-screwdriver-cd/screwdriver/tree/master/plugins/builds/steps

plugins/builds/steps/get.js
```diff
@@ -21,18 +21,10 @@ module.exports = () => ({
...
-                    if (!buildModel) {
-                        throw boom.notFound('Build does not exist');
-                    }
-
-                    const stepModel = buildModel.steps.filter(step => (
-                        step.name === request.params.name
-                    )).pop();
+            const stepFactory = request.server.app.stepFactory;
 
+            stepFactory.get({ buildId: request.params.id, name: request.params.name })
```
tkyi:
```suggestion
            return stepFactory.get({ buildId: request.params.id, name: request.params.name })
```

should we add `return`?

plugins/builds/steps/logs.js
```diff
@@ -160,21 +160,13 @@ module.exports = config => ({
...
-                .then((model) => {
-                    if (!model) {
-                        throw boom.notFound('Build does not exist');
-                    }
-
-                    const stepModel = model.steps.filter(step => (
-                        step.name === stepName
-                    )).pop();
-
+            stepFactory.get({ buildId, name: stepName })
```
tkyi:
```suggestion
            return stepFactory.get({ buildId, name: stepName })
```
same here?

yuichi10:
@tkyi  Thank you for your suggestion.

Issue #1716, opened 2019-08-09
artifact deep link


uevilwizard:
I need to be able to supply a link to a report that my build generates.   

Something like this:
`https://cd.screwdriver.cd/pipelines/<pipeline-id>/builds/<build-id>/artifacts/<report-filename>`

Additionally, I'd like to be able to supply something like 'latestSuccessfulBuild' as a build-id.   Then I would be able to supply an evergreen link to the latest copy of my report.  ([That's a separate issue.](https://github.com/screwdriver-cd/screwdriver/issues/1717))

Some cases where this feature is needed:
   - I have an HTML report artifact that contains hyperlinks to other artifacts.  Currently all of these links are broken.
   - Often if a test in my pipeline fails, resolving that failure will involve creating a Jira ticket containing a link to the failure.   Currently there is no way to do this.  

Without supporting links, it is quite difficult to use the artifacts generated by SDv4.

adong:
@uevilwizard this feature has been implemented.
Please refer https://github.com/screwdriver-cd/ui/pull/514

klu909:
added capability to link to latest build, following link format can be used
pipelines/{pipelineId}/jobs/{jobName}
pipelines/{pipelineId}/jobs/{jobName}/step/{stepName}
pipelines/{pipelineId}/jobs/{jobName}/artifacts
pipelines/{pipelineId}/jobs/{jobName}/artifacts/{artifactPath}

an optional status parameter can be used for links above to query latest build of specific build status, for example,
pipelines/{pipelineId}/jobs/{jobName}/artifacts?status=SUCCESS

anilkumarmyla:
Is there a quick way to generate the deep link prefix using env variable say `SD_ARTIFACTS_URL`  which would translate to `${SD_UI_URL}/pipelines/${SD_PIPELINE_ID}/builds/${SD_BUILD_ID}/artifacts/` ? 

This would be helpful to deep link an artifact using the url parameter in "Additional Pull Request Checks" https://docs.screwdriver.cd/user-guide/metadata#additional-pull-request-checks

PR #1917, opened 2020-01-18
[Don't Merge]feat(1716): enhance latest build end point

klu909:
## Context

currently latest build end point only returns latest build, we should allow fetching build in relative position to latest build, such as second to latest build.

## Objective

fetching build in relative position to latest build

## References

https://github.com/screwdriver-cd/screwdriver/issues/1716

## License


tkyi:
Can you update https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/pipelines/README.md#get-latest-build-for-a-single-job with more details like in https://github.com/screwdriver-cd/models/pull/428#issuecomment-575837628

klu909:
As per discussion with Jithin, this change was not necessary. Will pull this back for now

PR #1918, opened 2020-01-19
fix: allow pipeline token to disable and enable jobs

thrushanth:
## Context

to enable automated disable and enable of jobs via pipeline is not working as the token is enabled for only user.

## Objective

This will enable pipeline token to enable or disable the job

## License


jithine:
cc @klu909

jithine:
replaced with https://github.com/screwdriver-cd/screwdriver/pull/1933

PR #1924, opened 2020-01-23
fix: log unhandledRejection promise errors

jithine:
## Context

With newest API release we are seeing an uptick in API calls made to Git. This is causing Git to sometimes reject API requests to to API abuse protections. Screwdriver API is not handling these errors properly and logs lack proper stack trace to debug which logic is causing this. 
```
(node:18) UnhandledPromiseRejectionWarning: HttpError: You have triggered an abuse detection mechanism. Please wait a few minutes before you try again.
```

## Objective

Add error log for `unhandledRejection` so that we can debug which application logic is causing this.

## License


lib/server.js
```diff
@@ -4,6 +4,10 @@ const Hapi = require('hapi');
 const registrationMan = require('./registerPlugins');
 const logger = require('screwdriver-logger');
 
+process.on('unhandledRejection', (reason, p) => {
```
pritamstyz4ever: Unhandled Promise rejection is getting Deprecated so we should also start adding catch blocks to code blocks where we do new Promise so that we don't end up getting unhandled promise rejection errors.

jithine: noted. This is a catch all to start identifying areas we are missing proper error handling.

PR #1944, opened 2020-01-31
fix(1785): Fix to get parent event builds recursively

### LINKED ISSUE #1785: Restarted job does not always trigger subsequent jobs
**What happened**:

If a job is restarted from an event which is result of a previous restart, then context of previous  jobs which successfully completed are not carried over.

For example consider the following scenario

```
jobs:
  A:
    requires: [~commit]
  B:
    requires: [~commit]
  C:
     requires: [A, B]
```

1. In event E1, A is successful but B fails.
2. Restart job B in event E1 to create event E2. Job B fails again in event E2.
3. Restart job B in event E2 to create event E3. Job B is successful in event E3. But job C won't run in event E3.
4. Restart Job B in E1 to create E4 where B is successful. Job C will now start running as part of event E4

**What you expected to happen**:

Context of an jobs in an event should be traceable back to original event from which Job restarts started. In the scenario described above in Step 3 job C should have started in event E3.

**How to reproduce it**:

This problem manifests when a job is being restarted across multiple events.

**Updated** to add one more scenario https://github.com/screwdriver-cd/screwdriver/issues/1785#issuecomment-589771683

kumada626:
## Context

fix #1785 

## Objective

In getFinishedBuilds, it gets all parent event and merges parent build status data recursively.

## References

#1785 

## License

### DIFF
--- a/plugins/builds/index.js
+++ b/plugins/builds/index.js
@@ -607,6 +607,8 @@ async function getFinishedBuilds(event, eventFactory) {
         return event.getBuilds();
     }
 
+    // If parent event id, merge parent build status data recurively and
+    // rerun all builds in the path of the startFrom
     const parentEvent = await eventFactory.get({ id: event.parentEventId });
     const parents = await getFinishedBuilds(parentEvent, eventFactory);
     const upstreamBuilds = await removeDownstreamBuilds({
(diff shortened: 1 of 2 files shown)


kumada626:
This gets all parent event, not all events of pipeline.

Issue #1091, opened 2018-05-25
Show popularity of templates and commands


jithine:
## Context

With ever increasing number of commands & templates, users need means to know which ones are popular and in heavy usage. This helps users in picking up the appropriate template based on usage of their peers.

## Objective

1. Find a metric to track for commands. 
1. Find a metric to track for templates. 
1.1 Metric can be stars, or usage statistics
1. Display in UI a metric which show the popularity of a template.
1. Display in UI a metric which show the popularity of a command.

## References

1. Templates https://docs.screwdriver.cd/user-guide/templates
1. Commands https://docs.screwdriver.cd/user-guide/commands

jithine:
One problem with current implementation is that template information is not stored in job/build model. If template name was stored then we can easily query for number of builds using the template

tkyi:
Results:
<img width="1680" alt="Screen Shot 2020-02-14 at 11 59 53 AM" src="https://user-images.githubusercontent.com/3230529/74563409-9923d980-4f21-11ea-8c83-36339dd0e833.png">

DekusDenial:
@InderH If I stop a job from using a template, will the `Jobs` counter go down by 1?

InderH:
@DekusDenial yes the Jobs counter should go down, however, it looks like this is currently not the case. PR to fix this: https://github.com/screwdriver-cd/models/pull/436

kumada626:
Currently, if a job that uses template is archived, `Jobs` counter does not go down.
I think the `Jobs` counter should go down in such case.

PR #1952, opened 2020-02-05
feat(1091): add endpoint to list template versions and associated popularity metrics

InderH:
## Context

Need to return associated popularity metrics with template versions so the metrics can be displayed on the templates page.

## Objective

This PR adds an endpoint that returns template versions along with their associated popularity metrics.

## References

https://github.com/screwdriver-cd/screwdriver/issues/1091

## License


test/plugins/templates.test.js
```diff
@@ -387,6 +389,56 @@ describe('template plugin test', () => {
...
+
+        beforeEach(() => {
+            options = {
+                method: 'GET',
+                url: '/templates/screwdriver%2Fbuild/metrics'
+            };
+        });
+
+        it('returns 200 and all template versions and metrics for a template name', () => {
+            templateFactoryMock.listWithMetrics.resolves(testtemplateversionsmetrics);
```
klu909: it's better to add this line `assert.expect(x);` with `x` being number of asserts in your test

InderH: assert.expect() does not exist in this repo

test/plugins/templates.test.js
```diff
@@ -10,6 +10,7 @@ const testtemplate = require('./data/template.json');
 const testtemplates = require('./data/templates.json');
 const testtemplatetags = require('./data/templateTags.json');
 const testtemplateversions = require('./data/templateVersions.json');
+const testtemplateversionsmetrics = require('./data/templateVersionsMetrics.json');
```
tkyi: camelcase would be better
```suggestion
const testTemplateVersionsMetrics = require('./data/templateVersionsMetrics.json');
```

InderH: updated it, thanks

Issue #1722, opened 2019-08-20
Use Node.js v12


tk3fftk:
Screwdriver.cd is using Node.js v8. It will be EOL in 2019-12-31.   
We should use Node.js v10.

Note: Node.js v12 will be LTS since 2019-10-22 but I think it's hard to bump 2 major versions at one time.
Reference: [Releases | Node.js](https://nodejs.org/en/about/releases/)

tk3fftk:
[Node.js 12 to LTS and Node.js 13 is here! - Node.js Foundation - Medium](https://medium.com/@nodejs/node-js-12-to-lts-and-node-js-13-is-here-e28d6a4a2bd)

jithine:
We should go directly to node 12. Also this will be a good opportunity to upgrade our `hapi` dependencies. We are running very old versions.

adong:
Upgrade from `node:10` to `node:12`
UI PR: https://github.com/screwdriver-cd/ui/pull/528

jithine:
Pending ones are queue-worker, since we are working on a `queue-service` node:12 will be addressed in that project https://github.com/screwdriver-cd/screwdriver/issues/1951

jithine:
Closing this, Rest of the libraries/services in node:8 will be updated on a case by case basis.

PR #1957, opened 2020-02-10
fix: use node:12 container

jithine:
## Context

Screwdriver API is running on node:8 container which is no longer supported.

## Objective

Upgrade to latest LTS of node. 

## References

https://github.com/screwdriver-cd/screwdriver/issues/1722

## License

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:8
+FROM node:12
 
 # Screwdriver Version
 ARG VERSION=latest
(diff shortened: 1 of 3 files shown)


tkyi:
Why are we getting rid of jenkins-mocha?

jithine:
> Why are we getting rid of jenkins-mocha?
It's pinned to an older version of mocha, and is just a wrapper on top of `nyc` & `mocha`. I don't see an extra value in having this.

tkyi:
@jithine we should be consistent and remove it in all repos then

jithine:
> @jithine we should be consistent and remove it in all repos then

@tkyi yep, one repo at a time. we will have to update all repos to build on node:12. Will update one by one.

Issue #1710, opened 2019-08-06
Feature request: support parallel fork join for remote jobs


updong:
**What happened**:
I found out that I cannot use AND conditions on remote jobs 

```
"sd@223215:qa-minos-functional-test-us" fails to match the required pattern: /^~(sd@\d+:[\w-]+|(pr|commit|release|tag)(:(.+))?)$/,
```
and found out that SD doesn't support parallel fork join for remote jobs

**What you expected to happen**:
something like this. 

*A job which can wait for multiple remote job with AND condition*
```
  requires:
      - sd@123:A
      - sd@123:B
      - sd@456:C
      - sd@456:D
      - E
```
this job will wait for A B C D and E which runs in parallel and gets triggered when they're all successful.

**How to reproduce it**:
try adding remote job in requires field without `~`

vnugopal:
@DekusDenial

d2lam:
This can be implemented only if `sd@123:A, sd@123:B, sd@456:C, sd@456:D, E` all start from a single point, which will look like this:
![Image from iOS (2)](https://user-images.githubusercontent.com/3401924/65733140-1ee1de00-e082-11e9-91c1-04377fce7b37.jpg)

Without this constraint, it will get into race conditions & inconsistencies. For example, `sd@123:A` runs 1 week ago, `B,C,D,E` run now. Should `F` start or not? It's also almost impossible to implement since there is no way for `F` to track whether `A,B,C,D,E` finish in the same run because they can belong to different events. These events' end times can be within several minutes or several hours. How do we know which ones are supposed to be grouped together? This is probably why Jenkin's Join Plugin also relies on this constraint. 

I assume this was your intention as well since you mention `A B C D and E which runs in parallel ` @updong 

I can start with some design doc on how we implement this and people can review.

updong:
@d2lam yes I forgot to mention about that constraint. yes I was assuming that constraint to be forced. then need to be triggered by same event.

d2lam:
[WIP]
### DESIGN
Using the above example: `A -> [sd@123:A, sd@123:B, sd@456:C, sd@456:D, E] -> F`
Let's say A and F belong to `sd@789`

- After A is done, it figures the next jobs to run are `[sd@123:A, sd@123:B, sd@456:C, sd@456:D, E]` based on `sd@789` workflow graph. Let's say A's `eventId = 1`
- Trigger all these jobs in parallel with `upstreamEvent = 1`
[...]

PR #1960, opened 2020-02-11
fix(1710): Restart internal fork

tkyi:
## Context

Sometimes users might want to restart a job that is part of a join. The join job should still trigger as a result if the parent event had successful parent builds.

## Objective

This PR fixes restart case for internal join when external join flag is true. Also passes in parentBuilds info from event create.
_TODO: Still need to fix restart case for external join_

## References

Related to #1710 

## License


plugins/builds/index.js
```diff
@@ -720,17 +669,39 @@ async function createOrRunNextBuild({ buildFactory, jobFactory, eventFactory, pi
...
-    });
+    let nextBuild;
+
+    // If next build is external, return the latest build with same job ID
+    if (isExternal) {
+        const p = await pipelineFactory.get(externalPipelineId);
+        const jobArray = await p.getJobs({ params: { name: externalJobName } });
+        const j = await jobFactory.get(jobArray[0].id);
+
+        nextBuild = await j.getLatestBuild({ status: 'CREATED' });
```
jithine: will this be accurate ? shouldn't it lookup based on the original event  ID  in `externalPipeline`?

ie rather than doing `job/builds` we should grab `nextBuild` from `event/builds`

plugins/builds/index.js
```diff
@@ -720,17 +669,39 @@ async function createOrRunNextBuild({ buildFactory, jobFactory, eventFactory, pi
...
+                }
+            });
+        });
+    }
+
+    // If next build is internal, look at the finished builds for this event
+    const jobId = workflowGraph.nodes.find(node =>
+        node.name === trimJobName(nextJobName)).id;
+
+    nextBuild = finishedInternalBuilds.find(b => b.jobId === jobId);
```
jithine: I don't see an `else` wrap here, so this `nextBuild` will over write `external` logic from line `675`

plugins/events/create.js
```diff
@@ -22,32 +22,32 @@ module.exports = () => ({
...
+                            .then((j) => {
+                                const restartConfig = {
+                                    pipelineId: j.pipelineId,
+                                    startFrom: j.name,
+                                    parentBuildId: b.parentBuildId,
+                                    parentEventId: b.eventId
+                                };
+
+                                if (parentBuilds) {
+                                    restartConfig.parentBuildsInfo = parentBuilds;
```
jithine: can't this just be `parentBuilds` instead of `parentBuildsInfo` ?

plugins/events/create.js
```diff
@@ -76,6 +77,10 @@ module.exports = () => ({
                     payload.parentBuildId = parentBuildId;
                 }
 
+                if (parentBuilds || parentBuildsInfo) {
```
jithine: based on previous comment, can't his just be passed `parentBuilds` here ? instead of introducing a `parentBuildsInfo`, can't we just have `parentBuilds` ?

plugins/events/create.js
tkyi: it's defined above so I needed to use another name

plugins/events/create.js
tkyi: it can be passed into the event start config or gotten from a build

plugins/events/create.js
jithine: can't it be `restartConfig.parentBuilds = parentBuilds` ?

then https://github.com/screwdriver-cd/screwdriver/pull/1960/files#diff-6590031cbab3d825a0a4cd19e5da419bR80 can change to

```
if (parentBuilds)```

plugins/builds/index.js
tkyi: fixed

plugins/builds/index.js
tkyi: updated

PR #1972, opened 2020-02-15
feat(1710): Restart external and internal join case

tkyi:
## Context

Restarting a job that is part of a join workflow should work for internal and external parallel fork joins.

## Objective

This PR adds logic so internal and external joins should work.

## References

Related to #1710 

## License


plugins/builds/index.js
```diff
@@ -910,16 +953,46 @@ exports.register = (server, options, next) => {
                     const jobId = parentWorkflowGraph.nodes.find(node =>
                         node.name === trimJobName(externalJobName)).id;
                     // Get next build
-                    const nextBuild = finishedExternalBuilds.find(b => b.jobId === jobId);
+                    const nextBuild = finishedExternalBuilds.find(b => b.jobId === jobId
+                        && b.status === 'CREATED');
+                    const isRestartCase = finishedExternalBuilds.find(b => b.jobId === jobId
```
jithine: if I understand this correctly, this is`isRestartCaseWithSuccessfulBuild` any other scenario this build won't exist and `nextBuild` will be in `CREATED` state

plugins/builds/index.js
```diff
@@ -929,27 +1002,46 @@ exports.register = (server, options, next) => {
...
+
+                            const newEvent = await createExternalBuild({
+                                pipelineFactory,
+                                eventFactory,
+                                externalPipelineId: externalEvent.pipelineId,
+                                externalJobName,
+                                parentBuildId: build.id,
+                                parentBuilds: deepmerge.all([parentBuildsForJoin, parentBuilds]),
+                                causeMessage: `Triggered by sd@${pipelineId}:${currentJobName}`,
+                                parentEventId: event.id,
```
jithine: `parentEventId` for the external pipeline's event should be the event from which it was restarted, ie the original event where this external job's build failed.

plugins/builds/index.js
```diff
@@ -929,27 +1002,46 @@ exports.register = (server, options, next) => {
...
-                            username,
-                            scmContext,
-                            build: parentBuild, // this is the parentBuild for the next build
-                            baseBranch: event.baseBranch || null,
-                            parentBuilds: externalParentBuilds,
-                            parentBuildId: build.id,
-                            start: false
-                        });
+                        // if restart case, should create event
+                        if (isRestartCase) {
```
jithine: `isRestartCase` build very much specific to external join build restart scenario. We need a name to reflect it

plugins/builds/index.js
tkyi: the previous event could have had successful runs for the fork jobs and just failed during the join job

plugins/builds/index.js
tkyi: if you mean the event with the parent to the fork job that was restarted, I'm not sure if that makes sense

jithine: if `parentEventId` points to the external pipeline's event Id, then how can we associate the 2 events in original pipelines together ?

tkyi: so `parentEventId` will always only refer to events within the same pipeline?

tkyi: just seems a little inconsistent because if the main pipeline kicks off an external fork, I would expect the `parentEventId` to be the main pipeline.

jithine: so then do we need a different model field to group events in a pipeline other than `parentEventId` ?
If we keep `parentEventId` to strictly track the `eventId` of the build from which current builds is triggered - whether internal or external, then we need a different mechanism for grouping events together (for grouping events in same pipeline which is part of a `restart` situation)

jithine: This was not a problem without remote parallel join, since `parentEventId` can always track back to an event in same pipeline covering restart situation. But that's not so the case with parallel join. If we traverse back `parentEventId` we will get a situation like this.

`e1InternalPipeline -> e2ExternalPipeline(failed) -> e3ExtenalPipeline -> e2InternalPipeline`

Now grouping e1 & e1 in Internal Pipeline will not be possible with `parentEventId` spread across pipelines

tkyi: Yes, I think it would be ideal to use a different way to group events. We're currently overloading `parentEventId` to track restarted events as well as parallel fork joins. I think this can get kind of confusing.

jithine: agree we can tackle that when working on event grouping.

PR #1979, opened 2020-02-27
fix(1710): UNSTABLE build should not trigger join job

tkyi:
## Context

Some bugs with the new parallel fork join feature:
- UNSTABLE builds should be considered a build failure
- pipelines with a single external chain (no join) should start the build; this only happens when the parent builds contains a pipeline ID that matches the next build's pipeline ID

## Objective

This PR fixes the above issues.

## References

Related to #1710 

## License


plugins/builds/index.js
```diff
@@ -995,6 +995,21 @@ exports.register = (server, options, next) => {
...
+                        chainPR: externalPipeline.chainPR
+                    });
+                    const joinObjForJoin = nextJobsForJoin.reduce((obj, jobName) => {
+                        obj[jobName] = workflowParser
+                            .getSrcForJoin(parentWorkflowGraph, { jobName });
+
+                        return obj;
+                    }, {});
+                    const joinListForJoin = joinObjForJoin[externalJobName];
+                    const joinListNamesForJoin = joinListForJoin ?
```
jithine: if this is `undefined`, won't it break `getParentBuildStatus` https://github.com/screwdriver-cd/screwdriver/pull/1979/files#diff-7a514e440aa937beee722c7ac84dfb93R1075
Because `getParentBuildStatus` https://github.com/screwdriver-cd/screwdriver/blob/fc877d624044bda152fccef0c029404004a64a1d/plugins/builds/index.js#L542-L549 assumes it's an array

Issue #965, opened 2018-03-30
$GIT_BRANCH invalid value in PR builds?


drewfish:
**What happened**:
`git diff $GIT_BRANCH --numstat` results in `Invalid object name 'origin/refs/pull/14/head'`

**What you expected to happen**:
`$GIT_BRANCH` is a valid value which can be passed to `git diff`

**How to reproduce it**:

drewfish:
In a PR build I did a `tree .git/refs` and saw this:
```
.git/refs
 |-- heads
 | |-- master
 | `-- pr
 |-- remotes
 | `-- origin
 | `-- HEAD
 `-- tags
```

tkyi:
Are you expecting the branch name of the PR?

lfjewett:
This is the closest issue I could find to what we're searching for.  We have a pipeline workflow that runs a set of templates against different accounts depending on the branch.  Commits to the develop branch run templates in a DEV account and Commits to master run them in production.   It's fairly simple to write logic to switch the account parameters based off the GIT_BRANCH value for this job.  However if a job is started as part of ~pr then we have no idea if that PR is headed to develop or master and no easy way to find out.

One work around is to grab the value using a github API token with the following; 

`https://api.github.com/repos/org/project/pulls/10 | jq '.base["ref"]'`

It would be **very** helpful if Screwdriver provided this information for us instead, and would make for much simpler and cleaner pipelines.  Perhaps someone else has an easier way to find the real branch for a PR?

jithin1987:
What we could do is make `GIT_BRANCH` always be the target Git branch for which the job is configured. And if it's a Pull Request then introduce `GIT_PR_REF` which will hold the current `GIT_BRANCH` value for a Pull Request

yoshwata:
I hope that `GIT_BRANCH` would be the branch name that we can push to.

PR #1980, opened 2020-02-27
fix(965): [4]$GIT_BRANCH invalid value in PR builds

ibu1224:
## Context

I fix [issue](https://github.com/screwdriver-cd/screwdriver/issues/965).
I add new variables,`PR_BRANCH_NAME`.
I fix the `data-schema`,`scm-base` and `scm-github` accordingly.

This PR is requires below changes before merge.
[1] https://github.com/screwdriver-cd/data-schema/pull/386
[2] scm-base https://github.com/screwdriver-cd/scm-base/pull/75
[3] scm-github https://github.com/screwdriver-cd/scm-github/pull/150
## Objective
I modify it to return `prSource`.
Because I need `fork` information to set prBranchName.

## License


test/plugins/builds.test.js
```diff
@@ -1339,6 +1341,8 @@ describe('build plugin test', () => {
                             scmContext,
                             eventId: 'bbf22a3808c19dc50777258a253805b14fb3ad8b',
                             configPipelineSha,
+                            prSource: '',
+                            prInfo: '',
```
kumada626: This is not correct test case. This test case is the chainPRs test case. Therefore, in this test case, `prSource` and `prInfo` should have a value.

kumada626:
In the following test case also should have a value for `prSource` and `prInfo`.
https://github.com/screwdriver-cd/screwdriver/blob/b95467300416633e578311d3a6f816e3e95eb11f/test/plugins/builds.test.js#L1597

test/plugins/builds.test.js
jithine: @kumada626 @ibu1224 can this be fixed in a subsequent PR. I need to merge this since we need to get an internal release out for unblocking Parallel join.

ibu1224:
@jithine @kumada626 
Thank you review.
I create another PR to fix the test.

PR #1998, opened 2020-03-03
design(1314): add `--disable-image-pull` option

MysticDoll:
## Context

We are now developing `sd-local` and it always pulls build image each execution.
Therefore this PR suggests adding the option to disable that behaviour to make users run jobs on docker images in local machine.

## Objective
- add `--always-pull-disable` option.

## References
- #1314 
- screwdriver-cd/sd-local/pull/11

## License

### DIFF
--- a/design/sd-local.md
+++ b/design/sd-local.md
@@ -135,7 +135,7 @@ $ sdlocal build [job-name] [options]
 - `--artifacts-dir [path]` Path to the host side directory which is mounted into `$SD_ARTIFACTS_DIR`. (default: `./sd-artifacts`)
 - `-m, --memory [size]` Set memory size which Build Container can use. Either b, k, m, g can be used as a size unit. (default: ?)
 - `--src-url [repository url]` Set repository URL which is to build when user use the remote repository without local files.
-- `--always-pull-disable` Disable `sd-local` from always pulling build image.
+- `--disable-image-pull` Disable `sd-local` from always pulling build image.
 
 #### Output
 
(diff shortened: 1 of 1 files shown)


s-yoshika:
When we can use this option? The build with this option can't ensure the same build as the actual SD build, because SD builds always pull images.

yuichi10:
I think sometime user want to use this option for testing image of build.

s-yoshika:
I see, but it's hard to guess the usage from this name.

MysticDoll:
> it's hard to guess the usage from this name.

how about `--use-local-image`?

jithine:
can we make the default to use cached image ? like how `docker run` works.

s-yoshika:
> can we make the default to use cached image ? like how docker run works.

I concern it can't guarantee the reproducibile build, for example the image which has latest tag might not be latest without pulling. And the build environment can be different from those of actual Screwdriver build.

design/sd-local.md
```diff
@@ -135,6 +135,7 @@ $ sdlocal build [job-name] [options]
 - `--artifacts-dir [path]` Path to the host side directory which is mounted into `$SD_ARTIFACTS_DIR`. (default: `./sd-artifacts`)
 - `-m, --memory [size]` Set memory size which Build Container can use. Either b, k, m, g can be used as a size unit. (default: ?)
 - `--src-url [repository url]` Set repository URL which is to build when user use the remote repository without local files.
+- `--always-pull-disable` Disable `sd-local` from always pulling build image.
```
jithine: can we call it `--disable-image-pull` ?

MysticDoll: I fixed it. It seems better than before. thx

PR #2010, opened 2020-03-05
feat(1785): Add groupEventId to event create [3]

tkyi:
## Context

When users restart builds in events, it's not always clear which event to restart failed builds from. Due to this confusion, the join build doesn't always start as expected.
This logic can be made simpler by grouping all events with the same parent.

## Objective

This PR adds `groupEventId` to events during creation to allow for easier grouping. When an event is not restarted (ie is original), it will not have a `groupEventId`.

## References

Related to #1785 
Blocked by https://github.com/screwdriver-cd/models/pull/440/files

## License


plugins/events/create.js
```diff
@@ -39,11 +39,23 @@ module.exports = () => ({
...
-                                    parentEventId: b.eventId
+                                    parentEventId: b.eventId,
+                                    groupEventId: request.payload.groupEventId,
+                                    parentBuilds: b.parentBuilds
                                 };
 
-                                if (b.parentBuilds) {
-                                    restartConfig.parentBuildsInfo = b.parentBuilds;
+                                if (!restartConfig.parentBuilds) {
+                                    restartConfig.parentBuilds = request.payload.parentBuilds;
```
klu909: does this require ui changes?

tkyi: should be the same logic as before, just refactored

plugins/builds/index.js
```diff
@@ -267,17 +274,18 @@ async function createEvent(config) {
...
- * @param  {Number}   config.parentBuildId      Parent Build Id
+ * @param  {Number}   config.parentBuildId      Parent Build ID
  * @param  {Object}   config.parentBuilds       Builds that triggered this build
  * @param  {String}   config.causeMessage       Cause message of this event
- * @param  {Boolean}  [config.start]            Whether to start the build after creating
+ * @param  {Number}   [config.parentEventId]    Parent event ID
+ * @param  {Number}   [config.groupEventId]     Group parent event ID
  * @return {Promise}
  */
 async function createExternalBuild(config) {
```
jithine: how about `createInternalBuild` function ? how is `groupEventId` handled there  ?

tkyi: `groupEventId` is a field for an `event`. `createExternalBuild` creates a new `event` (with subsequent builds), but `createInternalBuild` creates a single `build`, so there's no where to set the `groupEventId`.

Issue #1977, opened 2020-02-26
Users who are not cluster admin cannot trash Template/Command even if the Template/Command ware published by them.


kumada626:
**What happened**:
The Template/Command which was published by not cluster admin user cannot be trashed by them.
Only cluster admins can trash Template/Command.
Users who have authority to repository that published Template/Command should be able to trash Template/Command.

**What you expected to happen**:
Users who have authority to repository that published Template/Command can trash Template/Command.

**How to reproduce it**:
1. Users who are not cluster admin publish Template/Command.
2. They cannot trash that Template/Command.

jithine:
Does this happen, when user deleted the pipeline which published the command/template in between ?

kumada626:
This happens not only  when user deleted the pipeline which published the command/template, but also even if the pipeline which published the command/template exists.

PR #2015, opened 2020-03-09
fix(1977): [1] The pipeline admin deletes template/command

ibu1224:
## Context
I fix [issue](https://github.com/screwdriver-cd/screwdriver/issues/1977).
Currently, template / command can only be deleted by cluster admin.
It is inconvenient because the admin of the template/command cannot be deleted.
Therefore, I modify it so that the pipeline admin can  be deleted.

## Objective
I modified it to look at the pipeline admin.

## References
https://github.com/screwdriver-cd/screwdriver/issues/1977

## License


plugins/templates/index.js
```diff
@@ -53,7 +53,7 @@ exports.register = (server, options, next) => {
                     }
 
                     return user.getPermissions(pipeline.scmUri).then((permissions) => {
-                        if (!permissions[permission]) {
+                        if (!permissions[permission] && username in pipeline.admins === false) {
```
kumada626: This `permissions` returns from scm. Therefore, the user is an admin of a pipeline that publishes template, pemissions[admin] might be true. I think that this PR is not needed to fix issue.

ibu1224: It was as pointed out. I close this PR.

Issue #1951, opened 2020-02-05
API performance impacted by executor-queue


jithine:
**What happened**:

`executor-queue` is responsible for running periodic build jobs and enforcing `freezeWindows`. However executor-queue which uses `node-resque` also runs as part of API and is competing for resources. API container is run with a simple `node start` and does not have capability to make use of multiple cpu cores if allocated. So it's important that API is kept light weight and heavy processing logic is split into separate micro services.

**What you expected to happen**:

API server should be light weight and should serve only API requests. Refactor code base to move `executor-queue` logic under `queue-worker`. API should only enqueue builds into the queue. Rest of logic should be moved under `queue-worker`

**How to reproduce it**:

N/A

**References**

1. https://github.com/screwdriver-cd/executor-queue
1. https://github.com/screwdriver-cd/queue-worker

jithine:
Pulling this back to `Doing` since we found few road blocks in deploying this to production.

1. Periodic builds stopped working.
    1. We noticed that while build schedule times are being updated, periodic build did not get created. Scheduler did not even call create event end point to create those builds.
   1. We suspect it's caused by the way the scheduler/worker is initialized.
   1. Also need to evaluate error handling, whether it stops working in response to an error
1. Circuit breaker errors for large pipeline.
    1. This is a scalability issue for pipelines with large number of jobs. 
    1. https://github.com/screwdriver-cd/models/pull/443 should address this. 
1. Error in fetching pipeline admin.
   1. This might be the reason for create event failing. 
   1. But could also be caused by pipelines who no longer have valid admin. Need further checking.

PR #2018, opened 2020-03-10
feat(#1951): Add api to get pipeline admin

pritamstyz4ever:
## Context

To create new queue service we need the pipeline admin info.

## Objective

This PR adds a new API /pipeline/$id/admin to get the first admin of pipeline

## References

https://github.com/screwdriver-cd/screwdriver/issues/1951

## License


plugins/pipelines/admins/get.js
```diff
@@ -0,0 +1,50 @@
...
+module.exports = () => ({
+    method: 'GET',
+    path: '/pipelines/{id}/admin',
+    config: {
+        description: 'Get the pipeline admin',
+        notes: 'Returns a pipeline admin record',
+        tags: ['api', 'pipelines'],
+        auth: {
+            strategies: ['token'],
+            scope: ['user', 'build', 'pipeline']
```
jithine: shouldn't `temporal` token have access ? since that's what's used by the `queue-service` ?
Ideally we should have a token which represents `queue-service`

pritamstyz4ever: Correct, adding temporal cope

Issue #1967, opened 2020-02-13
Aggregate view for a pipeline


jithine:
**What happened**:

Users do not have a view to see the overall status of a pipeline.

**What you expected to happen**:

UI Mock

![image (1)](https://user-images.githubusercontent.com/193126/74471030-6233c200-4e54-11ea-99ec-040b59aa91bf.png)

<img width="2440" alt="Screen Shot 2020-04-07 at 3 24 52 PM" src="https://user-images.githubusercontent.com/193126/78725270-fb76d600-78e3-11ea-8de4-a3a2934d92a4.png">

1. Provide an aggregate button in UI to display a workflow graph which shows latest build across current active jobs.
1. API should support an end point which returns latest build across current active jobs.

**How to reproduce it**:

N/A

jithine:
Few more pending items

- [ ] Build order list should be oldest -> newst. Reverse current order.
- [ ] Actions `Play` button should start a fresh event and not restart last successful  
- [ ] Job order should be based on position in workflow graph. 
        1. Group Jobs based on depth in workflow graph
        2. Within a job group sort alphabetically
- [ ] Build stats context window - should display  SCM `sha` and an option to restart
- [ ] Remove link from underlink of build bubbles. 



- [ ]

tk3fftk:
It looks great feature 👍 
By the way, is it OK that `models` has specific queries? Should these implementation are inside of `datastore` implementation?
Also, should we have database specific queries? It's may be hard to know where should be updated when we or someone add another database implementation.

adong:
> It looks great feature 👍
> By the way, is it OK that `models` has specific queries? Should these implementation are inside of `datastore` implementation?
> Also, should we have database specific queries? It's may be hard to know where should be updated when we or someone add another database implementation.

Are you referring to PR: https://github.com/screwdriver-cd/models/pull/441/files#diff-43e1dd855c9ebdae88cbf0102c6e05b2R8
 
Good question, let me escalate

tk3fftk:
@adong Thank you for escalation. Should I create another issue?

jithine:
@tk3fftk My replies inline

> By the way, is it OK that `models` has specific queries? Should these implementation are inside of `datastore` implementation?

[...]

PR #2020, opened 2020-03-11
feat(1967): add pipeline list view

InderH:
## Context

Need to add api endpoint to return build status history for jobs so they can be displayed in pipeline list view.

## Objective

Add getBuildStatuses endpoint to return build statuses for jobs.

## References

https://github.com/screwdriver-cd/screwdriver/issues/1967

## License


plugins/builds/getBuildStatuses.js
```diff
@@ -0,0 +1,55 @@
+'use strict';
+
+const boom = require('boom');
+const joi = require('joi');
+const stringSchema = joi.string().regex(/^[0-9]+$/);
```
klu909: why string? use joi.number().positive()

plugins/builds/getBuildStatuses.js
```diff
@@ -0,0 +1,55 @@
...
+const jobIdsSchema = joi.array().items(stringSchema);
```
klu909: const schema = require('screwdriver-data-schema');
const jobIdsSchema = joi.reach(schema.models.job.base, 'id');

plugins/builds/getBuildStatuses.js
```diff
@@ -0,0 +1,55 @@
...
+            'hapi-swagger': {
+                security: [{ token: [] }]
+            }
+        },
+        handler: (request, reply) => {
+            const buildFactory = request.server.app.buildFactory;
+            const { jobIds, numBuilds, offset } = request.query;
+
+            const payload = {
+                jobIds: jobIds.map(jobId => parseInt(jobId, 10)),
```
klu909: these parameters should of been passed in as integer

plugins/builds/getBuildStatuses.js
InderH: Because the parameters are from request.query they all come in as strings. Regardless of how they are passed in when calling the API they are converted to string when they are sent.

plugins/builds/getBuildStatuses.js
InderH: parameters are from request.query so they all come in as strings

plugins/builds/getBuildStatuses.js
InderH: parameters are from request.query they all come in as strings

plugins/auth/token.js
```diff
@@ -20,9 +20,7 @@ module.exports = () => ({
             scope: ['user', 'pipeline']
         },
         plugins: {
-            'hapi-swagger': {
-                security: [{ token: [] }]
-            }
+            'hapi-swagger': { security: [{ token: [] }] }
```
InderH, line 23: Reduced number of lines so that there would no longer be a warning and thus the number of warnings would not exceed the maximum limit.

Is there something better that can be done instead?

adong: @InderH Please rebase.

plugins/builds/getLatestBuild.js
```diff
@@ -0,0 +1,61 @@
...
+                    jobId
+                }
+            };
+
+            if (status) {
+                listConfig.params.status = status;
+            }
+
+            return buildFactory
+                .list(listConfig)
```
adong, line 41: @InderH  Good, just not sure why `listConfig` is an object consists params? @klu909 is this the way the APIs written?

plugins/auth/token.js
InderH: @adong do you want me to rebase specifically or is merging in master ok?

plugins/builds/getLatestBuild.js
InderH: config.params are the parameters that are used to filter when doing buildFactory.list

https://github.com/screwdriver-cd/models/blob/56e3ce9b02c68b99ca5848f3e9e029702f742f11/lib/buildFactory.js#L296

plugins/auth/token.js
adong: I usually do rebase, but merging master is fine too.

adong:
Once https://github.com/screwdriver-cd/models/pull/441 is merged, we are good to go.

plugins/builds/getBuildStatuses.js
```diff
@@ -0,0 +1,56 @@
+'use strict';
+
+const boom = require('boom');
+const joi = require('joi');
+const stringSchema = joi.string().regex(/^[0-9]+$/);
+const jobIdsSchema = joi.alternatives().try(joi.array().items(stringSchema), stringSchema);
```
tkyi, line 6: should this field be required?

plugins/builds/getLatestBuild.js
```diff
@@ -0,0 +1,61 @@
...
+const boom = require('boom');
+const joi = require('joi');
+const schema = require('screwdriver-data-schema');
+const buildSchema = schema.models.build.get;
+const jobIdSchema = joi.string().regex(/^[0-9]+$/);
+const statusSchema = joi.reach(schema.models.build.base, 'status');
+
+module.exports = () => ({
+    method: 'GET',
+    path: '/builds/latest',
```
tkyi, line 12: would it be better to modify/combine with this endpoint? https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/jobs/latestBuild.js
I think that path makes more sense to me

plugins/builds/update.js
```diff
@@ -195,8 +195,11 @@ module.exports = config => ({
                                     buildLink: `${buildFactory.uiUri}/pipelines/${pipeline.id}/builds/${id}`
                                 });
 
+                                const skipFurther = /\[(skip further)\]/.test(newEvent.causeMessage);
```
tkyi, line 198: what is this?

InderH: From the list view you can build a single job; having "skip further" in the message indicates that it shouldn't build any jobs that are after this job in the pipeline.

plugins/builds/getLatestBuild.js
InderH: 👍 , Completely missed this endpoint earlier

plugins/builds/getBuildStatuses.js
InderH: Are you referring to jobIdsSchema as a whole or to having a minimum length for the array?

tkyi: the jobIdsSchema

InderH: Not following what you mean by required, do you mean something like `const jobIdsSchema = require(...);`? Or is this referring a property in joi that explicitly makes something required?

plugins/builds/getLatestBuild.js
InderH: Done

plugins/builds/getBuildStatuses.js
tkyi: ah the second one
since the query would be kind of pointless without jobIds passed in, right?

tkyi:
Need to pull in latest datastore-sequelize ^v6.0.0 package now: https://github.com/screwdriver-cd/screwdriver/blob/master/package.json#L100

InderH:
> Need to pull in latest datastore-sequelize ^v6.0.0 package now: https://github.com/screwdriver-cd/screwdriver/blob/master/package.json#L100

@tkyi Updated - I'm assuming the v shouldn't be there? i.e. "screwdriver-datastore-sequelize": "^6.0.0", not "screwdriver-datastore-sequelize": "^v6.0.0"

plugins/builds/update.js
tkyi: is this a part of https://github.com/screwdriver-cd/screwdriver/issues/1967? just wondering if it's documented somewhere

InderH: I don't think its documented anywhere; it was just something I talked with Jithin about when planning what the 3 buttons in each row of the list view should do

plugins/builds/getBuildStatuses.js
InderH: Added, thanks!

plugins/builds/update.js
jithine: @InderH we need to document this. Please update guide with this info.

plugins/builds/getBuildStatuses.js
jithine: don't we have anything existing in `data-schema` ?

plugins/builds/update.js
InderH: @jithine Which guide are you referring to?

jithine: At docs.screwdriver.cd. Possibly an FAQ section. Not sure where else it fits.

PR #2014, opened 2020-03-07
feat: use eslint-config-screwdriver@5

jithine:
## Context

Upgrade to eslint-config-screwdriver@5 which brings new eslint version and rules.

## References

https://github.com/screwdriver-cd/eslint-config-screwdriver/pull/40

## License
I confirm that this contribution is made under a BSD license and that I have the authority necessary to make this contribution on behalf of its copyright owner.

Issue #1650, opened 2019-05-23
Improve SD ESLint rules.


wdstar:
**What happened**:

Now SD JS codes have many big functions, which readability are very low and it's not easy to fix or improve it. 

In the current [eslint-config-screwdriver](https://github.com/screwdriver-cd/eslint-config-screwdriver), the following rules are not activated yet. We should activate those rules by `warn` level and refactor SD sources step by step.
- [max-lines-per-function](https://eslint.org/docs/rules/max-lines-per-function): default max: 50
- [complexity](https://eslint.org/docs/rules/complexity): default max: 20
- [max-params](https://eslint.org/docs/rules/max-params): default max: 3
- [max-statements](https://eslint.org/docs/rules/max-statements): default max: 10

**References**:

- Currently source codes violations (excludes `test/` directory):

|Rules|[screwdriver](https://github.com/screwdriver-cd/screwdriver)|[models](https://github.com/screwdriver-cd/models)|
|:---|---:|---:|
|max-lines-per-function|121|28|
|complexity|3|8|
|max-params|19|7|
|max-statements|34|21|

Please post any comments. (e.g. we should activate other rules,...)

jithin1987:
cc @DekusDenial @adong

DekusDenial:
@wdstar what do you say if we up `eslint-config-screwdriver` to `^4.0.0` and add `prettier` and following mostly https://github.com/screwdriver-cd/ui/blob/master/.eslintrc.js minus the ember part?

wdstar:
@DekusDenial At first SD API's [current sources](https://github.com/screwdriver-cd/screwdriver) has new 453 errors if we update `eslint-config-screwdriver` to `^4.0.0`. 28 errors still remain after hitting `eslint --fix .` It takes some cost to fix its errors (mostly `max-len` violation) manually.

```
$ git diff
diff --git a/package.json b/package.json
index f424079..d1ecb52 100644
--- a/package.json
+++ b/package.json
@@ -123,7 +123,7 @@
     "coveralls": "^2.11.12",
     "cucumber": "2.3.1",
     "eslint": "^5.16.0",
-    "eslint-config-screwdriver": "^3.0.1",
+    "eslint-config-screwdriver": "^4.0.0",
     "form-data": "^2.3.3",
     "jenkins-mocha": "^7.0.0",
     "mockery": "^2.0.0",
$ npm install
...
$ eslint .
...
[...]

PR #2023, opened 2020-03-12
feat(1650): use eslint-config-screwdriver@5

jithine:
## Context

Upgrade to eslint-config-screwdriver@5 which brings new eslint v6 version and updated airbnb base rules.

## References

https://github.com/screwdriver-cd/screwdriver/pull/2014
https://github.com/screwdriver-cd/eslint-config-screwdriver/pull/40
https://github.com/screwdriver-cd/screwdriver/issues/1650

## License


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "API server for the Screwdriver.cd service",
   "main": "index.js",
   "scripts": {
-    "pretest": "eslint .",
+    "pretest": "eslint --max-warnings 440 .",
```
tkyi: what is this for?

plugins/auth/index.js
```diff
@@ -158,88 +184,90 @@ exports.register = (server, options, next) => {
...
                     // using function syntax makes 'this' the request
                     const request = this;
-                    const tokenFactory = request.server.app.tokenFactory;
-                    const userFactory = request.server.app.userFactory;
-                    const pipelineFactory = request.server.app.pipelineFactory;
-                    const collectionFactory = request.server.app.collectionFactory;
+                    const { tokenFactory } = request.server.app;
+                    const { userFactory } = request.server.app;
+                    const { pipelineFactory } = request.server.app;
+                    const { collectionFactory } = request.server.app;
```
tkyi: these lines can kind of be combined
but we can do that later

plugins/auth/login.js
```diff
@@ -12,41 +12,39 @@ const uuid = require('uuid/v4');
...
-
-                profile.token = request.server.plugins.auth.generateToken(
-                    profile, config.sessionTimeout
-                );
-                request.cookieAuth.set(profile);
-
-                if (request.params.web === 'web') {
-                    return reply('<script>window.close();</script>');
+    return [
+        {
```
tkyi: seems like an unnecessary newline
is there anything we can change in the rules to handle this issue?

package.json
jithine: we have lots of warnings, 437 all of them are number of function arguments and number of lines in a function. We should strive to refactor current functions and make sure that this number does not increase. Any new code should not fall into this state.

plugins/auth/index.js
jithine: yea, it's not smart enough to combine lines.

plugins/auth/login.js
jithine: it's a prettier rule on items in an array being kept in individual lines. eslint has something similar but `prettier` overrides it to be off https://github.com/prettier/eslint-config-prettier/blob/master/index.js#L24
have to see if this can be tweaked. IMO this is better with multiple object items in array, not need to fiddle around beginning and ending of object paranthesis

jithine: yea this is an intentional decision with prettier https://github.com/prettier/prettier/pull/6694#issuecomment-566898456 since we are adopting `prettier` we have to stuck with it's design philosophies 

https://prettier.io/docs/en/option-philosophy.html

tkyi: okay sounds good
as long as we are aware and okay with it

PR #2032, opened 2020-03-18
feat(1785): Use groupEventId to get parent builds [4]

tkyi:
## Context

Would be nice to use `groupEventId` to get parent builds instead of just the immediate `parentBuild` to determine start logic.

## Objective

This PR:
- uses `groupEventId` to get parentBuilds/finishedBuilds info
    - it gets all events with the same groupEventId, removes downstream builds, filters out builds with the same `jobId` to return the most recent build
- fixes some eslint warnings
- adds `prSource` and `prInfo` data to build create

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/1785

## License


plugins/builds/index.js
```diff
@@ -363,6 +363,8 @@ async function createInternalBuild(config) {
     } = config;
     const event = await eventFactory.get(build.eventId);
     const prRef = event.pr.ref ? event.pr.ref : '';
+    const prSource = event.pr.prSource ? event.pr.prSource : '';
+    const prInfo = event.pr.prInfo ? event.pr.prInfo : '';
```
pritamstyz4ever, line 367:
```suggestion
    const prSource = event.pr.prSource || '';
    const prInfo = event.pr.prInfo || '';
```

plugins/builds/index.js
```diff
@@ -560,15 +564,55 @@ function parseJobInfo({ joinObj, currentJobName, nextJobName, pipelineId, build
...
+            params: {
+                groupEventId: event.groupEventId
+            }
+        });
+        const builds = await event.getBuilds();
+        let parentBuilds = [].concat(builds);
+
+        await Promise.all(
+            parentEvents.map(async pe => {
+                const eventBuilds = await pe.getBuilds();
```
jithine, line 590: can we do a single query here ? something like
```
SELECT * FROM  builds 
     WHERE build.eventId IN (
         SELECT event.id FROM events WHERE event.id = groupEventId
     )
     GROUP BY build.jobId

```
and maybe limit as well ?
else we are looking at multiple db queries equivalent to number of events here, which might not be that many in numbers 🤔  exploring if we can offload all this logic to DB

cc @parthasl

jithine: so idea is to add a method in build model to fetch builds  (latest) for jobs for a given groupEventId

tkyi: thinking about using a query like
```
SELECT * FROM (
    SELECT *, RANK() OVER ( PARTITION BY "jobId" ORDER BY "id" DESC ) AS rank
    FROM public.builds WHERE "eventId" IN 
        (SELECT "id" FROM public.events WHERE "groupEventId" = '12345')) AS foo
WHERE rank = 1
```
This should return an _array of builds_ consisting of the latest build for each job that ran in an event that has a matching `groupEventId` 12345

tkyi: https://github.com/screwdriver-cd/models/pull/450

Issue #1847, opened 2019-11-18
[Feature Request] Pipeline token should be able to create Secrets.


kumada626:
**What happened**:
Currently, create Secrets API is only allowed for User's token.
Since secrets are bound to pipelines, it seems better to be able to create secrets for pipeline tokens as well.

**What you expected to happen**:
Allow Pipeline API Token to create Secrets.

**How to reproduce it**:
POST to `/v4/secrets` with Pipeline Token.

jithine:
We should allow pipeline token to also support enable/disable jobs

jithine:
Available with `screwdriver-api@0.5.840`

PR #2034, opened 2020-03-19
fix(1847): Allow pipeline token to remove and update secret

kumada626:
## Context

Currently only user token can delete and update secrets. We should also allow pipeline token for more flexibility.

## Objective

Allow pipeline token to delete and update secrets.

## References

#1847 

## License

### DIFF
--- a/plugins/secrets/index.js
+++ b/plugins/secrets/index.js
@@ -58,6 +58,10 @@ exports.register = (server, options, next) => {
                 });
             }
 
+            if (scope.includes('pipeline') && secret.pipelineId !== credentials.pipelineId) {
+                throw boom.forbidden('Token does not have permission to this secret');
+            }
+
             if (secret.pipelineId !== credentials.pipelineId && secret.pipelineId !== credentials.configPipelineId) {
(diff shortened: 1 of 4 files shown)


plugins/secrets/index.js
```diff
@@ -58,6 +58,10 @@ exports.register = (server, options, next) => {
                 });
             }
 
+            if (scope.includes('pipeline') && secret.pipelineId !== credentials.pipelineId) {
```
wahapo: I think the code below is more concise.
```
if (secret.pipelineId !== credentials.pipelineId) {
  if (scope.includes('pipeline') {
    throw boom.forbidden('Token does not have permission to this secret');
  }
  if (secret.pipelineId !== credentials.configPipelineId) {
    throw boom.forbidden(`${username} is not allowed to access this secret`);
  }
}
```

kumada626: I think the above code is a little bit confusing to read. In L44, this method checks if token's scope is user or not. Therefore, it is natural to check if token's scope is pipeline or not first.

wahapo: I see, I think that's fine.

PR #2042, opened 2020-03-24
fix: When external downstream trigger is part of a join, should not be triggered by external event

tkyi:
## Context

When user has a pipeline like below, an extra event is getting triggered.
Pipeline 52:
Screen Shot 2020-03-23 at 9 15 47 PM
Pipeline 53:
Screen Shot 2020-03-23 at 9 14 06 PM
Pipeline 54:
Screen Shot 2020-03-23 at 9 14 01 PM

## Objective

This PR does a check to make sure the issue above does not happen by getting the workflowGraph and checking if the downstream external trigger has a join.

## References

Related #1710 

## License

### DIFF
--- a/plugins/builds/index.js
+++ b/plugins/builds/index.js
@@ -909,6 +909,70 @@ async function createOrRunNextBuild({
 }
 
 /**
+ * Finds unique pipeline IDs and filters them out to return duplicates
+ * @param  {Array} externalJobPipelineIds External job pipeline IDs
+ * @return {Array}                        Duplicate external job pipeline IDs
+ */
+function getDuplicatePipelineIds(externalJobPipelineIds) {
+    // Find uniq pipelineIds
+    const uniqPipelineIds = externalJobPipelineIds
+        .map(pid => ({ count: 1, pid }))
+        .reduce((a, b) => {
+            a[b.pid] = (a[b.pid] || 0) + b.count;
+
+            return a;
+        }, {});
+    const duplicatePipelineIds = Object.keys(uniqPipelineIds).filter(a => uniqPipelineIds[a] > 1);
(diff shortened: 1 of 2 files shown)


plugins/builds/index.js
```diff
@@ -936,6 +939,31 @@ async function handleDuplicatePipelines(config) {
...
+    await Promise.all(
+        duplicatePipelineIds.map(async id => {
+            const duplicateJobNames = externalJobNamesWithNoJoinArr.filter(jName => jName.indexOf(`sd@${id}:`) >= 0);
+            const pipeline = await pipelineFactory.get(id);
+
+            if (pipeline && pipeline.workflowGraph) {
+                const { workflowGraph } = pipeline;
+                // Check for join in workflowGraph
+                const containsJoin = duplicateJobNames.some(name => {
+                    const jobName = name.split(':')[1];
```
pritamstyz4ever: do we need length check here?

plugins/builds/index.js
```diff
@@ -949,9 +977,9 @@ async function handleDuplicatePipelines(config) {
 
     // Handle external events
     // If no join array and external and pipeline the same, should be same event
-    if (duplicatePipelineIds.length) {
+    if (pipelinesToStart.length) {
```
pritamstyz4ever: should this be (pipelinesToStart.length > 0)

tkyi: I think both work; if `pipelinesToStart.length` evaluates as anything other than `0`, it will return `true`; otherwise will return `false`
I can add the check if it's more readable/clear that way

pritamstyz4ever: Cool, this works

Issue #1922, opened 2020-01-22
Improve on-boarding experience for new users


jithine:
**What happened**:

New Screwdriver users find it difficult to understand how to create a project and how to create a `screwdriver.yaml` for using supported templates. 

**What you expected to happen**:
Provide step by step instructions to users on creating a pipeline. Support auto generating `screwdriver.yaml` for supported templates.

**UI mocks**

![unnamed](https://user-images.githubusercontent.com/193126/72927590-6cf2af80-3d0b-11ea-97a7-9430aa763eff.png)
![unnamed(1)](https://user-images.githubusercontent.com/193126/72927591-6cf2af80-3d0b-11ea-8576-cfbecf686991.png)
![unnamed(2)](https://user-images.githubusercontent.com/193126/72927592-6cf2af80-3d0b-11ea-89af-f9712639fb34.png)

Screwdriver cluster admins should have flexibility to customize the instruction text and provide supported template list.

**How to reproduce it**:
N/A

PR #2075, opened 2020-04-17
feat(1922): add openPr endpoint

klu909:
## Context

Sd onboarding feature requires an api to open pull request

## Objective

Add a endpoint openPr for opening pull request to support sd onboarding feature

## References

https://github.com/screwdriver-cd/screwdriver/issues/1922

## License


plugins/pipelines/openPr.js
```diff
@@ -0,0 +1,66 @@
+'use strict';
+
+const boom = require('boom');
+const joi = require('joi');
+
+module.exports = () => ({
+    method: 'POST',
+    path: '/pipelines/openPr',
```
tkyi, line 8:
```suggestion
    path: '/pipelines/{id}/openpr',
```
not sure if camel case makes sense here

klu909: we don't really need `id` here because we don't use it

tkyi: you could take in ID instead of scmUri

adong:
Related PRs:
https://github.com/screwdriver-cd/scm-base/pull/77
https://github.com/screwdriver-cd/scm-github/pull/156

plugins/pipelines/openPr.js
klu909: @adong what you think about this?

plugins/pipelines/openPr.js
```diff
@@ -0,0 +1,84 @@
...
+            scope: ['user', '!guest']
+        },
+        plugins: {
+            'hapi-swagger': {
+                security: [{ token: [] }]
+            }
+        },
+        handler: (request, reply) => {
+            const { userFactory } = request.server.app;
+            const { username } = request.auth.credentials;
```
tkyi, line 30: can combine this line and line below

plugins/pipelines/openPr.js
```diff
@@ -0,0 +1,84 @@
...
+                        })
+                        .then(token => userFactory.scm.openPr({
+                            checkoutUrl,
+                            files,
+                            token,
+                            scmContext,
+                            title,
+                            message
+                        }))
+                        .then(pullrequest => reply(pullrequest.url).code(201));
```
tkyi, line 65: camel case might be better
```suggestion
                        .then(pullRequest => reply(pullRequest.url).code(201));
```

tkyi:
- missing README
- missing tests
- need to pull in latest scm-github and scm-router

plugins/pipelines/openPr.js
```diff
@@ -0,0 +1,89 @@
...
+                            })
+                        )
+                        .then(scmUri => user.getPermissions(scmUri))
+                        .then(permissions => {
+                            if (!permissions.push) {
+                                throw boom.forbidden(`User ${username} does not have push access for this repo`);
+                            }
+                        })
+                        .then(token =>
+                            userFactory.scm.openPr({
```
tkyi, line 57: do you need to implement something like this in scm-gitlab? https://github.com/screwdriver-cd/scm-gitlab/blob/master/index.js#L748-L761

klu909: yeah we can do that

klu909: https://github.com/screwdriver-cd/scm-gitlab/pull/28

plugins/pipelines/openPr.js
```diff
@@ -0,0 +1,89 @@
...
+                                checkoutUrl,
+                                files,
+                                token,
+                                scmContext,
+                                title,
+                                message
+                            })
+                        )
+                        .then(pullRequest => reply(pullRequest.data.url).code(201));
```
tkyi, line 66: if scm-gitlab returns `null`, we need to account for that here, right?

klu909: added null check and test for gitlab case

plugins/pipelines/openPr.js
```diff
@@ -0,0 +1,95 @@
...
+                                files,
+                                token,
+                                scmContext,
+                                title,
+                                message
+                            })
+                        )
+                        .then(pullRequest => {
+                            if (!pullRequest) {
+                                throw boom.notImplemented('openPr not implemented for gitlab');
```
tkyi, line 68: so this will always throw an error for scm-gitlab?

jithine: from what I see, scm-gitlab resolves it with no action https://github.com/screwdriver-cd/scm-gitlab/blob/master/index.js#L888
while scm-base rejects with error https://github.com/screwdriver-cd/scm-base/blob/master/index.js#L584

so this should be okay. (though we didn't need that change in scm-gitlab and let scm-base throw error.)

plugins/pipelines/openPr.js
jithine: I would prefer having `id` here just for logging purposes and having a consistent API URLs, 
Though I agree we don't need it for creating the PR.

jithine: else it will de difficult when debugging to trace which URL corresponds to which pipeline creation

plugins/pipelines/openPr.js
klu909: yeah just as Jithin said, this will always throw 501 for scm-gitlab because scm-gitlab returns null.

tkyi: I think I'm confused why opening PR should cause an error when it's not implemented for scm-gitlab
is it a critical operation?

jithine: that is the correct code https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501
It's up to the caller (UI) to display a nice message when scm-gitlab or bitbucket is where where this feature is not supported. 

It's not a critical feature, UI should handle this case when PR cannot be created and display appropriate message to user.

tkyi: yeah I wasn't confused about the code, just how the experience would be for the user
but if the UI is going to show a better message, then that's fine

klu909: I think it make sense to return an error, because UI is expecting api to return url when pull request was created. It's up to UI to handle that error properly

PR #2079, opened 2020-04-24
fix: Add additional logic for getting jobId for trigger

tkyi:
## Context

Currently, we're seeing errors in our logs trying to get jobId info from the workflowGraph during triggerNextJobs function. This is because sometimes the node name saved in the graph is the full external trigger name, not just the shortened job name.

## Objective

This PR
- adds additional logic to get the jobId information when triggering the next job; this should fix an issue where external pipelines with circular workflows don't trigger the later build
- adds try/catch logic around handling triggering external jobs with matching pipelines
- gets eventId info from the build itself for parentBuilds rather than just using the current build's event info

### Example
With three pipelines: 2, 3, and 4, imagine an overall workflow that works like this:
2 -> 3 -> 4 -> 2

Before: The workflow would work like this 2 -> 3 -> 4. The "tag" job should be triggered by pipeline 4, but is never triggered. Below is a screenshot of pipeline 2.
Screen Shot 2020-04-23 at 8 00 04 PM
After: The workflow goes like this 2 -> 3 -> 4 -> 2. The "tag" job is correctly triggered by pipeline 4. Below is a screenshot of pipeline 2.
Screen Shot 2020-04-23 at 8 00 13 PM

## License

### DIFF
--- a/plugins/builds/index.js
+++ b/plugins/builds/index.js
@@ -1059,7 +1059,7 @@ async function handleDuplicatePipelines(config) {
                         parentEventId: event.id
                     });
                 } catch (err) {
-                    logger.info(
+                    logger.error(
                         `Could not create external build - pipeline:${pid} startFrom:~sd@${pipelineId}:${currentJobName} `,
                         err
                     );
@@ -1250,7 +1250,7 @@ exports.register = (server, options, next) => {
                     const nextNode = parentWorkflowGraph.nodes.find(
                         node => node.name === trimJobName(externalJobName) || node.name.includes(nextJobName)
                     );
-                    const jobId = nextNode.id;
+                    const jobId = nextNode ? nextNode.id : null;
                     // Get next build
                     const nextBuild = finishedExternalBuilds.find(b => b.jobId === jobId && b.status === 'CREATED');
                     // The next build has been restarted and this was the original run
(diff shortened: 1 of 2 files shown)


plugins/builds/index.js
```diff
@@ -1229,7 +1247,10 @@ exports.register = (server, options, next) => {
                     const externalPipeline = await pipelineFactory.get(externalEvent.pipelineId);
                     const parentWorkflowGraph = externalEvent.workflowGraph;
                     const finishedExternalBuilds = await externalEvent.getBuilds();
-                    const jobId = parentWorkflowGraph.nodes.find(node => node.name === trimJobName(externalJobName)).id;
+                    const nextNode = parentWorkflowGraph.nodes.find(
+                        node => node.name === trimJobName(externalJobName) || node.name.includes(nextJobName)
+                    );
+                    const jobId = nextNode.id;
```
jithine, line 1253: should we check if `nextNode` exists ?

plugins/builds/index.js
```diff
@@ -1035,17 +1046,24 @@ async function handleDuplicatePipelines(config) {
...
+                        eventFactory,
+                        externalPipelineId: pid,
+                        startFrom: `~sd@${pipelineId}:${currentJobName}`,
+                        parentBuildId: build.id,
+                        parentBuilds,
+                        causeMessage: `Triggered by sd@${pipelineId}:${currentJobName}`,
+                        parentEventId: event.id
+                    });
+                } catch (err) {
+                    logger.info(
```
jithine, line 1062: Can you make this `error` log level..

plugins/builds/index.js
```diff
@@ -1263,9 +1283,10 @@ exports.register = (server, options, next) => {
                                             node.name.includes(`sd@${pid}:${jName}`)
                                         ).id;
                                     }
-                                    parentBuilds[pid].jobs[jName] = finishedInternalBuilds.find(
-                                        b => b.jobId === joinJobId
-                                    ).id;
+                                    const targetBuild = finishedInternalBuilds.find(b => b.jobId === joinJobId);
+
```
jithine, line 1287: should we check if target exists ?

plugins/builds/index.js
tkyi: sure

plugins/builds/index.js
tkyi: updated

plugins/builds/index.js
tkyi: added check

Issue #2050, opened 2020-04-01
Re-considering local config in  sd-local


kkokufud:
**What happened**:
We had feedback for [local config](https://github.com/screwdriver-cd/screwdriver/blob/master/design/sd-local.md#options-1) in sd-local.
- It's hard for users to understand the use case.
  - Originally, we thought that `local` option is needed in the environment which has some clusters. 
- The name of `local` option is not good because it's the same as `sd-local`.

**What you expected to happen**:
We would like to re-consider the design of a new feature instead of `local` config.

kkokufud:
## design

In order to support multiple clusters, we propose a concept like kubectl context.
To do this, we need to change the config file as follows.

### config file
```
configs:
  default:
    api-url: 
    store-url: 
    token: 
    launcher:
      version: latest
      image: screwdrivercd/launcher
  abc:
    api-url: https://api-cd.sample.com
    store-url: https://store-cd.sample.com
    token: hoge
    launcher:
      version: latest
      image: screwdrivercd/launcher
  def:
    api-url: https://api-cd.sandbox.com
    store-url: https://store-cd.sandbox.com
    token: fuga
    launcher:
      version: latest
      image: screwdrivercd/launcher
current: abc
```
- `configs` is a map. The key is name of config, the value is specific config.
- `current` is current config name.
- `default` config is created  automatically.

### command usage
- To create a new config,  we use `sd-local config create ghe`.
- To change a current config, we use `sd-local config use abc`
- To set a config parameter, we use `sd-local config set` as before.
- `sd-local config view` outputs all of config, and `sd-local config view abc` outputs specific config.

yoshwata:
I agree that it is hard to understand what sd-local's `local` option for

sugarnaoming:
I have two questions.
- When is the `default` created?
- What will happen if the name is the same as the existing config name when the `create` command is executed?

kkokufud:
> When is the default created?

This is the timing when .sdlocal/config is created.

> What will happen if the name is the same as the existing config name when the create command is executed?

It returns an error.
If you have any other things, please comment.

sugarnaoming:
> >What will happen if the name is the same as the existing config name when the create command is executed?
>
> It returns an error.

Ok.
I think that's fine with me too.

> > When is the default created?
>
> This is the timing when .sdlocal/config is created.

I think it's redundant to create `default` even if the user runs `sd-local config create hoge` first.

[...]

PR #2080, opened 2020-04-24
fix(2050) update sd-local config design to use multiple cluster

yuichi10:
## Context
Update sd-local config design by screwdriver-cd/screwdriver#2050. 

## Objective
Update configuration that handles multiple cluster configs.

## References
screwdriver-cd/screwdriver#2050

## License

### DIFF
--- a/design/sd-local.md
+++ b/design/sd-local.md
@@ -190,8 +190,7 @@ You can create a cluster config with the command below:
 $ sdlocal config create [config-name]
 ```
 
-note:
-`create` sub command should be error if it is passed the name which is already exists.
+- You can not passed the name which is already exists.
 
 #### use
 You can change which cluster with the command below:
--- a/design/sd-local.md
+++ b/design/sd-local.md
@@ -15,7 +15,7 @@ As a result, User cannot confirm whether the result obtained on CI is the expect
 - March 4th, 2020: Added `src-url` option and updated env options
 - March 16th, 2020: Added `sudo` option
 - March 24th, 2020: Added `--local` option
-- Apr 24th, 2020: Added multiple cluster config and droped `--local` option
+- Apr 24th, 2020: Added multiple cluster config and dropped `--local` option
 
 ## Proposal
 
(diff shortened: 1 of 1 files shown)


design/sd-local.md
```diff
@@ -157,6 +157,46 @@ $ sdlocal build [job-name] [options]
...
+
+- `default` is a special config which are created automatically when the first time the user use sd-local. However, default is not created if the user create a config explicitly with sd-local config create <config-name> in the first time, and, the specified name is used instead of default.
+
+- `current` points to default (or first created config) in the first time
+
+You can create a cluster config with the command below:
+
+```bash
+$ sdlocal config create [config-name]
+```
```
sugarnaoming, line 190: I would be better to add that can't create duplicate cluster names.

design/sd-local.md
```diff
@@ -15,6 +15,7 @@ As a result, User cannot confirm whether the result obtained on CI is the expect
 - March 4th, 2020: Added `src-url` option and updated env options
 - March 16th, 2020: Added `sudo` option
 - March 24th, 2020: Added `--local` option
```
sakka2, line 17: Is it better to delete this line because we do not use `--local` option anymore?
Or we should add some easy explanation that `--local` option has been changed to multiple cluster configs.

design/sd-local.md
```diff
@@ -171,26 +211,34 @@ The chart below shows relationship between `key` and `value`.
...
+      launcher:
+        version: latest
+        image: screwdrivercd/launcher
+   cluster2:
+     api-url: https://cluster2-another-api-screwdriver.com
+     store-url: https://cluster2-another-store-screwdriver.com
+     token: fuga
+     launcher:
+       version: latest
+       image: screwdrivercd/launcher
```
kkisic, line 231: I think it looks like the indentations are not unified.

design/sd-local.md
sugarnaoming: I think it's better to add a message with the "--local" option deleted. because this section is the change log.

s-yoshika: fixed

design/sd-local.md
s-yoshika: added

design/sd-local.md
s-yoshika: fixed

Issue #2087, opened 2020-05-12
Cache cleanup is broken when disk based cache is used


jithine:
**What happened**:

Cache cleanup buttons in Pipeline settings page, does not clear the cache when using disk cache. 

**What you expected to happen**:

Cache cleanup should work no matter which cache strategy is used. 

**How to reproduce it**:

Configure Screwdriver cluster to use disk based cache.

parthasl:
Approach:

Pipeline Options -> Clear Cache

UI -> api
   - check multi-build cluster enabled & cache strategy is disk
   - along with full build config message, add message property type 
            1. pipelines clear cache - {"resource":"caches","type":"invalidate","entity":"pipelines"} or 
            2. jobs clear cache - {"resource":"caches","type":"invalidate","entity":"jobs"})
   - push message to all queues (even if the cluster is disabled). queues will be configured as part of multi-build cluster.

consumer (queue-consumer)
   - mount /cache folder queue-consumer pod, if cache strategy is disk (cluster admins responsibility)
   - check message property type
          - if not available, regular flow (start or stop build)
          - if pipelines clear cache
                 - get pipeline id from full build config and remove the respective folder from the /cache/pipelines  path
          - if jobs clear cache
                 - get job id from full build config and remove the respective folder from the /cache/jobs  path

jithine:
API & UI changes 

1. UI to call new clear cache API
2. Clear cache API will forward requests to store (existing flow) if cache strategy is S3.
3. Clear cache API will forward request to queue-service if cache strategy is disk.

cc @pritamstyz4ever

PR #2093, opened 2020-05-21
feat(2087): Clear Cache implementation

pritamstyz4ever:
## Context
Cache cleanup buttons in Pipeline settings page, does not clear the cache when using disk cache.

## Objective
This PR implements a cache clear api for s3 or disk strategy and routes accordingly to store or queue service for further processing.

## References
https://github.com/screwdriver-cd/screwdriver/issues/2087

## License


plugins/caches/request.js
```diff
@@ -0,0 +1,86 @@
...
+ * @param {String} request.method
+ * @param {String} request.path
+ * @param {Object} request.payload
+ * @param {Object} request.headers
+ */
+async function invoke(request) {
+    const { method, path, payload, auth, params } = request;
+    const { store, queue, cache } = ecosystem;
+
+    const apiPath = path.replace(/\/v4/, '/v1');
```
DekusDenial, line 30: would be nice if we can get it from API config somewhere

plugins/caches/request.js
```diff
@@ -0,0 +1,86 @@
...
+    const options = {
+        json: true,
+        method,
+        uri: `${store}${apiPath}`,
+        headers: {
+            Authorization: `Bearer ${auth.token}`,
+            'Content-Type': 'application/json'
```
DekusDenial, line 37: u dont need this if `json: true` right?

plugins/caches/delete.js
```diff
@@ -0,0 +1,27 @@
...
+            strategies: ['token'],
+            scope: ['user', 'pipeline', '!guest']
+        },
+        plugins: {
+            'hapi-swagger': {
+                security: [{ token: [] }]
+            }
+        },
+        handler: async (request, reply) => {
+            const res = await api.invoke(request);
```
DekusDenial, line 22: if the request fails, even after retries, will this handler handle that error and pass error message back to the UI to display?

pritamstyz4ever: Pass, the message back to the UI. Also, I think the UI should show whether cache was cleared or not. Currently it doesn't do anything

plugins/stats.js
```diff
@@ -24,11 +24,15 @@ exports.register = (server, options, next) => {
                 schema: schema.api.stats
             }
         },
-        handler: (request, reply) =>
+        handler: async (request, reply) => {
+            const executorStatsresponse = await executor.stats({ token: '' });
+            const executorStats = executorStatsresponse.body;
```
DekusDenial, line 29: what is this change for?

plugins/caches/delete.js
```diff
@@ -0,0 +1,27 @@
...
+
+            return reply(res).code(res.statusCode);
+        }
+    }
+});
```
tkyi, line 27: can you add param validation?

plugins/stats.js
pritamstyz4ever: stats api was broken after the queue service change, this change should give the correct stats result of executor and scm

DekusDenial: oh so this is separate

jithine: does this works in a setup where `queue-service` is not used ?

plugins/pipelines/caches/request.js
```diff
@@ -0,0 +1,94 @@
...
+ * @param {Object} request.headers
+ */
+async function invoke(request) {
+    const { method, payload, query, auth, params } = request;
+    const pipelineId = params.id;
+    const { store, queue, cache } = request.server.app.ecosystem;
+    const { scope, cacheId } = query;
+    const { username, scmContext } = auth.credentials;
+
+    console.log(username);
```
jithine, line 31: remove console.logs

PR #2112, opened 2020-06-10
fix: openPR return object instead of plain string

adong:
## Context

UI expects return value to be a String that can be parsed via `JSON.parse`, if it is plain String like `thisisaStingValue`, then it will throw errors.

## Objective
Change return value from `String` to `Object`, so can be parsed by `JSON.parse`

## License

### DIFF
--- a/plugins/pipelines/openPr.js
+++ b/plugins/pipelines/openPr.js
@@ -72,7 +72,7 @@ module.exports = () => ({
                                 throw boom.notImplemented('openPr not implemented for gitlab');
                             }
 
-                            return reply(pullRequest.data.url).code(201);
+                            return reply({ prUrl: pullRequest.data.url }).code(201);
                         });
                 })
                 .catch(err => reply(boom.boomify(err)));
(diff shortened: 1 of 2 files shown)


plugins/pipelines/openPr.js
```diff
@@ -72,7 +72,7 @@ module.exports = () => ({
                                 throw boom.notImplemented('openPr not implemented for gitlab');
                             }
 
-                            return reply(pullRequest.data.url).code(201);
+                            return reply({ prUrl: pullRequest.data.html_url }).code(201);
```
adong, line 75: We need to use `html_url` instead of `url` from API response, see reference here: 
`https://developer.github.com/v3/pulls/#create-a-pull-request`

The difference between `url` and `html_url` is that
`url`: `https://api.github.com/repos/adong/sd-open-pr-w-branch/pulls/14`
`html_url`: `https://www.github.com/adong/sd-open-pr-w-branch/pulls/14`

PR #2118, opened 2020-06-12
fix: Use previous build ID as parentBuildId for restart event case

tkyi:
## Context

Currently, when a user restarts an event, the event's `parentBuildId` is passed as the new build's `parentBuildId`. This will cause the new event to have the wrong metadata.

## Objective

This PR passes in the correct information (the previous build's id) as the restarted event's `parentBuildId`.

## License


jithine:
Let's not do this, since this will change current behavior of restoring a build's meta to it's initial state.

Issue #2128, opened 2020-06-23
Github releases are breaking pipelines which doesn't have a release trigger


jithine:
**What happened**:
Pipeline event calls are failing on a pipeline resulting in 404 page. There is an invalid entry in events which contains `startFrom` as `release:`. This happened when user created a Github Release with out a name while the pipeline doesn't have any release/tag triggers. 

**What you expected to happen**:

If a pipeline doesn't have any tag/release trigger, there should not be any invalid events created.

**How to reproduce it**:
1. Create a release without a name, but just tag name.
2. Pipeline event gets creates with `startFrom` as `relase:`
2. Events API starts returning 500
3. Pipeline details page becomes 404 page

jithine:
cc @kumada626 @ibu1224 @wahapo

ibu1224:
@jithine 
The following pipeline was unable to reproduce the problem.
https://cd.screwdriver.cd/pipelines/4471/events

However, several pipelines within YJ are experiencing similar problems.

PR #2129, opened 2020-06-23
fix(2128): ignore empty release/tag trigger names

jithine:
## Context

https://github.com/screwdriver-cd/screwdriver/issues/2128

Pipelines are breaking when Github release is created without a name and if pipeline doesn't have a release trigger.

Release names can be empty string, but tag names must be present

## License


tkyi:
Seeing error in tests:
```
00:07:57 {"message":"The releaseName of SCM Webhook is missing.","level":"error","timestamp":"2020-06-24T00:15:12.726Z"}
00:07:57 Debug: internal, implementation, error 
00:07:57     AssertionError: expected '' to equal '~release'
```

jithine:
cc @ibu1224 @wahapo @kumada626

PR #2157, opened 2020-07-20
feat(1079): integrate deploy keys features with auth and pipeline routes [5]

supra08:
## Context

Fixes screwdriver-cd/screwdriver#1079

## Objective

This PR adds functionality of handling of autoKeysGeneration visibility in the UI and its response to trigger it in the scm-github module. 

## References

screwdriver-cd/screwdriver#1079

## License

### DIFF
--- a/plugins/auth/contexts.js
+++ b/plugins/auth/contexts.js
@@ -16,28 +16,42 @@ module.exports = config => ({
         description: 'Get all auth contexts',
         notes: 'Get all auth contexts',
         tags: ['api', 'auth', 'context'],
-        handler: (request, reply) => {
+        handler: async (request, reply) => {
             const { scm } = request.server.app.userFactory;
+            const { pipelineFactory } = request.server.app;
             const scmContexts = scm.getScmContexts();
             const contexts = [];
+            const promises = [];
 
             scmContexts.forEach(scmContext => {
-                const context = {
-                    context: scmContext,
-                    displayName: scm.getDisplayName({ scmContext })
-                };
+                const promise = pipelineFactory.scm.checkAutoDeployKeyGeneration({
+                    scmContext
+                });
 
-                contexts.push(context);
+                promises.push(promise);
             });
 
-            if (config.allowGuestAccess) {
-                contexts.push({
-                    context: 'guest',
-                    displayName: 'Guest Access'
-                });
-            }
+            return Promise.all(promises).then(async autoDeployKeyGenerationList => {
+                scmContexts.forEach((scmContext, i) => {
+                    const context = {
+                        context: scmContext,
+                        displayName: scm.getDisplayName({ scmContext }),
+                        autoDeployKeyGeneration: autoDeployKeyGenerationList[i]
+                    };
 
-            return reply(contexts);
+                    contexts.push(context);
+                });
+
+                if (config.allowGuestAccess) {
+                    contexts.push({
+                        context: 'guest',
+                        displayName: 'Guest Access',
+                        autoDeployKeyGeneration: false
+                    });
+                }
+
+                return reply(contexts);
+            });
         },
         response: {
             schema: schema.api.auth.contexts
--- a/plugins/auth/contexts.js
+++ b/plugins/auth/contexts.js
@@ -17,15 +17,15 @@ module.exports = config => ({
         notes: 'Get all auth contexts',
         tags: ['api', 'auth', 'context'],
         handler: (request, reply) => {
-            const { pipelineFactory, userFactory } = request.server.app;
-            const scmContexts = userFactory.scm.getScmContexts();
+            const { scm } = request.server.app.userFactory;
+            const scmContexts = scm.getScmContexts();
             const contexts = [];
 
             scmContexts.forEach(scmContext => {
                 const context = {
                     context: scmContext,
-                    displayName: userFactory.scm.getDisplayName({ scmContext }),
-                    autoDeployKeyGeneration: pipelineFactory.scm.autoDeployKeyGenerationEnabled({
+                    displayName: scm.getDisplayName({ scmContext }),
+                    autoDeployKeyGeneration: scm.autoDeployKeyGenerationEnabled({
                         scmContext
                     })
                 };
--- a/plugins/auth/contexts.js
+++ b/plugins/auth/contexts.js
@@ -24,7 +24,7 @@ module.exports = config => ({
             const promises = [];
 
             scmContexts.forEach(scmContext => {
-                const promise = pipelineFactory.scm.checkAutoDeployKeyGeneration({
+                const promise = pipelineFactory.scm.autoDeployKeyGenerationEnabled({
                     scmContext
                 });
 
--- a/plugins/auth/contexts.js
+++ b/plugins/auth/contexts.js
@@ -16,42 +16,32 @@ module.exports = config => ({
         description: 'Get all auth contexts',
         notes: 'Get all auth contexts',
         tags: ['api', 'auth', 'context'],
-        handler: async (request, reply) => {
-            const { scm } = request.server.app.userFactory;
-            const { pipelineFactory } = request.server.app;
-            const scmContexts = scm.getScmContexts();
+        handler: (request, reply) => {
+            const { pipelineFactory, userFactory } = request.server.app;
+            const scmContexts = userFactory.scm.getScmContexts();
             const contexts = [];
-            const promises = [];
 
             scmContexts.forEach(scmContext => {
-                const promise = pipelineFactory.scm.autoDeployKeyGenerationEnabled({
-                    scmContext
-                });
+                const context = {
+                    context: scmContext,
+                    displayName: userFactory.scm.getDisplayName({ scmContext }),
+                    autoDeployKeyGeneration: pipelineFactory.scm.autoDeployKeyGenerationEnabled({
+                        scmContext
+                    })
+                };
 
-                promises.push(promise);
+                contexts.push(context);
             });
 
-            return Promise.all(promises).then(async autoDeployKeyGenerationList => {
-                scmContexts.forEach((scmContext, i) => {
-                    const context = {
-                        context: scmContext,
-                        displayName: scm.getDisplayName({ scmContext }),
-                        autoDeployKeyGeneration: autoDeployKeyGenerationList[i]
-                    };
-
-                    contexts.push(context);
+            if (config.allowGuestAccess) {
+                contexts.push({
+                    context: 'guest',
+                    displayName: 'Guest Access',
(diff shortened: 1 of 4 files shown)


plugins/pipelines/create.js
```diff
@@ -24,9 +24,13 @@ module.exports = () => ({
         handler: (request, reply) => {
             const checkoutUrl = helper.formatCheckoutUrl(request.payload.checkoutUrl);
             const rootDir = helper.sanitizeRootDir(request.payload.rootDir);
+            const { autoKeysGeneration } = request.payload;
+            const { secretFactory } = request.server.app;
             const { pipelineFactory, userFactory, collectionFactory } = request.server.app;
             const { username } = request.auth.credentials;
             const { scmContext } = request.auth.credentials;
```
tkyi, line 31:
```suggestion
            const { pipelineFactory, userFactory, collectionFactory, secretFactory } = request.server.app;
            const { username, scmContext } = request.auth.credentials;
```

plugins/pipelines/create.js
```diff
@@ -35,7 +37,13 @@ module.exports = () => ({
                     .then(user =>
                         user
                             .unsealToken()
+                            .then(token => {
+                                pipelineToken = token;
+
+                                return token;
+                            })
                             .then(token =>
+                                // pipelineToken = token;
```
tkyi, line 46: commented code

plugins/pipelines/create.js
```diff
@@ -78,7 +86,7 @@ module.exports = () => ({
                                         return pipelineFactory.create(pipelineConfig);
                                     })
                                     // get the default collection for current user
-                                    .then(pipeline =>
+                                    .then(pipeline => {
                                         collectionFactory
```
tkyi, line 90: if you're going to add the brace (`{`) in the line above, it'd be safer to use `return` here
```suggestion
                                        return collectionFactory
```

plugins/pipelines/create.js
```diff
@@ -113,6 +121,41 @@ module.exports = () => ({
 
                                                 return defaultCollection.update();
                                             })
+                                            .then(() => {
+                                                if (autoKeysGeneration) {
+                                                    pipelineFactory.scm
```
tkyi, line 126:
```suggestion
                                                    return pipelineFactory.scm
```

plugins/pipelines/create.js
```diff
@@ -113,6 +121,41 @@ module.exports = () => ({
...
+                                                        .addDeployKey({
+                                                            scmContext,
+                                                            checkoutUrl,
+                                                            token: pipelineToken
+                                                        })
+                                                        .then(privateDepKey => {
+                                                            secretFactory
```
tkyi, line 133:
```suggestion
                                                            return secretFactory
```

test/plugins/pipelines.test.js
```diff
@@ -137,6 +137,16 @@ const getSecretsMocks = secrets => {
     return decorateJobMock(secrets);
 };
 
+// const getSecretMock = secret => {
```
tkyi, line 140: commented code

plugins/auth/contexts.js
```diff
@@ -16,28 +16,42 @@ module.exports = config => ({
...
-            if (config.allowGuestAccess) {
-                contexts.push({
-                    context: 'guest',
-                    displayName: 'Guest Access'
+            return Promise.all(promises).then(async autoDeployKeyGenerationList => {
+                scmContexts.forEach((scmContext, i) => {
+                    const context = {
+                        context: scmContext,
+                        displayName: scm.getDisplayName({ scmContext }),
+                        autoDeployKeyGeneration: autoDeployKeyGenerationList[i]
```
jithine, line 39: why is this not similar to `scm.getDisplayName` ? shouldn't this be `scm.autoDeployKeyGenerationEnabled({ scmContext });`

jithine: you can avoid all those promise wait and all. It's not really a Promise, data is there in the config.  you can give it same treatment as `getDisplayName`

test/plugins/auth.test.js
```diff
@@ -1058,13 +1060,14 @@ describe('auth plugin test', () => {
                     url: '/auth/contexts'
                 })
                 .then(reply => {
+                    console.log(reply.result[0]);
```
jithine, line 1063: remove log

plugins/pipelines/create.js
```diff
@@ -24,9 +24,11 @@ module.exports = () => ({
...
             const checkoutUrl = helper.formatCheckoutUrl(request.payload.checkoutUrl);
             const rootDir = helper.sanitizeRootDir(request.payload.rootDir);
-            const { pipelineFactory, userFactory, collectionFactory } = request.server.app;
-            const { username } = request.auth.credentials;
-            const { scmContext } = request.auth.credentials;
+            const { autoKeysGeneration } = request.payload;
+            const { pipelineFactory, userFactory, collectionFactory, secretFactory } = request.server.app;
+            const { username, scmContext } = request.auth.credentials;
+            let pipelineToken = '';
+            const depKeySecret = 'SD_SCM_DEPLOY_KEY';
```
jithine, line 31: call it `deployKeySecret`

plugins/pipelines/create.js
jithine: yet to remove this

plugins/pipelines/create.js
```diff
@@ -113,6 +121,43 @@ module.exports = () => ({
...
                                             })
+                                            .then(() => {
+                                                if (autoKeysGeneration) {
+                                                    return pipelineFactory.scm
+                                                        .addDeployKey({
+                                                            scmContext,
+                                                            checkoutUrl,
+                                                            token: pipelineToken
+                                                        })
+                                                        .then(privateDepKey => {
```
jithine, line 132: should add error handling. scm's which have not implemented `_addDeployKey` throws error. 

catch log and respond error part of API payload, so that UI can display.

plugins/pipelines/create.js
```diff
@@ -113,6 +121,43 @@ module.exports = () => ({
...
+                                                            return secretFactory
+                                                                .get({
+                                                                    pipelineId: pipeline.id,
+                                                                    name: depKeySecret
+                                                                })
+                                                                .then(secret => {
+                                                                    if (secret) {
+                                                                        throw boom.conflict(
```
jithine, line 140: is this feasible ? isn't this a new pipeline creation ?

plugins/auth/contexts.js
```diff
@@ -17,14 +17,17 @@ module.exports = config => ({
...
+            const scmContexts = userFactory.scm.getScmContexts();
             const contexts = [];
 
             scmContexts.forEach(scmContext => {
                 const context = {
                     context: scmContext,
-                    displayName: scm.getDisplayName({ scmContext })
+                    displayName: userFactory.scm.getDisplayName({ scmContext }),
+                    autoDeployKeyGeneration: pipelineFactory.scm.autoDeployKeyGenerationEnabled({
+                        scmContext
```
jithine, line 29: is pipelineFactory.scm & userFactory.scm referring to same ?

supra08: Yeah I guess I can use userFactory for both.

plugins/pipelines/create.js
```diff
@@ -113,6 +120,36 @@ module.exports = () => ({
...
+                                            .then(() => {
+                                                if (autoKeysGeneration) {
+                                                    return pipelineFactory.scm
+                                                        .addDeployKey({
+                                                            scmContext,
+                                                            checkoutUrl,
+                                                            token: pipelineToken
+                                                        })
+                                                        .then(privateDeployKey => {
+                                                            if (!privateDeployKey) {
```
jithine, line 132: is this needed ? won't the `addDeployKey` promise reject ?

plugins/pipelines/create.js
```diff
@@ -113,6 +120,36 @@ module.exports = () => ({
 
                                                 return defaultCollection.update();
                                             })
+                                            .then(() => {
+                                                if (autoKeysGeneration) {
```
adong, line 124: This boolean check if scm, such as `github` has ability to do `autoKeysGeneration`.

What if such scm, say `bitbucket` does not have such feature and `autoKeysGeneration = true` from the payload. How do we handle that case? Would there be: 

1) add a `catch` block
2) add another if condition to check both payload sent from ui has `autoKeysGeneration: true`, as well as check if given SCM has such ability.

In UI, we can prevent user giving input for if scm does not have `autoKeysGeneration` feature, but we also need to sanitize and guard on API. I can see some power users would call this API directly.


Just a thought, what do you think?


Note: 

From: https://github.com/screwdriver-cd/screwdriver/issues/1079#issuecomment-393631839

If we apply this feature to all SCMs: `scm-github`, `scm-bitbucket`, then we dont have such issue, just need to catch the errors.

plugins/pipelines/create.js
supra08: Yeah a catch should have been used. I'll correct it. 
Apart from this, I guess I found another place where it is used wrongly implemented:
https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/pipelines/openPr.js#L71

plugins/pipelines/create.js
supra08: Yeah I get the point. I didn't take the case of direct API calls.
Catch is definitely needed. And it will be handled by the not implemented promise rejection in scm base.
I am wondering if the 2nd point is needed because if `autoKeysGeneration` is set to true in the payload, it will be rejected in the addDeployKey method and eventually sent to the catch block.

adong: Cool. I meant either 1) or 2). 

Since handled that case in 1), 2) is no longer necessary

supra08: alright cool 👍

plugins/pipelines/create.js
```diff
@@ -145,7 +139,8 @@ module.exports = () => ({
                                                                 value: privateDeployKeyB64,
                                                                 allowInPR: false
                                                             });
-                                                        });
+                                                        })
+                                                        .catch(err => reply(boom.boomify(err)));
```
jithine, line 143: won't this still execute rest of the `Promise` chain ?
I thought was this catch is not required, and let the `err` propagate to the final `catch` at https://github.com/screwdriver-cd/screwdriver/pull/2157/commits/423742205e5363e72b09e8ff630d383d27c2d877#diff-1eb6375f360812e82460d42f926123b1R170

plugins/pipelines/create.js
jithine: don't we already have a `catch` at end of Promise chain ?

Issue #2152, opened 2020-07-15
Sonarqube Pull Request flow is broken


jithine:
**What happened**:

Screwdriver creates a new Sonar Project  for each pull request. This is not ideal and does not enable a project to make use of [Pull Request Analysis](https://docs.sonarqube.org/latest/analysis/pull-request/). Uses looses out on a unified Quality Gate Status check against Pipeline branch and Pull Requests. This also prevents Screwdriver from offering [Pull Request Decorations](https://docs.sonarqube.org/latest/analysis/pr-decoration/)

**What you expected to happen**:

1. When a Pull Request Analysis runs, Screwdriver Sonar Coverage book end must make use of [Pull Request Analysis](https://docs.sonarqube.org/latest/analysis/pull-request/). 
1. Screwdriver should not create a new Sonar Project for each Pull Request. We should rather use Pull Request Analysis](https://docs.sonarqube.org/latest/analysis/pull-request/) feature and use analysis data from Pull Request to populate coverage data on UI.

**How to reproduce it**:

Create a Pull Request and observer it's Sonar Project. It's a separate one from the Sonar Project for main repository.

**References**
1. https://docs.sonarqube.org/latest/analysis/pull-request/
1. https://docs.sonarqube.org/latest/analysis/pr-decoration/
1. Implementation references
   1. https://github.com/screwdriver-cd/coverage-sonar/blob/master/commands.txt#L11
   1. https://github.com/screwdriver-cd/coverage-sonar/blob/master/index.js#L215-L226
   1. https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/coverage/token.js#L24-L28

renaud-vmg:
Adding: we're also missing the "source off truth" project key, which is to say, the Parent Job's project key, which is likely known by the engine. We have a local workaround in that we set a certain meta value to tell Screwdriver which project to get the required key from, but it would be easier for integration to have something such as `PR_PARENT_SQ_PROJECT_ID`.

jithine:
Tasks


1. Add SD_SONAR_ENTERPRISE=false (by default)
2. coverage-sonar command to check this variable, and if present and equals true, do the following
   1.  pass projectKey=`SD_PR_PARENT_JOB_ID` in this API call  https://github.com/screwdriver-cd/coverage-sonar/blob/master/commands.txt#L6
[...]

PR #2161, opened 2020-07-25
feat(2152): Support coverage scope param [3]

tkyi:
## Context

It would be nice if users could configure projectKey/name scope based on an annotation. 

## Objective

This PR fetches job annotations in the API so it can easily be passed to the coverage sonar plugin. Alternatively, if `scope` query param is passed in, the API will skip fetching.

## References

Related to #2152 

## License

### DIFF
--- a/plugins/coverage/info.js
+++ b/plugins/coverage/info.js
@@ -19,10 +19,35 @@ module.exports = config => ({
             }
         },
         handler: (request, reply) => {
-            config.coveragePlugin
-                .getInfo(request.query)
-                .then(reply)
-                .catch(err => reply(boom.boomify(err)));
+            const { jobId, pipelineId, startTime, endTime, jobName, pipelineName, scope } = request.query;
+            const { jobFactory } = request.server.app;
+            const infoConfig = { jobId, pipelineId, startTime, endTime, jobName, pipelineName };
+
+            return Promise.resolve()
+                .then(() => {
+                    if (scope) {
+                        return { 'screwdriver.cd/coverageScope': request.query.scope };
+                    }
+                    if (!scope && jobId) {
+                        return jobFactory.get(jobId).then(job => {
+                            if (!job) {
+                                throw boom.notFound('Job does not exist');
+                            }
+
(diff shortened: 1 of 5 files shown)


plugins/coverage/info.js
```diff
@@ -19,10 +19,35 @@ module.exports = config => ({
...
+            const { jobId, pipelineId, startTime, endTime, jobName, pipelineName, scope } = request.query;
+            const { jobFactory } = request.server.app;
+            const infoConfig = { jobId, pipelineId, startTime, endTime, jobName, pipelineName };
+
+            return Promise.resolve()
+                .then(() => {
+                    if (scope) {
+                        return { 'screwdriver.cd/coverageScope': request.query.scope };
+                    }
+                    if (!scope && jobId) {
```
jithine, line 31: hopefully we can retire this end point once build end point fetches coverage info.

plugins/coverage/info.js
```diff
@@ -19,10 +19,35 @@ module.exports = config => ({
...
-                .getInfo(request.query)
-                .then(reply)
-                .catch(err => reply(boom.boomify(err)));
+            const { jobId, pipelineId, startTime, endTime, jobName, pipelineName, scope } = request.query;
+            const { jobFactory } = request.server.app;
+            const infoConfig = { jobId, pipelineId, startTime, endTime, jobName, pipelineName };
+
+            return Promise.resolve()
+                .then(() => {
+                    if (scope) {
```
jithine, line 28: Thinking a bit more on this, do we even need all this ?

build.environment should have https://github.com/screwdriver-cd/coverage-sonar/pull/41/files#diff-168726dbe96b3ce427e7fedce31bb0bcR312 

This end point can simply take that value and avoid calling job model at all.

Only weirdness will be it's called an explicit Sonar Project Key, But that should be okay, in future if we add a new coverage source, we can handle it then.

plugins/coverage/info.js
tkyi: yes, @klu909 should be working on that

PR #2159, opened 2020-07-21
fix: Join acceptance tests

esuc4se:
## Context
We need to acceptance test for remote join & external join.

## Objective
- Add remote join & external join accesptance test
- Refactor trigger.js, world.js

## License
I confirm that this contribution is made under a BSD license and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #2166, opened 2020-07-30
fix: support only beta jobs

esuc4se:
## Context
(#2159) caused failure of beta functional test.
The reason was that functional test didn't work on beta.
So, I solved this problem.

this PR diff
https://github.com/screwdriver-cd/screwdriver/pull/2166/commits/969592a3f9529d826372cf928cb6a6396c5416de

## Objective
- Remote join acceptance test enable work on beta
- Add required jobs in beta
- change test pipeline name (to clearly separate by beta functional test)
  - remote1 -> remote_join1, beta-remote_join1
  - remote2 -> remote_join2, beta-remote_join2
  - remoteA -> external_trigger1
  - remoteB -> external_trigger2
- Add remote join & external join accesptance test (#2159)
- Refactor trigger.js, world.js (#2159)

## References
- functional-trigger test branch (example)
  - https://github.com/screwdriver-cd-test/functional-trigger/tree/beta-remote_join1

## License

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ package-lock.json
 docker-compose.yml
 mw-data
 .vscode
+.envrc
--- a/features/step_definitions/trigger.js
+++ b/features/step_definitions/trigger.js
@@ -2,6 +2,7 @@
 
 const Assert = require('chai').assert;
 const { Before, Given, When, Then } = require('cucumber');
+const github = require('../support/github');
 const sdapi = require('../support/sdapi');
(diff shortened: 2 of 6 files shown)


tk3fftk:
How about running `functional-beta` in PR build when `/features` is updated? We will able to see the update works or not.
Ideally, we should run only an updated test case, but it may be tough to run specific case dynamically.

tkyi:
Would be nice to run functional-beta tests in PR 👍 
Will merge this PR first

PR #2178, opened 2020-08-12
fix(2177): reduce metrics API data output for no query params

jithine:
## Context

1. Reduce Metrics API data output for no query params, to reduce load on API. 
1. Fix for webhook not triggering pipelines when there are other pipelines under different branch with no admin

## License


test/plugins/webhooks.test.js
```diff
@@ -1501,14 +1501,14 @@ describe('webhooks plugin test', () => {
                 });
             });
 
-            it('returns 500 when getCommitSha() is rejected on 5xx error', () => {
+            it('returns 204 when getCommitSha() is rejected on 5xx error', () => {
```
pritamstyz4ever, line 1504: should it return 204 even on a 5xx error?

jithine: so with the 5xx is from gitCommitSha end point, previously webhook used to error out, now it continues. 

What we are trying to solve is this., suppose there are 2 pipelines related to the same git repo, and the first one has error in permissions (can't fetch token etc).  We still want to process the 2nd pipeline and webhook to return success.

Whether we should use a different status code can be tackled later.

Issue #2185, opened 2020-08-19
Use default_branch as default branch


tkyi:
**What happened**: Currently, when branch is not provided upon pipeline creation, we set the branch to `master`.

**What you expected to happen**: Some repositories might not have `master` as the default branch, it would be better to use `default_branch`.

**How to reproduce it**: Go to https://cd.screwdriver.cd/create and use `git@github.com:tkyi/mytest.git` as the checkoutUrl.

PR #2186, opened 2020-08-19
feat(2185): Handle default branch in scm plugin [2]

tkyi:
## Context

Currently, default branch for Screwdriver pipeline is hardcoded to `master`. It would be nice if the default branch actually took in the Github `default_branch`.

## Objective

This PR moves handling of default branch to the scm plugin.

## References

Related to #2185 

## License


plugins/pipelines/helper.js
```diff
@@ -13,18 +13,18 @@ const formatCheckoutUrl = checkoutUrl => {
...
-        .split('#')[0]
-        .toLowerCase()
-        .concat(branchName);
+    if (branchName) {
+        result = result
+            .split('#')[0]
+            .toLowerCase()
+            .concat(branchName);
+    } else {
+        result = result.toLowerCase();
```
adong, line 26:
```
    // Do not convert branch name to lowercase
```

Hmm? 🤔

tkyi: there's no branch name here

Issue #2068, opened 2020-04-15
Large step logs causes browser to hang


jithine:
**What happened**:

UI is loading all the step log data in memory and looping through the data to create a blob to make it downloadable. This does not work for large step log files where there are hundred thousands of log lines.

**What you expected to happen**:

1. UI should not do any custom logic and should defer to API for log download.
1. New API to download step logs with proper caching.
1. UI should remove all custom processing for download step log.

**How to reproduce it**:
1. Create  a large step log file with about 153, 000 lines. 
1. Attempt to download the step log using Download button in log page.

sladyn98:
@jithine I could give this a shot, any tips on how to begin

jithine:
@sladyn98 my apologies, I didn't see your message, now it's a little late I guess 😢

PR #2200, opened 2020-09-01
fix(2068): Large step logs causes browser to hang

klu909:
## Context

current step with huge log will freeze browser when downloading log
## Objective

move download logic to api

## References

https://github.com/screwdriver-cd/screwdriver/issues/2068

## License


plugins/builds/steps/logs.js
```diff
@@ -191,9 +191,19 @@ module.exports = config => ({
                                 maxLines
                             })
                         )
-                        .then(([lines, morePages]) =>
-                            reply(lines).header('X-More-Data', (morePages || !isDone).toString())
-                        );
+                        .then(([lines, morePages]) =>{
+                            if (type !== 'download') {
+                                return reply(lines).header('X-More-Data', (morePages || !isDone).toString());
```
jithine, line 196: we should do this for `lines` in  non-downloads also. UI should not spend time doing any looping.

adong (approved):

PR #2226, opened 2020-09-25
fix: Use template

tkyi:
## Context

Now that we have bumped Screwdriver to a major version, we can use semantic release template to publish new versions.

## Objective

This PR switches to using semantic-release template.

## License

### DIFF
--- a/screwdriver.yaml
+++ b/screwdriver.yaml
@@ -1,3 +1,6 @@
+cache:
+  event: [ $SD_SOURCE_DIR/node_modules ]
+
 shared:
     image: node:12
 
--- a/screwdriver.yaml
+++ b/screwdriver.yaml
@@ -26,14 +26,14 @@ jobs:
 
     publish:
         requires: [main]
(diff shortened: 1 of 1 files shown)


screwdriver.yaml
```diff
@@ -26,14 +26,14 @@ jobs:
...
         requires: [main]
+        template: screwdriver-cd/semantic-release
         steps:
-            - setup-ci: git clone https://github.com/screwdriver-cd/toolbox.git ci
-            - install-ci: npm install npm-auto-version
-            - publish-npm-and-git-tag: ./ci/publish.sh
-            - publish-docker: |
-                export DOCKER_TAG=`cat VERSION`
+          - postpublish: |
+                git clone https://github.com/screwdriver-cd/toolbox.git ci
```
jithine, line 32: can we make this a shared command.

tkyi: just the clone part?

jithine (approved):

Issue #2199, opened 2020-08-28
Build metrics - Display Pipeline downtime information


jithine:
**What happened**:
Pipeline owners need to know how long a pipeline was left in broken state due to build failures

**What you expected to happen**:

Build Metrics shows appropriate graph to denote how long a Pipeline was in broken state due to build failures.
User should be able to exclude jobs from downtime decision, ie failure of certain jobs in the pipeline need not be considered as a downtime.

**UI Mocks**

![unnamed](https://user-images.githubusercontent.com/193126/91511953-b99bfb80-e895-11ea-902f-d70a16e4e8a7.png)

**How to reproduce it**:

N/A

jithine:
cc @adong @chasturansky

PR #2237, opened 2020-10-02
feat(2199): Add endpoints to get and update user settings [3]

tkyi:
## Context

It would be nice if we stored user settings and pipeline settings in the backend.

## Objective

This PR adds endpoints to get and update user settings.
Also allows user to update pipeline settings.

## References

Related to #2199 

## License


plugins/users/settings/get.js
```diff
@@ -0,0 +1,51 @@
...
+        },
+        plugins: {
+            'hapi-swagger': {
+                security: [{ token: [] }]
+            }
+        },
+        handler: async (request, h) => {
+            const { userFactory } = request.server.app;
+
+            return userFactory
```
pritamstyz4ever, line 28: should we use async-await here?

test/lib/registerPlugins.test.js
```diff
@@ -6,7 +6,7 @@ const sinon = require('sinon');
 
 sinon.assert.expose(Assert, { prefix: '' });
 
-describe('Register Unit Test Case', () => {
+describe.only('Register Unit Test Case', () => {
```
pritamstyz4ever, line 9: remove .only?

plugins/users/settings/update.js
```diff
@@ -0,0 +1,58 @@
...
+                        throw boom.forbidden(`User ${username} cannot update user settings for user ${user.username}`);
+                    }
+
+                    return user.updateSettings(settings).then(results => {
+                        return h.response(results).code(200);
+                    });
+                })
+                .catch(err => {
+                    throw err;
+                });
```
pritamstyz4ever, line 47: change to async await, and remove catch err

tkyi: updated

plugins/users/settings/get.js
tkyi: updated

plugins/users/settings/update.js
```diff
@@ -0,0 +1,52 @@
...
+            const { userFactory } = request.server.app;
+            const { username } = request.auth.credentials;
+            const { settings } = request.payload;
+            const user = await userFactory.get(request.params.id);
+
+            if (!user) {
+                throw boom.notFound('User does not exist');
+            }
+
+            if (user.username !== username) {
```
adong, line 35: is username `uuid`?  Curious if two users can have same username?

adong (approved):

Issue #2240, opened 2020-10-08
BuildCluster became enabled even if settings for buildCluster is "false"


kkokufud:
**What happened**:
In order to disable to buildCluster, we set `false` to ` MULTI_BUILD_CLUSTER_ENABLED` in [API](https://github.com/screwdriver-cd/screwdriver/blob/4e9468e90fae0ee95fe24422e156dd78522c1cea/config/custom-environment-variables.yaml#L371)  and `SCHEDULER_ENABLED` in [queue-service](https://github.com/screwdriver-cd/queue-service/blob/03ac0fbbc78931b11ef3959ec406ce4d44764aa3/config/custom-environment-variables.yaml#L256). But buildCluster was not disabled.
When we cleared these settings, it was disabled.
Probably, these should be compared in boolean, not string.

**What you expected to happen**:
BuildCluster becomes disabled if `MULTI_BUILD_CLUSTER_ENABLED` and `SCHEDULER_ENABLED` is false.

**How to reproduce it**:
- Set `false` to ` MULTI_BUILD_CLUSTER_ENABLED`  and `SCHEDULER_ENABLED`
- Restart API and queue-service.

PR #2242, opened 2020-10-10
fix(2240): Convert MULTI_BUILD_CLUSTER_ENABLED value from string to boolean.

sakka2:
## Context
Kubernetes can not set boolean values as environment variables. 
So, we can not switch on/off by using `MULTI_BUILD_CLUSTER_ENABLED` environment variable.

## Objective
Convert `MULTI_BUILD_CLUSTER_ENABLED` value from string to boolean.

## References
Related Issue:
https://github.com/screwdriver-cd/screwdriver/issues/2240

## License


kkokufud:
[imo] Do we need to think about allowing the uppercase value?
ex. `True` `TRUE`

kkokufud (approved):
LGTM

tkyi (approved):

DekusDenial:
you can instead do something similar to https://github.com/screwdriver-cd/screwdriver/pull/2245/files#diff-f08ec2742d167ad6f3ad6485baab2bbbe1271d3e9eb7cf7e7792ba1c7e012171R405-R407

jithine (changes requested):
We should leverage native `config` package capabilities by using it's boolean parser. https://github.com/lorenwest/node-config/commit/97d4a5d251e69aeb3b6931af318a4a6a71360d7e#diff-774a94f3cb4758afd1f7feef2c1baa0a9a49ec1804a6a54c1264a607e30d7b27R7-R13

sugarnaoming:
> We should leverage native `config` package capabilities by using it's boolean parser. [lorenwest/node-config@97d4a5d#diff-774a94f3cb4758afd1f7feef2c1baa0a9a49ec1804a6a54c1264a607e30d7b27R7-R13](https://github.com/lorenwest/node-config/commit/97d4a5d251e69aeb3b6931af318a4a6a71360d7e#diff-774a94f3cb4758afd1f7feef2c1baa0a9a49ec1804a6a54c1264a607e30d7b27R7-R13)

Thank you.
Fixed use node-config parser, and revert to commits that were no longer needed.

DekusDenial:
I had to remove my previous comment because you would need to upgrade node-config to the version that includes boolean parser otherwise `__format: boolean` has no effect. Or, you can sidestep this by using `__format: json` so it can parse it as boolean.

sugarnaoming:
> I had to remove my previous comment because you would need to upgrade node-config to the version that includes boolean parser otherwise `__format: boolean` has no effect. But newer node-config is immutable and you would need to make some other changes as well. Or, you can sidestep this by using `__format: json` so it can parse string `true/false` as boolean.

Thank you.
In version 1.31.0, only the conversion to boolean was successful, so I was mistaken.  (In version 1.31.0, everything was converted to false.
This time we'll use the json format.

sugarnaoming:
Thank you for comment.
This time is fixed to use convert function.
There are two reasons for this
- Not enough version to use boolean config parser.
  - If we want to increase the version of config, we need to verify that there is no problem with the upgraded major version.
- If "__format: json" is specified, any value other than true/false (e.g. True/False) will result in an error.
  - yaml determines that `yes`, `y`, etc. are also true. (ref: https://yaml.org/type/bool.html)

kkokufud (approved):

jithine (approved):

bin/server
```diff
@@ -9,6 +9,24 @@ const hoek = require('@hapi/hoek');
...
+ * @method convertToBool
+ * @param {(Boolean|String)} value
+ * @return {Boolean}
+ */
+function convertToBool(value) {
+    if (typeof value === 'boolean') {
+        return value;
+    }
+
+    // trueList refers to https://yaml.org/type/bool.html
```
adong, line 23: 👍

sugarnaoming:
We created an issue to raise the version of node-config.
We may be able to work on it ourselves if we have the time.
https://github.com/screwdriver-cd/screwdriver/issues/2253

PR #2245, opened 2020-10-12
feat(ratelimit): add rate limiting

DekusDenial:
## Context

Given large enough user base, API might be subject to unexpected intense usage pattern from users. Providing an optional protection mechanism in place to limit the number of API request from each authenticated request would be helpful to maintain a healthy API service.

## Objective

- Add a rate limiting plugin behind a feature flag 
- Exclude rate limiting for paths which do not require authentication
    - GET /v4/auth/contexts
    - GET /v4/auth/crumb
    - GET /v4/auth/key
    - GET /v4/auth/login/guest/{web?}
    - GET /v4/auth/token/{buildId?}
      - This one is special because it takes in api_token (user/pipeline’s) and generates auth token whose jti field would be cache keyed if consumed on endpoints requiring authentication
    - GET /v4/banners
    - GET /v4/banners/{id}
    - GET /v4/documentation
    - GET /v4/metrics
    - GET /v4/pipelines/{id}/{jobName}/badge
    - GET /v4/pipelines/{id}/badge
    - GET /v4/stats
    - GET /v4/status
    - GET /v4/versions
    - POST /v4/validator
    - POST /v4/validator/command
    - POST /v4/validator/template
    - POST /v4/webhooks

## References

https://github.com/screwdriver-cd/guide/pull/419

## License

### DIFF
--- a/plugins/ratelimit.js
+++ b/plugins/ratelimit.js
@@ -1,0 +1,38 @@
+'use strict';
+
+const rateLimit = require('hapi-rate-limit');
+const config = require('config');
+const rateLimitConfig = config.get('rateLimit');
+
+module.exports = {
+    name: 'ratelimit',
+    async register(server) {
+        server.ext('onPreAuth', async (request, h) => {
+            // see https://github.com/glennjones/hapi-swagger/issues/623
+            if (request.path.startsWith('/v4/swagger')) {
+                request.route.settings.plugins['hapi-rate-limit'] = {
+                    enabled: false
+                };
+            }
+
+            return h.continue;
+        });
+
+        await server.register({
+            plugin: rateLimit,
+            options: {
+                enabled: rateLimitConfig.enabled || false,
+                userLimit: rateLimitConfig.limit,
+                userAttribute: 'jti',
+                userCache: {
+                    expiresIn: rateLimitConfig.duration
+                },
+                authLimit: false,
+                headers: false,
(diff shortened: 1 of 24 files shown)


tkyi (dismissed):

tkyi:
test failed because: https://github.com/screwdriver-cd/config-parser/blob/master/package.json#L57

plugins/ratelimit.js
```diff
@@ -0,0 +1,38 @@
...
+const rateLimit = require('hapi-rate-limit');
+const config = require('config');
+const rateLimitConfig = config.get('rateLimit');
+
+module.exports = {
+    name: 'ratelimit',
+    async register(server) {
+        server.ext('onPreAuth', async (request, h) => {
+            // see https://github.com/glennjones/hapi-swagger/issues/623
+            if (request.path.startsWith('/v4/swagger')) {
```
jithine, line 12: should we store all whitelisted paths in an array config, and handle in preAuth hook, rather than adding to individual plugins like in the rest of the files in this PR ?

DekusDenial: Keeping the configuration close to where the route is defined promote integrity similar to how `auth` is defined for each route. Also, one less place to change if certain routes change to require authenticate. This swagger hook you are seeing is due to current limitation on hapi-swagger.

tkyi (approved):

PR #2251, opened 2020-10-14
fix: Bump up API npm version

kumada626:
## Context

API npm package version is updated by [npm-auto-version](https://www.npmjs.com/package/npm-auto-version).
This package updates tag based on version in package.json.
So new tag is `v4.0.X`.
https://github.com/screwdriver-cd/screwdriver/tags
However, we had already tagged `v4.1.0` by template in [this build](https://cd.screwdriver.cd/pipelines/1/builds/578833).
Therefore, we use `v4.1.0` as latest tag instead of `4.0.X`.

## License

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "screwdriver-api",
-  "version": "4.0.0",
+  "version": "4.1.0",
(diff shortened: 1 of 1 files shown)


jithine (approved):
👏 so this is what was going on.

PR #2254, opened 2020-10-16
feat: add provision to set canary cookie

pritamstyz4ever:
## Context

To support canary data flow consistency across screwdriver applications (ui,api,store) we need environment information to add a new cookie

## Objective

This PR adds a cookie if certain env vars are present to make a canary session sticky.

## License


config/default.yaml
```diff
@@ -300,6 +300,20 @@ ecosystem:
     max_size_mb: 0
     max_go_threads: 10000
 
+# environment release information
+release:
+  # release mode stable or canary
+  mode: stable
+  # name of cookie
+  cookieName: biscotti
```
jithine, line 308: let's call these something related to SD ?
`cookieName:release`, `cookieValue: stable`, header should follow suit

lib/server.js
```diff
@@ -14,9 +14,13 @@ process.on('unhandledRejection', (reason, p) => {
  * @param  {Hapi.Request}    request Hapi Request object
  * @param  {Hapi.h}     h   Hapi Response Toolkit
  */
-function prettyPrintErrors(request, h) {
+function prettyPrintErrors(request, h, release) {
     const { response } = request;
 
+    if (release && release.mode === 'canary' && !(request.state && request.state[release.cookieName])) {
```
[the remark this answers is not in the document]
adong: I think we can get the cookie value of `release` from `request.state.data`.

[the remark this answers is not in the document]
adong: And get `release` value from `config.release` rather than a passing parameter. 

So this code does not have to change.

`server.ext('onPreResponse', prettyPrintErrors);`

adong:
I was thinking if we can take an advantage of using [hapi lifecycle](https://futurestud.io/files/hapi/hapi-request-lifecycle.pdf) to handle this scenario.

1. in `onPreHandler` check cookies and route to add 
2. in `onPreResponse` to set cookie if not set already.

I felt like this is way is less intrusive and pluggable, in `lib/server.js`

```javascript
if (config.release) {
        server.ext('onPreHandler', (request, h) => {
           console.log('onPreHandler');
        })

        server.ext('onPreResponse', (request, h) => {
            console.log('onPostHandler');
        })
}
```

I might get the life cycle names wrong here, but you get the idea.

lib/server.js
```diff
@@ -130,13 +135,23 @@ module.exports = async config => {
...
         };
 
         const bellConfigs = await config.auth.scm.getBellConfiguration();
 
         config.auth.bell = bellConfigs;
 
+        if (config.release && config.release.cookieName) {
+            server.state(config.release.cookieName, {
+                path: '/',
+                ttl: (config.auth.sessionTimeout / 60) * 60 * 1000, // setting to 2 minutes
```
adong, line 149: Hmm, why not just `config.auth.sessionTimeout * 1000` ?

config/default.yaml
```diff
@@ -300,6 +300,9 @@ ecosystem:
     max_size_mb: 0
     max_go_threads: 10000
 
+# environment release information
+release: {}
```
adong, line 304: Looks like the previous sample was removed.

pritamstyz4ever: Yeah, we don't need to keep that, default can be empty {}

lib/server.js
pritamstyz4ever: Agree, although kind of changes the understanding of config.auth.sessionTimeout being set to 120 minutes.
https://github.com/screwdriver-cd/screwdriver/blob/master/config/default.yaml#L39-L40

config/default.yaml
jithine: I liked having the previous default, it helped anyone overriding to know what exactly should go in `RELEASE_ENVIRONMENT_VARIABLES`.

pritamstyz4ever: ok, although if we keep the default values the cookie we will need the override in release.

jithine (approved):

Issue #2288, opened 2020-11-30
Ensure all Screwdriver projects have proper code quality analysis


klu909:
**What happened**:
Screwdriver projects must have static and dynamic code analysis done as part of build process, most of them has broken sonarqube setting.
**What you expected to happen**:
Fix sonarqube settings and have proper static and dynamic code analysis .
**How to reproduce it**:

PR #2291, opened 2020-12-01
fix(2288): fix sonarqube setting

klu909:
## Context

Sonarqube setting was broken and doesn't have test execution setting
## Objective

Fix sonarqube setting
## References

https://github.com/screwdriver-cd/screwdriver/issues/2288

## License


screwdriver.yaml
```diff
@@ -22,6 +22,7 @@ jobs:
             - coverage: ./ci/coverage.sh
```
klu909, line 22: Do we still need coverAll? If not we can remove some unnecessary steps.

jithine (approved):

jithine:
There is a difference in coverage reported by coverall & sonarqube

klu909:
> There is a difference in coverage reported by coverall & sonarqube

There's probably subtle difference in which file they choose to scan?

Issue #2168, opened 2020-07-31
Highlight the latest commit in the Events tab


ibu1224:
**What happened**:
When there are a lot of events, it's hard to know which is the recent commit.

**What you expected to happen**:
It makes it easier to see which is the latest commit.

**How to reproduce it**:
Look for the latest commit in the pipeline with many events in the pipeline.

yoshwata:
Let me propose about how work it looks like.

<img src="https://user-images.githubusercontent.com/32473622/98073145-04944980-1eab-11eb-8805-94b0aa1766e9.jpg" width=300 />

Is it looks good? @jithine @tkyi @adong

wahapo:
Let me suggest another idea. What do you think if we put a green box around the latest commit SHA and show the tag as last successful?

yoshwata:
Your suggestion seems reasonable. What we want to express about sha should be decorated on sha. 

<img src="https://user-images.githubusercontent.com/32473622/98076513-bc2c5a00-1eb1-11eb-8fc4-7d540907fe24.jpg" width=300 />

- Green box means latest commit sha
- Last successful is moved from its original location and displayed in a blue box

wahapo:
It would be better to have a couple of target SHA in the image to help us visualize the image, but I think it's good.

ibu1224:
Basically, I agree with Proposal 2.
However, we need to think about the color of the latest commit sha.
In screwdriver.cd, green means success.
The color of `last successful` should be green.

yoshwata:
# Proposal 2-2

@ibu1224 Thank you. The image below reflects your idea. And I think this is the best.

<img src="https://user-images.githubusercontent.com/32473622/98083082-239bd700-1ebd-11eb-9b8f-cc661abd7dac.jpg" width=300 />

jithine:
@yoshwata couple of questions

1. will events still be sorted chronologically in the events list ?
2. last successful will have the green label
   1. Since we already have the green check mark against the successful event what more info does the label provide ?
   1. Is the idea to also leverage success states of the pipelines from https://github.com/screwdriver-cd/screwdriver/issues/2199 ?
3. latest commit will have the blue label
    1. If there are multiple events with last commit sha will all of them get the blue label ?
     1. Slightly related to https://github.com/screwdriver-cd/screwdriver/issues/1820

yoshwata:
@jithine Thanks

> will events still be sorted chronologically in the events list ?

[...]

PR #2295, opened 2020-12-03
fix(2168): add latestCommitEvent

wahapo:
## Context
To get the latest commit event SHA, this PR add the API to get the latest commit event.

## Objective
The following API returns the most recent event with `parentEventId` is null.
```
GET /v4/pipelines/{id}/latestCommitEvent
```

The API `/v4/pipelines/{id}/events` cannot be reused for this case. Since only a certain number of events at a time are returned, it will not work as expected if all the events returned at once have a `parentEventId`.

For example, if you go to the following URL(This is a permanent link to an old event), id:599751 is the only one that does not have a parentEventID, so it can be mistakenly assumed to be the most recent commit event.
https://cd.screwdriver.cd/pipelines/4862/events/599751

In fact, If you scroll down the event list, you will find more new commit events.

So I created a dedicated API this PR.

## References
https://github.com/screwdriver-cd/screwdriver/issues/2168

## License


ibu1224 (approved):

plugins/pipelines/latestCommitEvent.js
```diff
@@ -0,0 +1,50 @@
...
+        auth: {
+            strategies: ['token'],
+            scope: ['user', 'build', 'pipeline']
+        },
+
+        handler: async (request, h) => {
+            const { eventFactory } = request.server.app;
+
+            return eventFactory
+                .list({
```
klu909, line 24: Do we need use `list` here? Using `find` with same parameter will get what you want.

wahapo: Thanks, a lot of lists are unnecessary.
Is the `find` method already there somewhere? I added `limit 1` because I couldn't find it.

plugins/pipelines/latestCommitEvent.js
```diff
@@ -0,0 +1,53 @@
...
+const joi = require('joi');
+const schema = require('screwdriver-data-schema');
+const idSchema = schema.models.pipeline.base.extract('id');
+
+module.exports = () => ({
+    method: 'GET',
+    path: '/pipelines/{id}/latestCommitEvent',
+    options: {
+        description: 'Get latest commit event for a given pipeline',
+        notes: 'Return latest commit sha',
```
yoshwata, line 13: It seems that it returns a event.

yoshwata (approved):

plugins/pipelines/latestCommitEvent.js
klu909: yeah `limit 1` should be sufficient

klu909 (approved):
LGTM

wahapo:
@jithine please review and merge this PR:bow:

jithine (approved):

Issue #2189, opened 2020-08-24
`Pipeline has no admin` is handling as 500 error


s-yoshika:
**What happened**:
Recently, the number of requests with 500 error is increasing because of `pipeline has no admin` in our SD API.
Especailly, it tends to increase the error number if the jobs in the pipeline with no admins are configured as periodic build.
It would be noise for monitoring API succesful rate.

**What you expected to happen**:
I'd like to change the status code of `pipeline has no admin` to appropriate one.

**How to reproduce it**:

PR #2296, opened 2020-12-03
fix(2189): modify error code for GET  pipelines/{}/admin

MysticDoll:
## Context
The pipeline that has no admins causes 500 error and we should return another status code.

## Objective
- fix to use 404 for the error of "pipeline has no admin"

## References
screwdriver-cd/screwdriver#2189

## License


tkyi (approved):

yuichi10 (approved):
LGTM

Issue #1826, opened 2019-11-02
Clickable URLs in console output


ashwinjk:
**What happened**:
When the build console log for a job contains URLs, it is rendered as plain text

**What you expected to happen**:
Detect URLs in build logs and make them anchor link so they can be easily clicked

coldmarble:
We have a similar issue happening in our pipelines where in the links are not clickable like they were in sd v3.

klu909:
We have tried ui/api changes to render clickable url, in both case we observe a notice slow down for large step logs. For now this feature is place on hold, and we will revisit this feature again when we redesign step logs.

PR #2300, opened 2020-12-08
feat(1826):Clickable URLs in console output

klu909:
## Context

When the build console log for a job contains URLs, it is rendered as plain text
## Objective

Replace urls in with anchor link and escape rest of log message.

## References

https://github.com/screwdriver-cd/screwdriver/issues/1826
## License

### DIFF
--- a/plugins/builds/steps/logs.js
+++ b/plugins/builds/steps/logs.js
@@ -8,7 +8,7 @@ const MAX_LINES_SMALL = 100;
 const MAX_LINES_BIG = 1000;
 const jwt = require('jsonwebtoken');
 const uuid = require('uuid');
-
+const validator = require('validator');
 const logger = require('screwdriver-logger');
 
 /**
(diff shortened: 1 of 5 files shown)


plugins/builds/steps/logs.js
```diff
@@ -207,20 +207,36 @@ module.exports = config => ({
...
+                                    .response(res)
+                                    .type('text/plain')
+                                    .header('content-disposition', `attachment; filename="${stepName}-log.txt"`);
                             }
 
-                            return h
-                                .response(res)
-                                .type('text/plain')
-                                .header('content-disposition', `attachment; filename="${stepName}-log.txt"`);
+                            const formatedLine = lines.map(line => {
```
jithine, line 223: I think looping through is  inefficient, using a regex over the string is what comes to my mind, and regex optimization is an already understood problem.

Issue #2313, opened 2021-01-12
Sonar Github PR decoration not always added


tkyi:
**What happened**: When using the Sonar [Github PR decoration feature](https://docs.screwdriver.cd/user-guide/configuration/code-coverage#sonarqube), the Sonar configuration is not always added correctly to the Sonar project.

**What you expected to happen**: Should see Sonar Github PR decoration in pull request. (See Pull Request Decoration section: https://blog.screwdriver.cd/post/626730108375465985/sonarqube-enterprise-edition-support)

**How to reproduce it**:
1. Add the Screwdriver Sonar PR Checks Github App to your repository.
2. Open a PR.

Related to https://github.com/screwdriver-cd/screwdriver/issues/2198

PR #2317, opened 2021-01-15
fix(2313): Send pipelineName and jobName when creating Sonar project [1]

tkyi:
## Context

Some job-scoped Sonar projects have `undefined:undefined` set as their Github bindings.

## Objective

This PR sends `jobName` and `pipelineName` when creating Sonar projects and setting Github bindings.

## References

Related to #2313 

## License

### DIFF
--- a/plugins/coverage/token.js
+++ b/plugins/coverage/token.js
@@ -16,9 +16,9 @@ module.exports = config => ({
         },
 
         handler: async (request, h) => {
-            const { jobFactory } = request.server.app;
+            const { jobFactory, pipelineFactory } = request.server.app;
             const buildCredentials = request.auth.credentials;
-            const { jobId } = buildCredentials;
+            const { jobId, pipelineId } = buildCredentials;
             const { scope, projectKey, projectName, username } = request.query;
             const tokenConfig = {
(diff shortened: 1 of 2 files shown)


plugins/coverage/token.js
```diff
@@ -37,27 +37,33 @@ module.exports = config => ({
...
+                tokenConfig.jobName = job.name;
                 tokenConfig.scope =
                     job.permutations[0] && job.permutations[0].annotations
                         ? job.permutations[0].annotations[COVERAGE_SCOPE_ANNOTATION]
                         : null;
+            }
 
-                data = await config.coveragePlugin.getAccessToken(tokenConfig);
+            // Get pipeline name
+            if (pipelineId && (!projectName || projectName.includes('undefined'))) {
```
pritamstyz4ever, line 56: is projectName an array? should we make it plural?

tkyi: No, it's just a string
sometimes it's set to `undefined:undefined`, so just trying to catch that
https://www.w3schools.com/jsref/jsref_includes.asp

jithine (approved):

PR #2353, opened 2021-02-22
fix: remove user id from url param

adong:
## Context

userId is not exposed to UI easily, and I was able to find my userId by using the return value of creating a pipeline collection. 
## Objective

Backend needs to find the user by using its username and scmContext
## References

https://github.com/screwdriver-cd/ui/pull/671
## License

### DIFF
--- a/plugins/users/settings/get.js
+++ b/plugins/users/settings/get.js
@@ -1,9 +1,7 @@
 'use strict';
 
 const boom = require('@hapi/boom');
-const joi = require('joi');
 const schema = require('screwdriver-data-schema');
-const idSchema = schema.models.user.base.extract('id');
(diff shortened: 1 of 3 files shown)


jithine (approved):

tkyi:
I think this makes sense for PUT user endpoint but less so for GET, since anyone should be able to do GET

jithine:
> I think this makes sense for PUT user endpoint but less so for GET, since anyone should be able to do GET

should anyone be able to do a GET on other users config data ? With this change only the user present in Bear token's config can be fetched/updated

adong:
> I think this makes sense for PUT user endpoint but less so for GET, since anyone should be able to do GET

`userId` is not exposed in the token info.

PR #2374, opened 2021-03-17
fix: refactor triggerNextJobs function

jithine:
## Context

`triggerNextJobs` function is responsible for running next jobs after current running job is success. It has grown too big and code has become difficult to track.

## Objective

1. Remove externalJoin flag and make remote-join feature permanent.
2. refactor code to clear flows

## References

https://github.com/screwdriver-cd/screwdriver/pull/2374#issuecomment-812766411

## License
I confirm that this contribution is made under a BSD license and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #2405, opened 2021-04-08
fix: do proper merging of parent build data

jithine:
## Context

1. getFinishedBuilds doesn't return build model
2. Prevent _.mergeWith from mutating sources and `null` values needs to be handled properly

## References

https://github.com/screwdriver-cd/screwdriver/pull/2374

## License

### DIFF
--- a/plugins/builds/index.js
+++ b/plugins/builds/index.js
@@ -763,6 +763,7 @@ const buildsPlugin = {
 
                     newBuild = await createInternalBuild(internalBuildConfig);
                 } else {
+                    console.log('before updateParentBuilds', nextBuild.id);
                     // nextBuild is not build object, so fetch proper build
                     newBuild = await updateParentBuilds({
                         joinParentBuilds: parentBuilds,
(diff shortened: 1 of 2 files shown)


plugins/builds/index.js
```diff
@@ -401,7 +401,10 @@ async function getFinishedBuilds(event, buildFactory) {
  */
 async function updateParentBuilds({ joinParentBuilds, nextBuild, build }) {
     // Override old parentBuilds info
-    const newParentBuilds = merge({}, joinParentBuilds, nextBuild.parentBuilds, (objVal, srcVal) => objVal || srcVal);
+    const newParentBuilds = merge({}, joinParentBuilds, nextBuild.parentBuilds, (objVal, srcVal) =>
+        // passthrough objects, else mergeWith mutates source
+        srcVal && typeof srcVal === 'object' ? undefined : objVal || srcVal
```
adong, line 406:
```
(srcVal && typeof srcVal === 'object' ? undefined : objVal) || srcVal
```

If `typeof srcVal === 'object'` then we use `undefined`, otherwise use `objVal`? Not sure the intent here, it just sounds a little off.

jithine: https://lodash.com/docs/4.17.15#mergeWith

` If customizer returns undefined, merging is handled by the method instead` if value is object default mergeWith functionality. Else `mergeWith` will mutate source object

Issue #2071, opened 2020-04-16
Multiple builds are created for a join job


jithine:
**What happened**:

We're noticing cases where multiple builds are created for a join job. In instance we noticed there were 4 builds created for the join job, with 2 in `SUCCESS` state and 2 in `CREATED` state.

Also one of the `SUCCESS` builds was started because it was ran from a different event's join. See https://github.com/screwdriver-cd/screwdriver/issues/2070

So original state would have been 1 `SUCESS` and 3 `CREATED` builds. 

**What you expected to happen**:

Only one join build should be created for an event. 

**How to reproduce it**:

Difficult. In the event where this was observed, parent builds for the `job` in question completed about same time. 
For example the following config
```
jobs:
   parent-1:
   parent-2:
   parent-3:
   parent-4:
   join:
    requires: [parent-1, parent-2,  parent-3, parent-4]   
```
This issue can happen if builds for parent-1, parent-2,  parent-3, parent-4 completes about same time.

jithine:
Another side effect of this issue is that the join job sometimes doesn't run because the parentBuilds info gets overwritten when multiple parent's complete at about same time.

jithine:
Another scenario when locking is required is for `blockedBy` use case. If there are 2 jobs which blocks each other and when both builds are created almost at same time, it tends to break `blockedBy` guarantees.

jithine:
Explore making the following action into a transaction 
1. Update event.meta
2. SELECT build id for join job name
3. CREATE/UPDATE join build

itleigns:
This problem can still happen. Redlock prevents [triggerNextJobInSamePipeline](https://github.com/screwdriver-cd/screwdriver/blob/d8bd3979fa573ed2cebae7b99432e71b0df62a93/plugins/builds/index.js#L788) from running in parallel, but not for [build state updates](https://github.com/screwdriver-cd/screwdriver/blob/34d7a9ec545a0dc9516e3b5f6d9d18f47dc23d91/plugins/builds/update.js#L228). So this problem occurs in the following cases.
```
jobs:
   parent-1:
   parent-2:
   join:
    requires: [parent-1, parent-2]
```
- `parent-1` succeeds and the status of the build is set  SUCCESS
- `parent-2` succeeds and the status of the build is set  SUCCESS
- `parent-1` triggers `join`
- `parent-2` triggers `join`

[...]

PR #2408, opened 2021-04-09
fix(2071): trigger next build could create duplicate build

klu909:
## Context

multiple build could be created same job which splits parent build status and will never start

## Objective
only one build should be created for same job

## References
https://github.com/screwdriver-cd/screwdriver/issues/2071

## License

### DIFF
--- a/package.json
+++ b/package.json
@@ -125,7 +125,9 @@
     "sqlite3": "^5.0.1",
     "tinytim": "^0.1.1",
     "uuid": "^8.3.2",
-    "verror": "^1.6.1"
+    "verror": "^1.6.1",
+    "ioredis": "^3.2.2",
+    "redlock": "^4.1.0"
   },
   "release": {
     "debug": false,
--- a/package.json
+++ b/package.json
@@ -87,6 +87,7 @@
     "hapi-auth-jwt2": "^10.2.0",
     "hapi-rate-limit": "^5.0.0",
     "hapi-swagger": "^14.1.0",
+    "ioredis": "^3.2.2",
     "joi": "^17.4.0",
     "js-yaml": "^3.14.1",
     "jsonwebtoken": "^8.5.1",
@@ -95,6 +96,7 @@
     "ndjson": "^1.4.3",
     "node-env-file": "^0.1.8",
     "prom-client": "^12.0.0",
+    "redlock": "^4.1.0",
     "request": "^2.88.2",
     "requestretry": "^4.1.2",
     "screwdriver-artifact-bookend": "^1.1.27",
@@ -125,9 +127,7 @@
     "sqlite3": "^5.0.1",
     "tinytim": "^0.1.1",
     "uuid": "^8.3.2",
-    "verror": "^1.6.1",
-    "ioredis": "^3.2.2",
-    "redlock": "^4.1.0"
+    "verror": "^1.6.1"
   },
   "release": {
     "debug": false,
--- a/plugins/builds/index.js
+++ b/plugins/builds/index.js
@@ -17,6 +17,8 @@ const listSecretsRoute = require('./listSecrets');
 const tokenRoute = require('./token');
 const metricsRoute = require('./metrics');
 const { EXTERNAL_TRIGGER_ALL } = schema.config.regex;
+const redLock = require('../lock.js');
+const ttl = 20000;
 
 /**
  * Checks if job is external trigger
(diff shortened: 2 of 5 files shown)


plugins/builds/index.js
```diff
@@ -936,7 +985,13 @@ const buildsPlugin = {
                 if (+pid === current.pipeline.id) {
                     for (const nextJobName of Object.keys(pipelineJoinData[pid].jobs)) {
                         try {
+                            const resource = `pipeline:${current.pipeline.id}:event:${current.event.id}`;
```
jithine, line 988: resource should be `nextPipelineId:nextJobName` if we do that, do we need any of the other changes ? since this will guarantee a sequential flow ?

plugins/builds/index.js
```diff
@@ -761,9 +794,14 @@ const buildsPlugin = {
                         parentBuildId: current.build.id
                     };
 
-                    newBuild = await createInternalBuild(internalBuildConfig);
-                } else {
-                    // nextBuild is not build model, so fetch proper build
+                    try {
+                        newBuild = await createInternalBuild(internalBuildConfig);
+                    } catch (err) {
+                        nextBuild = await getNextBuild(nextBuildConfig);
```
tkyi, line 800: isn't this redundant because of line 778?

plugins/builds/index.js
```diff
@@ -761,9 +794,14 @@ const buildsPlugin = {
...
+                    }
+                }
+
+                if (!newBuild && nextBuild) {
                     newBuild = await updateParentBuilds({
                         joinParentBuilds: parentBuilds,
                         nextBuild: await buildFactory.get(nextBuild.id),
```
tkyi, line 807: can you just pass nextBuild directly here?
```suggestion
                        nextBuild
```

plugins/builds/index.js
```diff
@@ -928,6 +975,8 @@ const buildsPlugin = {
                     groupEventId: null
                 };
 
+                console.log('here');
```
tkyi, line 978: remove this line

plugins/lock.js
```diff
@@ -0,0 +1,40 @@
+'use strict';
+
+const Redis = require('ioredis');
+const Redlock = require('redlock');
+const config = require('config');
+
+class Lock {
+    constructor() {
```
tkyi, line 8: probably should add jsdoc like this: https://github.com/screwdriver-cd/executor-k8s/blob/master/index.js#L163-L218

plugins/lock.js
```diff
@@ -0,0 +1,40 @@
...
+            this.redlock = new Redlock([this.redis], {
+                driftFactor: redisLockConfig.driftFactor,
+                retryCount: redisLockConfig.retryCount,
+                retryDelay: redisLockConfig.retryDelay,
+                retryJitter: redisLockConfig.retryJitter
+            });
+        }
+    }
+
+    getLock(resource, ttl) {
```
tkyi, line 31: missing jsdoc

package.json
```diff
@@ -125,7 +125,9 @@
     "sqlite3": "^5.0.1",
     "tinytim": "^0.1.1",
     "uuid": "^8.3.2",
-    "verror": "^1.6.1"
+    "verror": "^1.6.1",
+    "ioredis": "^3.2.2",
+    "redlock": "^4.1.0"
```
tkyi, line 130: should be alphabetical order

jithine (changes requested):
I think the logic change can be simplified.

1. OR join - duplicate build creation error should be ignored and returned immediately.
2. Join job - simply lock on event_id + job_id for next job. Lock should ensure the sequential logic and there is no extra handling needed.

plugins/builds/index.js
klu909: so this is a try catch for race condition, build creation fail if there's already a build created for that job. Since we are not using infinite retry lock, so just adding this try catch as a fall back mechanism.

plugins/builds/index.js
klu909: well point is that resources locked should be same for external event, for external event it's quite tricky to lock in format of nextPipelineId:nextJobName due to current implementation

plugins/lock.js
```diff
@@ -0,0 +1,51 @@
...
+                driftFactor: redisLockConfig.driftFactor,
+                retryCount: redisLockConfig.retryCount,
+                retryDelay: redisLockConfig.retryDelay,
+                retryJitter: redisLockConfig.retryJitter
+            });
+        }
+    }
+
+    /**
+     * attempt to acquire a lock for resource, will retry accuqiring lock depending
```
tkyi, line 35:
```suggestion
     * Attempt to acquire a lock for resource, will retry acquiring lock depending
```

jithine (approved):

tkyi (approved):

PR #2426, opened 2021-04-23
[2] fix: Add error handling for CircuitBreaker timeout

ibu1224:
## Context

When a timeout occurs in CircuitBreaker, err.status becomes undefined.
Now the following code will be processed successfully.
https://github.com/screwdriver-cd/screwdriver/blob/f1b25fd35576310e9119393cb890ffbb625f1317/plugins/webhooks/index.js#L416-L421

This will cause problems in later processing, Add the case when a timeout occurs in CircuitBreaker.

It also adds information on which pipeline the error occurred.

## Objective

In this PR, I will add error handling when CircuitBreaker times out in [getCommitSha()](https://github.com/screwdriver-cd/scm-github/blob/4c2a843ce5266a14b27d4788b3cd6d9b8f6e98a3/index.js#L860).

Circuitbreaker will output a log like the following when it times out.
```
Failed to getCommitSha:  CircuitBreaker timeout
```

~~Error handling is done with the `CircuitBreaker timeout` string in err.message. Is there a better way?~~
Fix circuit breaker to return error status 504 on timeout

## References
[1] https://github.com/screwdriver-cd/circuit-fuses/pull/30

## License

### DIFF
--- a/plugins/webhooks/index.js
+++ b/plugins/webhooks/index.js
@@ -404,8 +404,6 @@ async function createPREvents(options, request) {
                 } catch (err) {
                     if (err.status >= 500) {
                         throw err;
-                    } else if (err.message !== undefined && err.message.indexOf('CircuitBreaker timeout') !== -1) {
-                        throw err;
                     } else {
                         logger.info(`skip create event for branch: ${b}`);
                     }
@@ -417,8 +415,6 @@ async function createPREvents(options, request) {
                     configPipelineSha = await pipelineFactory.scm.getCommitSha(scmConfig);
                 } catch (err) {
                     if (err.status >= 500) {
-                        throw err;
-                    } else if (err.message !== undefined && err.message.indexOf('CircuitBreaker timeout') !== -1) {
                         throw err;
                     } else {
                         logger.info(`skip create event for branch: ${b}`);
@@ -891,8 +887,6 @@ async function createEvents(
                 } catch (err) {
                     if (err.status >= 500) {
                         throw err;
-                    } else if (err.message !== undefined && err.message.indexOf('CircuitBreaker timeout') !== -1) {
-                        throw err;
                     } else {
                         logger.info(`skip create event for branch: ${pipelineBranch}`);
                     }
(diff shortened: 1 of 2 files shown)


plugins/webhooks/index.js
```diff
@@ -887,6 +891,8 @@ async function createEvents(
                 } catch (err) {
                     if (err.status >= 500) {
                         throw err;
+                    } else if (err.message !== undefined && err.message.indexOf('CircuitBreaker timeout') !== -1) {
```
wahapo, line 894: How about using "if `err.status` is `undefined`"?

jithine, line 894: what's the status code for curcuit breaker errors ? perhaps we should define something at https://github.com/screwdriver-cd/circuit-fuses/blob/d99a1b3679803624f941eb8668ca2274eda37393/index.js to make that clear.

Also the `else` flow of simply logging seems to handle a valid scenario which is not well defined.

ibu1224: You're right.
I'll change the way I fix it.

ibu1224: @jithine 
Fixed by the following PR.
https://github.com/screwdriver-cd/circuit-fuses/pull/30

plugins/webhooks/index.js
```diff
@@ -542,7 +542,7 @@ async function pullRequestOpened(options, request, h) {
             return h.response().code(201);
         })
         .catch(err => {
-            logger.error(`[${hookId}]: ${err}`);
+            logger.error(`[${hookId}, pipeline:${options.pipeline.id}]: ${err}`);
```
yoshwata, line 545: Don't you touch pullRequestSync and pullRequestClosed?

yoshwata: I prefer such logs that shows where it occurs.
スクリーンショット 2021-04-26 15 18 21

wahapo (approved):

yoshwata (approved):

tkyi (approved):

Issue #2399, opened 2021-04-01
Can not remove tags from commands


itleigns:
**What happened**:
We do not have a way to remove tags from [commands](https://docs.screwdriver.cd/user-guide/commands).

**What you expected to happen**:
We want an API to remove a tag from the command in the same way as the tags of templates.

kumada626:
We added an API endpoint for remove command tag in #2400 .
We will continue to implement tag remove command on sd-cmd cli.

itleigns:
Currently, the API for remove command tag does not return response body in normal cases. So `sd-cmd removeTag` always throws error. We will fix the API to return response body.

PR #2431, opened 2021-04-28
feat(2399): Change remove command tag API

itleigns:
## Context
The APIs for commands return some kind of response body except for removeTag. In sd-cmd, the implementation assumes that the API for commands returns a response body. So we want to change the API for remove commad tags to return a response body.

## Objective
- We modify the command removeTag API to return a response body.
- We change the status code returned in normal case from 204 to 200

## References
https://github.com/screwdriver-cd/screwdriver/issues/2399
https://github.com/screwdriver-cd/sd-cmd/blob/master/screwdriver/api/api_test.go#L52-L57

## License

### DIFF
--- a/plugins/commands/removeTag.js
+++ b/plugins/commands/removeTag.js
@@ -46,10 +46,21 @@ module.exports = () => ({
                         }
 
                         // Remove the command tag, not the command
-                        return commandTag.remove();
+                        return commandTag.remove().then(() =>
+                            h
+                                .response(
+                                    JSON.stringify({
+                                        name,
+                                        namespace,
+                                        version: commandTag.version,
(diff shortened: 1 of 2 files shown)


wahapo (approved):

tkyi (approved):

plugins/commands/removeTag.js
```diff
@@ -46,10 +46,21 @@ module.exports = () => ({
                         }
 
                         // Remove the command tag, not the command
-                        return commandTag.remove();
+                        return commandTag.remove().then(() =>
+                            h
```
tk3fftk, line 50: has this indent broken? ESLint has passed but it looks weird.

Issue #2432, opened 2021-04-28
Commits created by detached jobs are not triggering the right previous pushed commit 


martoc:
**What happened**:

When a job `A` commits some code to the same repo then job `B` is triggered with the previous commit and not the new one we are using this `requires: [~commit, ~sd@1000000:A]` in job `B`

**What you expected to happen**:

it should trigger B with the new commit created by A

**How to reproduce it**:

```
  A:
    steps:
      - init: |
          date > file.txt
          git add file.txt
          git commit -m "Updating file at $(date)"
          git push origin master
B: 
  requires: [~commit, ~sd@1000000:A]
  steps:
     - build: echo $SD_BUILD_SHA
```

tkyi:
If you want job B to be triggered with job A's commit only, you can remove the `~commit` in the requires.
You can also drop the external syntax for A if it's in the same pipeline: `~sd@ 1000000 :A` => `~A`.
The result would then look like this:
```
A:
    steps:
      - init: |
          date > file.txt
          git add file.txt
          git commit -m "Updating file at $(date)"
          git push origin master
B: 
  requires: [~A]
  steps:
     - build: echo $SD_BUILD_SHA
```

PR #2433, opened 2021-04-28
fix(2432): support external trigger format for same pipeline

jithine:
## Context

Users can use external trigger syntax for same pipeline 🙄 

## References

https://github.com/screwdriver-cd/screwdriver/issues/2432

## License

### DIFF
--- a/plugins/builds/index.js
+++ b/plugins/builds/index.js
@@ -969,10 +969,10 @@ const buildsPlugin = {
                 if (isCurrentPipeline) {
                     for (const nextJobName of Object.keys(pipelineJoinData[pid].jobs)) {
                         try {
-                            const forceExternalTrigger = pipelineJoinData[pid].jobs[nextJobName].isExternal;
+                            const isExternal = pipelineJoinData[pid].jobs[nextJobName].isExternal;
 
-                            triggerCurrentPipelineAsExternal = triggerCurrentPipelineAsExternal || forceExternalTrigger;
-                            if (!forceExternalTrigger) {
+                            triggerCurrentPipelineAsExternal = triggerCurrentPipelineAsExternal || isExternal;
+                            if (!isExternal) {
                                 const resource = `pipeline:${current.pipeline.id}:event:${current.event.id}`;
                                 const lock = await locker.lock(resource);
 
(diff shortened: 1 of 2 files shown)


plugins/builds/index.js
```diff
@@ -618,6 +620,10 @@ async function createJoinObject(nextJobs, current, eventFactory) {
         const jobInfo = getPipelineAndJob(jobName, current.pipeline.id);
         const pid = jobInfo.externalPipelineId;
         const jName = jobInfo.externalJobName;
+        // needed to determine if join in same pipeline
+        // is expressed in external format
+        const isExternal = jobInfo.isExternal;
```
tkyi, line 625:
```suggestion
        // needed to determine if join in same pipeline
        // is expressed in external format
         const { externalPipelineId: pid, externalJobName: jName, isExternal } = jobInfo;
```

plugins/builds/index.js
```diff
@@ -961,23 +967,36 @@ const buildsPlugin = {
...
+                const isCurrentPipeline = +pid === current.pipeline.id;
+
+                if (isCurrentPipeline) {
                     for (const nextJobName of Object.keys(pipelineJoinData[pid].jobs)) {
                         try {
-                            const resource = `pipeline:${current.pipeline.id}:event:${current.event.id}`;
-                            const lock = await locker.lock(resource);
+                            const forceExternalTrigger = pipelineJoinData[pid].jobs[nextJobName].isExternal;
+
+                            triggerCurrentPipelineAsExternal = triggerCurrentPipelineAsExternal || forceExternalTrigger;
```
tkyi, line 978: should order be switched?
```suggestion
                            triggerCurrentPipelineAsExternal = forceExternalTrigger || triggerCurrentPipelineAsExternal;
```

plugins/builds/index.js
```diff
@@ -961,23 +967,36 @@ const buildsPlugin = {
...
+                            if (!forceExternalTrigger) {
```
tkyi, line 979:
```suggestion
                            if (!triggerCurrentPipelineAsExternal) {
```

plugins/builds/index.js
jithine: doesn't matter, we are looking for atleast one job which is of external format for same pipeline

plugins/builds/index.js
jithine: no this won't be correct, there could be mix of jobs, some in external format and some not. we are looking specifically at current job

klu909 (approved):

PR #2435, opened 2021-04-29
feat(466): add read protection against log and artifacts

klu909:
## Context

Step logs and artifacts should be protected for private pipeline
## Objective

Add read protection for step logs and artifacts
## References

## License

### DIFF
--- a/plugins/pipelines/index.js
+++ b/plugins/pipelines/index.js
@@ -115,25 +115,39 @@ const pipelinesPlugin = {
                     throw boom.notFound(`Pipeline ${pipelineId} does not exist`);
                 }
 
-                if (!scope.includes('user') || !pipeline.private) {
+                if (!pipeline.scmRepo.private) {
                     return true;
                 }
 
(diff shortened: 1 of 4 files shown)


plugins/pipelines/index.js
```diff
@@ -88,6 +89,54 @@ const pipelinesPlugin = {
...
+            if (credentials.scope.includes('admin')) {
+                return Promise.resolve(true);
+            }
+
+            return pipelineFactory.get(pipelineId).then(pipeline => {
+                if (!pipeline) {
+                    throw boom.notFound(`Pipeline ${pipelineId} does not exist`);
+                }
+
+                if (!scope.includes('user') || !pipeline.private) {
```
tkyi, line 118: what if scope is pipeline?
do we need to add a check like this: https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/secrets/index.js#L61

tkyi: could be a different pipeline's token?

tkyi (approved):

PR #2444, opened 2021-05-12
fix(466): allow user to set pipeline to public

klu909:
## Context

there could be situation where user need pipeline to stay public despite having a private repo
## Objective

allow user to set pipeline to public
## References

## License

### DIFF
--- a/plugins/pipelines/index.js
+++ b/plugins/pipelines/index.js
@@ -115,7 +115,7 @@ const pipelinesPlugin = {
                     throw boom.notFound(`Pipeline ${pipelineId} does not exist`);
                 }
 
-                if (!pipeline.scmRepo || !pipeline.scmRepo.private) {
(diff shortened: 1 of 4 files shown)


jithine (approved):

plugins/pipelines/index.js
```diff
@@ -115,7 +115,7 @@ const pipelinesPlugin = {
                     throw boom.notFound(`Pipeline ${pipelineId} does not exist`);
                 }
 
-                if (!pipeline.scmRepo || !pipeline.scmRepo.private) {
+                if (!pipeline.scmRepo || !pipeline.scmRepo.private || (pipeline.settings && pipeline.settings.public)) {
```
jithine, line 118: can you add unit test which covers this case, looks like coverage decreased after this change

Issue #2437, opened 2021-05-03
Support Read-Only SCM instance


jithine:
**What happened**:

For an enterprise organization it's desirable to create read-only mirrors of public Open source projects to reduce external internet access. This leads to a use case where users would want to create Screwdriver build pipelines when the source code SCM instance is read-only. Users can use External Config feature where build source code is in a separate SCM instance where usual development happens for the Oranization, and source code for the public Open Source projects exist in the read-only instance.

**What you expected to happen**:

Screwdriver should support a read only SCM instance. 

   1. Users should not have an option to login, since this is a read-only instance. 

When creating External Config pipeline (and only using this feature), child pipelines can be from this read-only SCM

  1. Since SCM itself is to be treated as read-only, there won't be any login user credentials we can use for making API calls to the SCM.
  2. We should use a headless user for handling access to SCM  API calls. This SCM token should be obtained from SCM admnistrator and only Screwdriver API has access to this token.
  3. When creating a child pipeine, if SD API determines it belongs to the read-only SCM, it should use the headless users access token for all API calls to the SCM.

SD API already uses a notion [of headless SCM user](https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/webhooks/index.js#L648-L652) for certain API calls. We need to extend that model to support all API calls for this read-only SCM

Things to keep in mind:
- difference between *user scm* (where the user is logged into) vs *pipeline scm* (where the repo lives)
- webhooks

**How to reproduce it**:

N/A

tkyi:
https://beta.cd.screwdriver.cd/pipelines/9276/child-pipelines
<img width="1128" alt="Screen Shot 2021-06-29 at 1 38 51 PM" src="https://user-images.githubusercontent.com/3230529/123864243-62b0e980-d8df-11eb-85b7-bf86dd4bb390.png">


Available with following versions:
- screwdrivercd/screwdriver: v4.1.120
- screwdrivercd/ui: v1.0.644

See https://docs.screwdriver.cd/cluster-management/configure-api#source-control-plugin for setup

PR #2460, opened 2021-05-25
feat(2437): Support read-only SCM [6]

tkyi:
## Context

Users should be able to make changes to read-only pipelines. Instead of checking permission against a child pipeline, can check against parent.

## Objective

This PR:
- updates SCM config with read-only config
- passes in `readOnly` info to `auth/contexts`
- changes permissions calls to get parent permissions for read-only scm pipelines

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/2437

## License


plugins/events/create.js
```diff
@@ -19,270 +20,216 @@ module.exports = () => ({
...
+                return boom.notFound('No jobs to start.');
+            }
+            // everything succeeded, inform the user
+            const location = urlLib.format({
+                host: request.headers.host,
+                port: request.headers.port,
+                protocol: request.server.info.protocol,
+                pathname: `${request.path}/${event.id}`
+            });
+
```
pritamstyz4ever, line 228: Refactoring into smaller functions would be nice and make it more readable. (can update later)

pritamstyz4ever (approved):
check if you need to mock ''helper.js' in tests

tkyi:
Shouldn't need to mock `helper.js` since all the underlying calls are mocked

plugins/events/create.js
tkyi: I'll save that for a future PR

Issue #2404, opened 2021-04-07
When using chainPR, PR_BRANCH_NAME set undefined


wahapo:
**What happened**:
In a PR build started with set `chainPR: rue` and chained, if the branch name includes a slash, `$PR_BRANCH_NAME` will be `origin/undefined`.

**What you expected to happen**:
`$PR_BRANCH_NAME` should be something like `origin/foo/bar`.

**How to reproduce it**:
1. Create screwdriver.yaml with `chainPR: true`.
    ```
    annotations:
      screwdriver.cd/chainPR: true

    jobs:
      job-1:
        requires: ~pr
        image: node:14
        steps:
          - echo ${PR_BRANCH_NAME}
      job-2:
        requires: job-1
        image: node:14
        steps:
          - echo ${PR_BRANCH_NAME}
    ```
2. Create a branch including a slash (e.g. `foo/bar`)
3. Run the PR build.
4. In [job-1](https://cd.screwdriver.cd/pipelines/6923/builds/693580/steps/step-1), `${PR_BRANCH_NAME}` is set to the correct value of `origin/foo/bar`.
5. Looking at the log of [job-2](https://cd.screwdriver.cd/pipelines/6923/builds/693581/steps/step-1), `${PR_BRANCH_NAME}` is `origin/undefined`.

PR #2464, opened 2021-05-27
fix(2404): Properly set prInfo.

kumada626:
## Context

In createInternalaBuild func, prInfo is not set properly.
I fixed the method creating pfInfo.

## Objective

In createInternalBuild func, prInfo is set from event object.

However, event has no prInfo property on its data-schema.
https://github.com/screwdriver-cd/data-schema/blob/32d1ef39ca80fbdb3bfbe3d9ddc6dbfdcf94698a/models/event.js#L75
https://github.com/screwdriver-cd/data-schema/blob/32d1ef39ca80fbdb3bfbe3d9ddc6dbfdcf94698a/core/scm.js#L94-L127
prInfo is used in the buildFactory.create for adding url and prBranchName info to event data.
https://github.com/screwdriver-cd/models/blob/d107819e0aadcbb6b31344b32729148c0165ce2a/lib/eventFactory.js#L692-L700
So I add url and prBranchName as prInfo from event object.

## References

#2404 

## License

### DIFF
--- a/plugins/builds/index.js
+++ b/plugins/builds/index.js
@@ -236,7 +236,12 @@ async function createInternalBuild(config) {
     } = config;
     const prRef = event.pr.ref ? event.pr.ref : '';
     const prSource = event.pr.prSource || '';
-    const prInfo = event.pr.prInfo || '';
+    const prInfo = event.pr.prBranchName
+        ? {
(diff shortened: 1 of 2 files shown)


tkyi (approved):
Thank you! 🙇‍♀️

Issue #2479, opened 2021-06-09
Display name validation fix


mchengat:
**What happened**:

Updating user name of user settings fails to validate in backend 

**What you expected to happen**:

Should fail to update the display user name if the length is more than 100 characters

**How to reproduce it**:

Set display name to above 99 in user settings. Value without proper validation get saved in database

PR #2480, opened 2021-06-09
fix(2479): display name length validation

mchengat:
## Context

Updating user name of user settings fails to validate in backend

## Objective

Fixing the validation

## References
https://github.com/screwdriver-cd/screwdriver/issues/2479

## License

### DIFF
--- a/plugins/users/settings/update.js
+++ b/plugins/users/settings/update.js
@@ -2,7 +2,6 @@
 
 const boom = require('@hapi/boom');
 const schema = require('screwdriver-data-schema');
-const updateSchema = schema.models.user.base.extract('settings');
 
 module.exports = () => ({
     method: 'PUT',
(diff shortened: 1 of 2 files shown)


tkyi (approved):

tkyi:
@jithine generally we use the pattern `PUT /model/:id` etc; like `/pipelines/:id`, `/jobs/:id`, `/events/:id`
would it be `PUT /users/:id`?
I also thought it would be more clear with `PUT /users/settings` that only `settings` can be updated, not the `username` or  `token` or anything else

jithine:
> @jithine generally we use the pattern PUT /model/:id etc; like /pipelines/:id, /jobs/:id, /events/:id
would it be PUT /users/:id?
I also thought it would be more clear with PUT /users/settings that only settings can be updated, not the username or token or anything else

You are right it's specifically picking up `settings` from input  https://github.com/screwdriver-cd/screwdriver/pull/2480/files#diff-2a289cd9c02e4aef3f319333c70d5cb587fd9670e293185b200b6e864df66b2fL22 Missed that.

Issue #2096, opened 2020-05-27
QUEUED notification is sent twice


kumada626:
**What happened**:
After we start to use queue-service, QUEUED notifications are sent twice per one build.
![queued_sent_twice](https://user-images.githubusercontent.com/8113204/82965593-7e263280-a003-11ea-8732-afcdbd26a729.jpg)

**What you expected to happen**:
QUEUED notification is sent only one time per one build.

**How to reproduce it**:
Using queue-service,
1. Setting `QUEUED` notifications on a build.
2. A build is queued.
3. `QUEUED` notifications are sent twice.

jithine:
cc @pritamstyz4ever did we create an issue already ? We've seen this before.

yoshwata:
# Cause

I analysed the cause of this issue.

![Screen Shot 2020-08-06 at 14 42 24](https://user-images.githubusercontent.com/32473622/89620585-d5316900-d8ca-11ea-8e4c-071bb6180199.jpg)

1. SDAPI accesses to queueService's [/queue/message](https://github.com/screwdriver-cd/executor-queue/blob/ccedf9e2f69317af9b5c1074bbd1c98344edde20/index.js#L155)
1. queueService [enqueues a build](https://github.com/screwdriver-cd/queue-service/blob/938d3fc5b41c675257806812cac65ecd3e1d0c36/plugins/queue/scheduler.js#L376) to Redis
1. queueService [accesses](https://github.com/screwdriver-cd/queue-service/blob/938d3fc5b41c675257806812cac65ecd3e1d0c36/plugins/queue/scheduler.js#L384) to  [SDAPI's build/{id}](https://github.com/screwdriver-cd/screwdriver/blob/5e60269e36c3bcf3da78a7935f438d922137b7b3/plugins/builds/update.js#L12) for updating build's stats([queueEnterTime](https://github.com/screwdriver-cd/queue-service/blob/938d3fc5b41c675257806812cac65ecd3e1d0c36/plugins/queue/scheduler.js#L285)).
1. queueService [gets a build queue](https://github.com/screwdriver-cd/queue-service/blob/938d3fc5b41c675257806812cac65ecd3e1d0c36/plugins/worker/lib/jobs.js#L155) while it polls Redis.
[...]

PR #2512, opened 2021-07-26
[1] fix(2096): Only emits build_status when status is passed.

kumada626:
## Context

Build status notifications should better to be sent only status is changed.
Therefore, only emits notification event when update status is set.

## Objective

Only emits build_status when status is passed.

## References

#2096 

## License

### DIFF
--- a/plugins/builds/update.js
+++ b/plugins/builds/update.js
@@ -180,6 +180,7 @@ module.exports = () => ({
             }
 
             const build = await getBuildToUpdate(id, buildFactory);
+            const currentStatus = build.status;
 
             if (!isBuild) {
                 await validateUserPermission(build, request);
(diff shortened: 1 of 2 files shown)


wahapo (approved):

plugins/builds/update.js
```diff
@@ -204,11 +205,10 @@ module.exports = () => ({
                 build.stats = Object.assign(build.stats, {
                     blockedStartTime: new Date().toISOString()
                 });
-            } else if (desiredStatus === 'QUEUED') {
+            } else if (desiredStatus === 'QUEUED' && currentStatus !== 'QUEUED') {
```
tkyi, line 208: should we skip `build.update()` on line 225 if `desiredStatus` and `currentStatus` are `'QUEUED'`?

wahapo: Maybe my understanding is that it could be updated by the value of `build.stats` passed from [queue-service](https://github.com/screwdriver-cd/queue-service/pull/31/files#diff-ea1f58226f3c81f477d5ed24aacecc91344e9c00c9a9540e080b39e73e06ddcdR399) in line 192.

tkyi: ah okay, so we still need to update

tkyi (approved):

PR #2518, opened 2021-07-30
fix(2484): UI artifact is slow for large html file

klu909:
## License


plugins/builds/artifacts/get.js
```diff
@@ -61,7 +63,24 @@ module.exports = config => ({
...
+                    let response = h.response(gotStream);
+
+                    if (request.query.type === 'download') {
+                        response.headers['content-type'] = 'application/octet-stream';
+                        response.headers['content-disposition'] = `attachment; filename="${encodeURI(encodedArtifact)}"`;
+                    } else if (request.query.type === 'preview') {
+                        const fileExt = encodedArtifact.split('.').pop();
+                        const mime = getMimeFromFileName(fileExt, encodedArtifact);
+
+                        response.headers['content-type'] = mime;
```
jithine, line 76: why are we computing these again at API ? Store is already doing this and setting appropriate headers. API  should be reading these data from store webservice response headers.

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "API server for the Screwdriver.cd service",
   "main": "index.js",
   "scripts": {
-    "pretest": "eslint --max-warnings 90 .",
```
jithine, line 7: why is this changed ?

PR #2537, opened 2021-08-23
feat(1961): Replace deprecated request package with screwdriver-request

tkyi:
## Context

It would be nice to replace deprecated request package with a different one.

## Objective

This PR attempts to replace [request package](https://www.npmjs.com/package/request) with [got package](https://github.com/sindresorhus/got) using screwdriver-request.

Note: `followAllRedirects` does not exist in got package, but `followRedirect` is turned on by default for GET requests (https://github.com/sindresorhus/got/blob/main/documentation/2-options.md#followredirect)

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/1961

## License

I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #2565, opened 2021-09-09
fix: Logs should not throw when store returns 404

tkyi:
## Context

Getting errors with [getMaxLines](https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/builds/steps/logs.js#L35) because got throws error for 404:
```
{"name":"HTTPError","timings":{"start":1631133820277,"socket":1631133820278,"lookup":1631133820313,"connect":1631133820316,"secureConnect":1631133820321,"upload":1631133820321,"response":1631133820379,"end":1631133820381,"phases":{"wait":1,"dns":35,"tcp":3,"tls":5,"request":0,"firstByte":58,"download":2,"total":104}},"level":"error","timestamp":"2021-09-08T20:43:40.382Z"}
210908/204340.383, (1631133820241:sdapi-beta-6f49c864b-hrmnd:18:ktbywcvt:10022) [request,server,error] data: Error: HTTPError: Response code 404 (Not Found)
    at getMaxLines (/usr/src/app/node_modules/screwdriver-api/plugins/builds/steps/logs.js:72:15)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
210908/204340.241, (1631133820241:sdapi-beta-6f49c864b-hrmnd:18:ktbywcvt:10022) 
```
## Objective

This PR does not throw when store returns 404.

## References

Related to https://github.com/screwdriver-cd/screwdriver/pull/2537

## License

### DIFF
--- a/plugins/builds/steps/logs.js
+++ b/plugins/builds/steps/logs.js
@@ -10,6 +10,7 @@ const uuid = require('uuid');
 
 const MAX_LINES_SMALL = 100;
 const MAX_LINES_BIG = 1000;
+const EOF_MESSAGE = 'Response code 404 (Not Found)';
 
 /**
  * Makes the request to the Store to get lines from a log
@@ -32,7 +33,13 @@ async function fetchLog({ baseUrl, linesFrom, authToken, page, sort }) {
 
     return new Promise((resolve, reject) => {
         requestStream
-            .on('error', e => reject(e))
+            .on('error', e => {
+                if (!e.message.includes(EOF_MESSAGE)) {
+                    return reject(e);
+                }
+
+                return resolve([]);
+            })
             // Parse the ndjson
             .pipe(ndjson.parse({ strict: false }))
             // Only save lines that we care about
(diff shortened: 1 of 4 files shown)


plugins/builds/steps/logs.js
```diff
@@ -32,7 +33,13 @@ async function fetchLog({ baseUrl, linesFrom, authToken, page, sort }) {
 
     return new Promise((resolve, reject) => {
         requestStream
-            .on('error', e => reject(e))
+            .on('error', e => {
+                if (!e.message.includes(EOF_MESSAGE)) {
```
jithine, line 37: look for explicit status code like this https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/builds/artifacts/get.js#L70
rather than string matching error message

tkyi: Yeah I originally tried that but there's no status code

jithine: hmm, that sounds like it's a caught exception and not the original exception from `got` ?
This might be why https://github.com/screwdriver-cd/request/blob/master/index.js#L56 It's not the original got response. So https://github.com/screwdriver-cd/request/blob/master/index.js#L56 also needs to be fixed to use just 

https://github.com/screwdriver-cd/request/blob/master/index.js#L16

jithine (approved):

Issue #1500, opened 2019-02-15
Make SD_ZIP_ARTIFACTS work for large content sizes


jithine:
**What happened**:

Current implementation of SD_ZIP_ARTIFACTS depends on running a AWS Lambda. But this brings in restriction on the size of data that can be unzipped, since the process is dependent on constraints like time & memory imposed by LAMBDA. 

**What you expected to happen**:

SD_ZIP_ARTIFACTS to always work. We can achieve this by leveraging on Screwdriver's internal queue and processing this on our servers instead of AWS Lambda.

**How to reproduce it**:

Use SD_ZIP_ARTIFACTS on a build with very large number of files. While uploading itself is fast, they don't appear in Artifacts section of the build.

kkisic:
Since S3 lambda is not available in our environment, we(YJ team) would like to proceed with the change to a configuration using queue-service too.

yuichi10:
I wrote design for this Issue.
https://github.com/screwdriver-cd/screwdriver/pull/2583

kkisic:
### Remaining Tasks
- Fix queue-service
  - in-progress: https://github.com/screwdriver-cd/queue-service/pull/37
- Implement unzip-worker
- Fix api
- Fix artifact-bookend

s-yoshika:
### Remaining Tasks (UPDATED 2021/12/20)
- [x] 1. Queue Service can accept `unzip` requests from API and enqueue the `unzip` job to `Resque(Redis)`
  - completed: https://github.com/screwdriver-cd/queue-service/pull/37
- [x] 2. Store can accept get/put requests for `artifacts` from `unzip-worker`
  - completed: https://github.com/screwdriver-cd/store/pull/130
- [x] 3. Unzip worker should dequeue `unzip` jobs and get the ZIP for the build from Store, unzip it and re-upload the unzipped artifacts to Store
  - completed: https://github.com/screwdriver-cd/artifacts-unzip-worker/pull/1 
- [ ] 4. API accepts `unzip` requests from `artifacts-bookend` in a build and requests for Queue Service to `unzip`.
  - in-progress:
    - https://github.com/screwdriver-cd/executor-base/pull/75
    - https://github.com/screwdriver-cd/executor-queue/pull/97
    - https://github.com/screwdriver-cd/models/pull/520
    - https://github.com/screwdriver-cd/screwdriver/pull/2616
- [ ] 5. `artifact-bookend` step should request for API to `unzip`
[...]

PR #2583, opened 2021-10-14
design(1500) sd zip artifacts design

yuichi10:
## Context

To provide the SD_ZIP_ARTIFACTS function by only using Screwdriver components, I create a design.

## Objective

- Proposal to implements SD_ZIP_ARTIFACTS by Screwdriver components.

## References

https://github.com/screwdriver-cd/screwdriver/issues/1500

## License

### DIFF
--- a/features/step_definitions/git-flow.js
+++ b/features/step_definitions/git-flow.js
@@ -213,7 +213,7 @@ When(
     },
     function step() {
         const { branch } = this;
-        const { tag } = this;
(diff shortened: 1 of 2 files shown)


yuichi10:
This is the token of the communication from the consumer (unzip worker/consumer) to the Store.
It is because the consumer needs to download and upload files to Store.
I thought it would be better to create a token for that consumer in the API, so I designed it this way.

PR #2594, opened 2021-11-04
[2] fix(1500): Add functional test about artifacts

sugarnaoming:
## Context
With the reimplementation of SD_ZIP_ARTIFACTS, we will add a functional test for artifacts.

## Objective
Add a scenario when SD_ZIP_ARTIFACTS is disabled and enabled

## References
#1500
https://github.com/screwdriver-cd-test/functional-artifacts/pull/1

## License

### DIFF
--- a/features/artifacts.feature
+++ b/features/artifacts.feature
@@ -24,6 +24,6 @@ Feature: Artifacts
     Scenario: Verify that the artifacts have been saved with zipping enabled.
         And "calvin" is logged in
         When the "zipped" job is started
-        Then an event is createdq
+        Then an event is created
         And the "zipped" build succeeds
         And artifacts were found in the build
--- a/features/artifacts.feature
+++ b/features/artifacts.feature
@@ -26,4 +26,4 @@ Feature: Artifacts
         When the "ziped" job is started
         Then an event is created
         And the "ziped" build succeeds
-        And artifacts were found in the build with the same event ID as the successful main job
+        And artifacts were found in the build with the same event ID as the successful ziped job
--- a/features/artifacts.feature
+++ b/features/artifacts.feature
@@ -1,0 +1,29 @@
+@artifacts
+Feature: Artifacts
+
+    Screwdriver allows you to view and download the artifacts generated during the build process at a later time.
+    Enabling zip compression when saving your artifacts may speed up the build.
+
+    Rules:
+    -　The user must be able to select and review all or some of the artifacts generated during the build
+
+    Background:
+        Given an existing pipeline with the workflow:
+            | job | triggers |
+            | main | ~commit |
+            | ziped | ~commit |
+        And "calvin" has admin permission to the pipeline
+
+    Scenario: Verify that artifacts have been saved
+        And "calvin" is logged in
+        When the "main" job is started
+        Then an event is created
+        And the "main" build succeeds
+        And artifacts were found in the build with the same event ID as the successful main job
+
+    Scenario: Verify that the artifacts have been saved with ziping enabled.
+        And "calvin" is logged in
+        When the "ziped" job is started
+        Then an event is created
+        And the "ziped" build succeeds
+        And artifacts were found in the build with the same event ID as the successful main job
--- a/features/artifacts.feature
+++ b/features/artifacts.feature
@@ -19,11 +19,11 @@ Feature: Artifacts
         When the "main" job is started
         Then an event is created
         And the "main" build succeeds
-        And Found artifacts in the "main" job that were successfully build
+        And artifacts were found in the build
 
     Scenario: Verify that the artifacts have been saved with zipping enabled.
         And "calvin" is logged in
         When the "zipped" job is started
         Then an event is createdq
         And the "zipped" build succeeds
-        And Found artifacts in the "zipped" job that were successfully build
+        And artifacts were found in the build
--- a/features/artifacts.feature
+++ b/features/artifacts.feature
@@ -11,7 +11,7 @@ Feature: Artifacts
         Given an existing pipeline with the workflow:
             | job | triggers |
             | main | ~commit |
-            | ziped | ~commit |
+            | zipped | ~commit |
         And "calvin" has admin permission to the pipeline
 
     Scenario: Verify that artifacts have been saved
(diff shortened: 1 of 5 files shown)


sugarnaoming:
The results of the lint run locally do not match the results of the lint run in the main job at all.
It shows that I need to add parentheses if I run it locally. And I believe the local point is correct.
image
In addition to the parentheses problem, there was also a contradiction between the ESLint and prettier settings for `trailingComma` . This is temporarily fixed by adding `.prettierrc.yaml` and setting the `trailingComma: none`.

These lint issues are not part of this PR, so I won't fix them here.

features/artifacts.feature
```diff
@@ -0,0 +1,29 @@
...
+    Enabling zip compression when saving your artifacts may speed up the build.
+
+    Rules:
+    -　The user must be able to select and review all or some of the artifacts generated during the build
+
+    Background:
+        Given an existing pipeline with the workflow:
+            | job | triggers |
+            | main | ~commit |
+            | ziped | ~commit |
```
s-yoshika, line 14: zipped?

features/artifacts.feature
```diff
@@ -0,0 +1,29 @@
...
+        And "calvin" has admin permission to the pipeline
+
+    Scenario: Verify that artifacts have been saved
+        And "calvin" is logged in
+        When the "main" job is started
+        Then an event is created
+        And the "main" build succeeds
+        And artifacts were found in the build with the same event ID as the successful main job
+
+    Scenario: Verify that the artifacts have been saved with ziping enabled.
```
s-yoshika, line 24: zipping?

features/artifacts.feature
```diff
@@ -0,0 +1,29 @@
...
+        And "calvin" is logged in
+        When the "ziped" job is started
+        Then an event is created
+        And the "ziped" build succeeds
+        And artifacts were found in the build with the same event ID as the successful main job
```
s-yoshika, line 29: Is it `ziped` job instead of `main` job?
```suggestion
        And artifacts were found in the build with the same event ID as the successful ziped job
```

features/step_definitions/artifacts.js
```diff
@@ -0,0 +1,99 @@
...
+
+Then(/^the "ziped" build succeeds$/, { timeout: TIMEOUT }, function step() {
+    return this.waitForBuild(this.buildId).then((resp) => {
+        Assert.equal(resp.body.status, 'SUCCESS');
+        Assert.equal(resp.statusCode, 200);
+    });
+});
+
+Then(
+    /^artifacts were found in the build with the same event ID as the successful main job$/,
```
s-yoshika, line 58: Why event ID? The actual codes seem not to check the event ID.

features/artifacts.feature
s-yoshika: Do we need to disable this test case until the feature is implemented?

features/artifacts.feature
sugarnaoming: thx.
it's a typo

features/artifacts.feature
sugarnaoming: Same here.

features/artifacts.feature
sugarnaoming: thx.

features/step_definitions/artifacts.js
sugarnaoming: It was a leftover from when I ported it over from `event.js`.
I haven't checked the event id, so I'll fix the wording.

features/artifacts.feature
sugarnaoming: No, it's not.
This test should work regardless of the implementation of the SD_ZIP_ARTIFACTS function, so it should be fine to leave the test enabled.

s-yoshika (dismissed):
LGTM

features/artifacts.feature
```diff
@@ -0,0 +1,29 @@
...
+        And "calvin" is logged in
+        When the "main" job is started
+        Then an event is created
+        And the "main" build succeeds
+        And artifacts were found in the build
+
+    Scenario: Verify that the artifacts have been saved with zipping enabled.
+        And "calvin" is logged in
+        When the "zipped" job is started
+        Then an event is createdq
```
yuichi10, line 27:
```suggestion
        Then an event is created
```

features/step_definitions/artifacts.js
```diff
@@ -0,0 +1,95 @@
...
+    this.repoName = 'functional-artifacts';
+
+    // Reset shared information
+    this.buildId = null;
+    this.pipelineId = null;
+    this.eventId = null;
+    this.jwt = null;
+});
+
+When(/^the "zipped" job is started$/, { timeout: TIMEOUT }, function step() {
```
yuichi10, line 19: Perhaps, is it possible to aggregate this function to /^the "main" job is started$/ scenario?
This function looks very similar to the scenario.

features/step_definitions/artifacts.js
```diff
@@ -0,0 +1,95 @@
...
+});
+
+Then(/^artifacts were found in the build$/, { timeout: TIMEOUT }, function step() {
+    const artifactName1 = 'sample1.txt';
+    const artifactName2 = 'sample2.txt';
+
+    const retryConfig = {
+        limit: 6,
+        statusCodes: [408, 404, 413, 429, 500, 502, 503, 504, 521, 522, 524],
+        calculateDelay: ({ computedValue }) => {
```
yuichi10, line 64: Is it better to increase the delay time based on the number of attempts?
If the request fails, you should always wait 30 seconds.

sugarnaoming: I was mistaken about the setting method.
It is now fixed to retry at 30-second intervals.

sugarnaoming: Fixed the delay time to increase with the number of retries.
The maximum delay time is 30 seconds.

features/step_definitions/artifacts.js
```diff
@@ -0,0 +1,98 @@
...
+    const artifactName2 = 'sample2.txt';
+    const maxRetryInterval = 30000;
+
+    const retryConfig = {
+        limit: 6,
+        statusCodes: [408, 404, 413, 429, 500, 502, 503, 504, 521, 522, 524],
+        calculateDelay: ({ attemptCount }) => {
+            // 1st time: 5sec, 2nd time: 10sec...
+            // max delay is 30sec.
+            return attemptCount * 50000;
```
yuichi10, line 68:
```suggestion
            return attemptCount * 5000;
```

yuichi10 (approved):
LGTM

ibu1224 (approved):

tkyi (approved):

Issue #1956, opened 2020-02-07
Event builds API should not send step information


jithine:
**What happened**:

Pipeline detail page calls `GET /event/ID/builds` for each of the event. This API also fetches steps information for each of the build which is not required. Only build detail page should request for build steps. 

**What you expected to happen**:

1. Events API end point should not return step information. Make it optional.
1. UI should handle any logic involved where builds may not have step information.

**How to reproduce it**:

Visit pipeline page for any event and observe API calls to event builds end point.

s-yoshika:
`GET /jobs/ID/builds` which is called in PR tab on pipeline page also returns steps information. Could we make this endpoint optional as well?

tkyi:
Made backend changes to optionally skip fetching steps, but requires UI changes to complete. Putting back into backlog.

PR #2619, opened 2021-12-17
fix(1956): Add option to omit step info when fetching builds for event and job API

tkyi:
## Context

Some endpoints don't need to fetch step info for builds.

## Objective

This PR adds option to set query param `steps=false` if a user wants to skip fetching steps alongside build info for certain endpoints. Also uses readOnly DB to get build info.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/1956

## License


plugins/events/listBuilds.js
```diff
@@ -24,14 +24,23 @@ module.exports = () => ({
...
             const event = await eventFactory.get(request.params.id);
+            const { steps } = request.query;
 
             if (!event) {
                 throw boom.notFound('Event does not exist');
             }
 
-            const buildsModel = await event.getBuilds();
+            const buildsModel = await event.getBuilds({
+                readOnly: true
```
jithine, line 34: do we need to hit `readOnly` DB instance ? shouldn't build data be always shown up to date ?

jithine (approved):

PR #2641, opened 2022-01-21
fix(2640): errors in individual event processing should not affect rest of event creation

jithine:
## Context

Ensure that all individual event creations gets processed. `Promise.all` fails on first failure, however we do want to ensure that all events are processed. We might have to look and fix similar patterns across codebase. 

## Objective

1. Use `Promise.allSettled` to ensure that all Promises are processed. 
2. Log errors for event creation failure

## License

### DIFF
--- a/plugins/webhooks/helper.js
+++ b/plugins/webhooks/helper.js
@@ -348,6 +348,37 @@ async function triggeredPipelines(
 }
 
 /**
+ * Start Events
+ * @async  startEvents
+ * @param  {Array}  eventConfigs Array of event config objects
+ * @param  {Object} eventFactory Factory to create events
+ * @return {Promise<Array>}  Array of created events
+ */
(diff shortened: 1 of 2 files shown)


plugins/webhooks/helper.js
```diff
@@ -347,6 +347,47 @@ async function triggeredPipelines(
...
+            if (result.value) events.push(result.value);
+        } else {
+            errorCount += 1;
+            logger.error(`pipeline:${eventConfigs[i].pipelineId} error in starting event`, result.value);
+        }
+    });
+
+    if (errorCount && errorCount === eventsCount) {
+        // preserve current behavior of returning 500 on error
+        throw new Error('Failed to start eny events');
```
tkyi, line 385:
```suggestion
        throw new Error('Failed to start any events');
```

tkyi (approved):

PR #2655, opened 2022-02-03
fix: return 403 or 404 error when mismatch scm host

y-oksaku:
## Context

When scm host that user logged in does not match pipeline's scm host, the 500 error is returned.
However, since this is an authorization error caused by users, the 403(or 404 when private) error should be returned.

## Objective

- Add error handling for [this error](https://github.com/screwdriver-cd/scm-github/blob/a12f1e0bd9bd24b1d03c55cb6dbfe220d109f347/index.js#L254-L257).

## References

- https://github.com/screwdriver-cd/scm-github/pull/202

## License

### DIFF
--- a/plugins/events/create.js
+++ b/plugins/events/create.js
@@ -131,10 +131,7 @@ module.exports = () => ({
             try {
                 permissions = await user.getPermissions(scmUri);
             } catch (err) {
-                const statusCode = (!pipeline.scmRepo || !pipeline.scmRepo.private) ? 403 : 404;
-                throw boom.boomify(err, {
-                    statusCode: err.status || statusCode
-                });
+                throw boom.boomify(err, { statusCode: err.statusCode });
             }
 
             // Update admins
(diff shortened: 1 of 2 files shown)


yoshwata (dismissed):

plugins/events/create.js
```diff
@@ -126,7 +126,16 @@ module.exports = () => ({
...
             // Check the user's permission
-            const permissions = await user.getPermissions(scmUri);
+            let permissions;
+
+            try {
+                permissions = await user.getPermissions(scmUri);
+            } catch (err) {
+                const statusCode = (!pipeline.scmRepo || !pipeline.scmRepo.private) ? 403 : 404;
+                throw boom.boomify(err, {
+                    statusCode: err.status || statusCode
```
y-oksaku, line 136: If we catch an error such as `TypeError, CircuitBreaker error` here, the response code will be 403 or 404.
This is bad.

tkyi (commented):
Will this cause issues with this feature? https://github.com/screwdriver-cd/screwdriver/issues/2437

kumada626:
I think this will not effect on #2437.
Since read-only scm's pipeline uses parent pipeline's scm uri for checking user permission, users have to login parent pipeline's scm even without this PR.
The purpose of this PR is only to correctly return error response codes.

wahapo (approved):

itleigns (approved):

tkyi (approved):

PR #2670, opened 2022-02-28
fix(1393) add a scenario into event.feature for the case of parent event

s-yoshika:
## Context
The current feature test for events doesn't have the scenario for restarting job from the existing event, but this is the common use case.

## Objective
This PR add a test scenario for the event that has a parentEventId into `event.feature` in functional-tests.

## References

https://github.com/screwdriver-cd/screwdriver/issues/1393

## License

### DIFF
--- a/features/step_definitions/events.js
+++ b/features/step_definitions/events.js
@@ -1,8 +1,9 @@
 'use strict';
 
 const Assert = require('chai').assert;
-const { Before, Given, Then } = require('cucumber');
+const { Before, Given, When, Then } = require('cucumber');
 const request = require('screwdriver-request');
+const sdapi = require('../support/sdapi');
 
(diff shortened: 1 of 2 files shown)


features/step_definitions/events.js
```diff
@@ -21,6 +23,65 @@ Given(/^an existing pipeline with the workflow:$/, { timeout: TIMEOUT }, functio
...
+        context: {
+            token: this.jwt
+        }
+    })
+        .then(resp => {
+            Assert.equal(resp.statusCode, 201);
+            this.previousEventId = resp.body.id;
+        })
+        .then(() =>
+            sdapi.cleanupBuilds({
```
s-yoshika, line 44: No need to wait for the build of the parent event to succeed, so we can abort it.

tkyi (approved):

Issue #2665, opened 2022-02-16
Bug when the first directory of the source directory is two characters or less


yoshwata:
**What happened**:

If the first directory of the source directory is set to two characters or less, the first directory will be deleted from settings.

For example, if you make the following settings:
- ci/sd/realtime
- sd/realtime
- pa/to
- p/t

It will be interpreted as follows.
- sd/realtime
- realtime
- to
- t

After that, the resetting will result in the following error.

![スクリーンショット 2022-02-16 11 21 39](https://user-images.githubusercontent.com/32473622/154185291-e71854cd-e411-427c-97f6-d32ebfa585c2.png)

**What you expected to happen**:

Even if the first directory is two characters or less, the source directory is interpreted correctly as set.
Resetting also works in the same way

**How to reproduce it**:

I reproduced it in the following pipeline.
https://cd.screwdriver.cd/pipelines/8718/events

**Question**:

The following code formats the settings of the source directory. Is this intended to remove the leading `/`, `./`, `../`?
https://github.com/screwdriver-cd/screwdriver/blob/a746e5718dba2fa2c4abc687cc55fa631c0e7fa4/plugins/pipelines/helper.js#L40

tkyi:
Yeah I think originally we remove leading `/` and `./` so I think I thought `../` made sense to remove too.
https://github.com/screwdriver-cd/screwdriver/pull/2600/files#diff-b64239ecc01fa5d69099c49a4a9f7ec4d8541b8cbc082840a08392e5426d44fd

What do you think? Is there a better way to handle this?

yoshwata:
> What do you think? Is there a better way to handle this?

It's an idea, but maybe it should be escaping `.`
```
> "aa/bb/cc".replace(/^(\/+|\.\/|\.\.\/)|\/+$/g, '');
'aa/bb/cc'
> "./aa/bb/cc".replace(/^(\/+|\.\/|\.\.\/)|\/+$/g, '');
'aa/bb/cc'
> "../aa/bb/cc".replace(/^(\/+|\.\/|\.\.\/)|\/+$/g, '');
'aa/bb/cc'
> "/aa/bb/cc".replace(/^(\/+|\.\/|\.\.\/)|\/+$/g, '');
'aa/bb/cc'
```

PR #2684, opened 2022-03-17
fix(2665): Enables to set sourceDirectory even if top dir was two characters or less

yoshwata:
## Context

Fix https://github.com/screwdriver-cd/screwdriver/issues/2665

## Objective

- This PR enables to create pipeline even if two characters or less top dir was set as rootDir like `aa/bb/cc`.
- This PR enables to change pipeline's rootDir even if two characters or less top dir was set.
- `/`, `./`, `../` are omitted as [originally intended](https://github.com/screwdriver-cd/screwdriver/issues/2665#issuecomment-1041982316).

## References

https://github.com/screwdriver-cd/screwdriver/issues/2665

## License


tkyi (approved):
Thank you for the fix 🙏

PR #2691, opened 2022-04-05
fix: Return 404 error when github api returns 404

sakka2:
## Context

Screwdriver API returns 500 error when github api returns 404

## Objective

Screwdriver API also returns 404 error when github api returns 404

## License

### DIFF
--- a/plugins/events/create.js
+++ b/plugins/events/create.js
@@ -200,8 +200,10 @@ module.exports = () => ({
                 // User has good permissions, create an event
                 sha = await scm.getCommitSha(scmConfig);
             } catch (err) {
-                if (err.statusCode) {
-                    throw boom.boomify(err, { statusCode: err.statusCode });
+                if (err.status && err.status === 404) {
+                    throw boom.notFound(err.message);
+                } else {
+                    throw boom.boomify(err, { statusCode: err.status });
                 }
             }
 
--- a/plugins/events/stopBuilds.js
+++ b/plugins/events/stopBuilds.js
@@ -46,14 +46,14 @@ module.exports = () => ({
             }
 
             // Check permissions
-            const permissions = await user.getPermissions(pipeline.scmUri);
-                .catch(err => {
-                    if (err.status === 404) {
-                        throw boom.notFound(`Repository ${pipeline.scmUri} does not exist`);
-                    } else {
-                        throw err;
-                    }
-                });
+            const permissions = await user.getPermissions(pipeline.scmUri)
+            .catch(err => {
+                if (err.status === 404) {
+                    throw boom.notFound(`Repository ${pipeline.scmUri} does not exist`);
+                } else {
+                    throw err;
+                }
+            });
             const adminDetails = request.server.plugins.banners.screwdriverAdminDetails(username, scmContext);
             const isPrOwner = hoek.reach(event, 'commit.author.username') === username;
 
--- a/plugins/events/stopBuilds.js
+++ b/plugins/events/stopBuilds.js
@@ -47,6 +47,13 @@ module.exports = () => ({
 
             // Check permissions
             const permissions = await user.getPermissions(pipeline.scmUri);
+                .catch(err => {
+                    if (err.status === 404) {
+                        throw boom.notFound(`Repository ${pipeline.scmUri} does not exist`);
+                    } else {
+                        throw err;
+                    }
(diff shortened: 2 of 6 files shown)


plugins/webhooks/index.js
```diff
@@ -103,6 +104,9 @@ const webhooksPlugin = {
                     } catch (err) {
                         logger.error(`[${hookId}]: ${err}`);
 
+                        if (err.statusCode === 404) {
```
tk3fftk, line 107: it looks `err.status` and `err.statusCode` are mixed. Is this intended?

sakka2: It seems api returns both `status`, and `statusCode`. I will fix to use `statusCode`.
```
{
  status: 404,
  response: {
    url: 'xxxxxx'
    status: 404,
    headers: {
       **snip**
    },
    data: {
      message: 'Branch not found',
      documentation_url: 'https://docs.github.com/enterprise/3.2/rest/reference/repos#get-a-branch'
    }
  },
  request: {
    method: 'GET',
    url: 'xxxxxxx'
    headers: {
      accept: 'application/vnd.github.v3+json',
      'user-agent': 'octokit-rest.js/18.12.0 octokit-core.js/3.6.0 Node.js/12.22.11 (linux; x64)',
      authorization: 'token [REDACTED]'
    },
    request: { hook: [Function: bound bound register] }
  },
  statusCode: 404
}
```

yuichi10:
Do we need to add test for webhooks.test.js?

plugins/events/create.js
```diff
@@ -200,7 +200,9 @@ module.exports = () => ({
                 // User has good permissions, create an event
                 sha = await scm.getCommitSha(scmConfig);
             } catch (err) {
-                if (err.statusCode) {
+                if (err.statusCode && err.statusCode === 404) {
+                    throw boom.notFound(err.message);
+                } else {
```
yoshwata, line 205: Why do you need this change? If err.statusCode contains 404, boomify seems to treat it as 404.

plugins/events/stopBuilds.js
```diff
@@ -46,7 +46,13 @@ module.exports = () => ({
             }
 
             // Check permissions
-            const permissions = await user.getPermissions(pipeline.scmUri);
+            const permissions = await user.getPermissions(pipeline.scmUri).catch(err => {
+                if (err.statusCode === 404) {
+                    throw boom.notFound(`Repository ${pipeline.scmUri} does not exist`);
```
yoshwata, line 51: Using `throw boom.boomify(err, { statusCode: err.statusCode });` style might be better for throwing error other than 404 error.

yoshwata: I'll omit comments, but the same is true for other fixes.

plugins/events/create.js
sakka2: This error has occured only under my private circumstance. So, I deleted.

sakka2:
> Do we need to add test for webhooks.test.js?

Sorry, I deleted the modification of webhooks.

plugins/events/stopBuilds.js
sakka2: https://github.com/screwdriver-cd/screwdriver/blob/6ac28557524d0bbac61880e31c2402c91f3d7d98/plugins/pipelines/sync.js#L51

If we use `boom.boomify` style at the case above, error message will be `not found`. boomify seems to be able to override the message, but I think `boom.notFound` is easier.

In other cases, I fixed to use `boom.boomify` style.

yuichi10:
It looks  when create.js get error, it return the scm error as it is, however sync.js check the error is 404. 
Is it same reason of [this comment](https://github.com/screwdriver-cd/screwdriver/pull/2691#discussion_r845654447)?

plugins/events/stopBuilds.js
sakka2: I fixed to use `boom.boomify`.

sakka2:
> It looks when create.js get error, it return the scm error as it is, however sync.js check the error is 404. Is it same reason of [this comment](https://github.com/screwdriver-cd/screwdriver/pull/2691#discussion_r845654447)?

I fixed to use boom.boomify

yoshwata (approved):

yuichi10 (approved):

PR #2733, opened 2022-07-12
feat: add design for pipeline stage

adong:
## Context
Add initial design doc for Pipeline Stage

## License

### DIFF
--- a/design/pipeline-stage.md
+++ b/design/pipeline-stage.md
@@ -1,0 +1,191 @@
+## Status
+- July 11, 2022: Proposal submitted
+
+## Context
+
+Current screwdriver reacts and send user notification based on single job's status from given pipeline, and it lacks of way of reacting to a group of jobs status changes. For example, 
+
+```
+Pipeline1: 
+
+[jobA] --> [jobB, jobC]
+       \               
+        -> [jobD, jobE]
+```
+
+We can get notifaction from any individual jobs of `A,B,C,D,E`, but there's no way to get notification when both job B and C finishes, and there's no way to get notification when all jobs `A,B,C,D,E` finishes. Users can setup an "end job" to make sure that is the downstream job, i.e. `jobF`
+
+```
+Pipeline1: 
+
+[jobA] --> [jobB, jobC] -> [jobF]
+       \                /
+        -> [jobD, jobE]/
+```
+
+So, when `jobF` statu's changed, users will get a notification, which implys all jobs `A,B,C,D,E`'s statuses were updated.
+
+
+## Proposal
+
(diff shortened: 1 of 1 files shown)


design/pipeline-stage.md
```diff
@@ -0,0 +1,192 @@
...
+
+color: optional, a hex presentation of color value for this stage to be displayed on the UI, if not present, UI will present its color in the default [color sequence](NEED to update the link). 
+
+jobs: list of job names that belong to a stage; one job should not be in multiple stages
+
+## Details
+
+Stage Table
+
+| id |jobIds | name | eventId |pipelineId | Other metadata|
```
tkyi, line 98: I think we originally considered having field `state` with values `ARCHIVED`/`ACTIVE`

but with the event-based stages, I think it would be hard to update the right stage's `state` when it needs to be `ARCHIVED`

maybe we could just have a field `createTime` and the latest one will be considered the "active"/current one

design/pipeline-stage.md
```diff
@@ -0,0 +1,192 @@
...
+- Create one stage table row with corresponding eventId, 
+- if it's triggered from the latest commit, and check if there's a corresponding pipelineId with eventId to be `null`, 
+	- if it does not exist, create another row with eventId to be null to indicate this is the latest eventId for the given pipeline
+	- if it does exist, update the definition from that row to reflect the latest changes
+- if it was not triggered from the latest commit, meaning it was a start from an existing event, including the restart of a given event, then we create a new event with parentEvent set, set the definition of the stage from retrieving the old stage definition, no extra step needed for set the stages.
+	
+
+So, we have the following API exposed: 
+
+`pipelines/:id/stages`
```
tkyi, line 118: with my above DB ideas, this endpoint would be difficult to return; or it could return latest based on `createTime`
might also want to tweak the endpoint to have optional query param `pipelines/:id/stages?eventId=eventId`

design/pipeline-stage.md
```diff
@@ -0,0 +1,192 @@
...
+  seahorse:
+    jobs: [job3]
+    description: "these are seahorse jobs"
+    settings:
+      email: [test@email.com, test2@email.com]
+      slack: 'mychannel'
+```
+
+
+Other jobs can depend on stages, and stages can trigger other jobs within the same pipeline.
```
tkyi, line 164: seems like an elaborate `blockedBy`/join
https://docs.screwdriver.cd/user-guide/configuration/workflow.html#blocked-by
https://docs.screwdriver.cd/user-guide/configuration/workflow.html#join-example

tkyi: we would need to consider circular dependencies

Issue #2719, opened 2022-06-17
Scheduled SD jobs stop without notice if all admins leave a project


adong:
**What happened**:
Scheduled screwdriver jobs fail when all admins leave the github repo. This doesn't prevent checkout or even commits to the repo since devs can have write permission without admin. 

When this happens, no notification is sent to any user that this has stopped running. Though users can check back to the pipeline page to see `no admin`, and manually sync (see fat: https://github.com/screwdriver-cd/ui/pull/771). Still, it is not an ideal situation. 
**What you expected to happen**:
Email specified in the screwdriver.yaml receives an email notification when the pipeline has no active admin, and prompts users that scheduled/periodical jobs will stop working because of no active admins

**How to reproduce it**:
Refer: https://github.com/screwdriver-cd/ui/pull/771 to see how to run SQL query to reproduce

PR #2751, opened 2022-08-12
feat(2719): Scheduled job stop with notification if there's no admin

klu909:
## Context

Queue-service needs a way to notify user when it was not able to start a scheduled job due to no admin in the pipeline
## Objective

Add an endpoint for queue-service to notify user
## References
https://github.com/screwdriver-cd/screwdriver/issues/2719

## License


plugins/jobs/notify.js
```diff
@@ -0,0 +1,61 @@
...
+module.exports = () => ({
+    method: 'POST',
+    path: '/jobs/{id}/notify',
+    options: {
+        description: 'Get a single pipeline',
+        notes: 'Returns a pipeline record',
+        tags: ['api', 'jobs'],
+        auth: {
+            strategies: ['token'],
+            scope: ['pipeline']
```
tkyi, line 19: need to update `pipeline` => `job`

plugins/jobs/notify.js
```diff
@@ -0,0 +1,61 @@
...
+
+            const pipelineId = job.pipelineId;
+
+            if (pipelineId !== credentials.pipelineId) {
+                throw boom.forbidden('Token does not have permission for this pipeline');
+            }
+
+            const pipeline = pipelineFactory.get(pipelineId);
+
+            await request.server.events.emit('job_status', {
```
tkyi, line 40: should this be wrapped in a try/catch with some error logging?

klu909: i am just following the pattern from place where it emits build_status

klu909: and also it's just emitting an event from server, i don't think this will be the point of failure.

tkyi (dismissed):

tkyi (approved):

PR #2762, opened 2022-08-31
feat: Delete method for user settings

banusha2:
## Context

To Reset user settings. 

## Objective

Implement delete method to reset user settings.

## License

### DIFF
--- a/plugins/users/settings/delete.js
+++ b/plugins/users/settings/delete.js
@@ -1,7 +1,6 @@
 'use strict';
 
 const boom = require('@hapi/boom');
-const schema = require('screwdriver-data-schema');
 
 module.exports = () => ({
     method: 'DELETE',
@@ -18,19 +17,13 @@ module.exports = () => ({
         handler: async (request, h) => {
             const { userFactory } = request.server.app;
             const { scmContext, username } = request.auth.credentials;
-            const { settings } = request.payload;
             const user = await userFactory.get({ username, scmContext });
 
             if (!user) {
                 throw boom.notFound('User does not exist');
             }
 
-            return user.updateSettings(settings).then(results => {
-                return h.response(results).code(200);
-            });
-        },
-        validate: {
-            payload: schema.models.user.update
+            return h.response('Uuser settings reset successful').code(204);
         }
     }
 });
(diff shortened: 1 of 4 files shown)


plugins/users/settings/delete.js
```diff
@@ -0,0 +1,36 @@
...
+        tags: ['api', 'users'],
+        auth: {
+            strategies: ['token'],
+            scope: ['user', '!guest']
+        },
+
+        handler: async (request, h) => {
+            const { userFactory } = request.server.app;
+            const { scmContext, username } = request.auth.credentials;
+            const { settings } = request.payload;
```
adong, line 21: This is a `delete`, it expects no payload

adong: Okay, found an example. Please refer, https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/pipelines/caches/delete.js

adong: Actually, here's another example https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/buildClusters/remove.js

plugins/users/settings/delete.js
```diff
@@ -0,0 +1,36 @@
...
+            const user = await userFactory.get({ username, scmContext });
+
+            if (!user) {
+                throw boom.notFound('User does not exist');
+            }
+
+            return user.updateSettings(settings).then(results => {
```
adong, line 28: Just checked `user.removeSettings` does not exist, please add one in https://github.com/screwdriver-cd/models/blob/90de0b119ff542dba128ea690a1530896515d940/lib/user.js#L120-L126, and call it here.

It will be easier to modify `user.updateSettings` later if `user.removeSettings` is used here.

tkyi (changes requested):
Need to add the new route here too: https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/users/index.js

tkyi (commented):
Can you also update https://github.com/screwdriver-cd/screwdriver/tree/master/plugins/users#readme

tkyi (dismissed):

tkyi (commented):
need to fix test
I think you need to add `mock.removeSettings = sinon.stub();`  here:
https://github.com/screwdriver-cd/screwdriver/blob/master/test/plugins/user.test.js#L16

tkyi (approved):

PR #2789, opened 2022-11-17
fix: Simplify badges endpoint

tkyi:
## Context

We are noticing the badges endpoint call often takes a long time.

## Objective

This PR attempts to simplify the endpoint to reduce the amount of time the endpoint takes.

Note: Previously, the badges endpoint fetched all pipeline events, then got builds for the latest event and checked if they exist. If not, it would iterate on the next event.
This enhancement only fetches the most recent event and creates a badge based off that event's builds. This should help reduce number of calls to API and size of responses.

## References
NA

## License

### DIFF
--- a/plugins/pipelines/badge.js
+++ b/plugins/pipelines/badge.js
@@ -2,8 +2,9 @@
 
 const joi = require('joi');
 const schema = require('screwdriver-data-schema');
-const workflowParser = require('screwdriver-workflow-parser');
 const idSchema = schema.models.pipeline.base.extract('id');
+const logger = require('screwdriver-logger');
+const workflowParser = require('screwdriver-workflow-parser');
 const { getPipelineBadge } = require('./helper');
 
 /**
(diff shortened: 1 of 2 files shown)


plugins/pipelines/badge.js
```diff
@@ -54,64 +55,71 @@ module.exports = config => ({
...
+                const lastEvent = latestEvents[0];
+                const builds = await lastEvent.getBuilds({ readOnly: true });
+
+                if (!builds || builds.length < 1) {
+                    return h.response(getPipelineBadge(badgeConfig)).header('Content-Type', contentType);
+                }
+
+                const buildsStatus = builds.reverse().map(build => build.status.toLowerCase());
+                // Get downstream jobs
+                const nextJobs = dfs(lastEvent.workflowGraph, lastEvent.startFrom, lastEvent.prNum);
```
adong, line 100: Is it intentional that `if (lastEvent.workflowGraph) {}` check is removed?

tkyi: yeah I think workflowgraph will always exist
but ended up removing that entire section

pritamstyz4ever (approved):
LGTM

klu909 (approved):

redonkulus:
@tkyi do you know the latency difference before and after your change here?

Issue #2765, opened 2022-09-07
Collections issues


jithine:
- [x]  `Last Run Job` column's data is incorrect. Also build history seems wrong. 
- [x]  Display full name instead of truncating pipeline name
- [x]  Search results displayed when adding multiple pipelines to a collection from Collection page does not display branch and source path info. UX should be similar to search results page. 
- [x]  Support sorting on all columns 
- [x]  Clarify what does sorting on `build history` mean. Let's add a tool tip which says `sort on number of failed builds`
- [x]  Adding pipelines to a collections from different browser tabs is overwriting collection data. 
- [x]  Cannot reset pipeline's `aliasName` (Alan checked, API issue)
- [x]  Once any pipeline's aliasName is set in a collection, that collection shows pipline's aliasName for all pipeline, which isn't ideal.
- [x] handle user type long alias name for pipeline, like `this is a loonnnnng long longggggg alias name`

banusha2:
@jithine `Cannot reset pipeline's aliasName ` is already fixed. Could you please test it again?
API fix PR: https://github.com/screwdriver-cd/data-schema/pull/501
If there is still some issue, let me know.

tkyi:
I don't think sorting on status is working as expected
<img width="222" alt="Screen Shot 2022-11-02 at 11 50 20 AM" src="https://user-images.githubusercontent.com/3230529/199576686-09936d00-5725-4c5d-a350-828d670441dd.png">

tkyi:
Sorting on last run looks like it's sorting by month vs year/date
<img width="143" alt="Screen Shot 2022-11-02 at 11 52 35 AM" src="https://user-images.githubusercontent.com/3230529/199577008-e0522f7d-e278-43c5-bdfa-25e7ed40fbc4.png">

PR #2797, opened 2022-12-06
feat(2765): Add APIs to selectively add/remove collection pipelines

sagar1312:
## Context

API to associate pipelines to a collection expected the complete list of pipeline identifiers whenever a pipeline is either added or removed to the collection. 
This approach may lead to accidental removal of pipelines from a collection when multiple clients (UI) are simultaneously making requests to either add or remove pipelines.

## Objective

Add new APIs to selectively add a pipeline to a collection or remove one or more pipelines from a collection.

#### Add one or more pipelines to a collection

`PUT /collections/{id}/pipelines?ids[]={pipelineId}`

Examples:

`PUT /collections/123/pipelines?ids[]=11&ids[]=23` adds pipelines with `id` equal to `11` and `23` to the collection with `id` equal to `123`

#### Remove one or more pipelines from a collection

`DELETE /collections/{id}/pipelines?ids[]={pipelineId}`

Examples:

`DELETE /collections/123/pipelines?ids[]=11&ids[]=23` removes pipelines with `id` equal to `11` and `23` from the collection with `id` equal to `123`

`DELETE /collections/123/pipelines` removes all the pipelines from the collection with `id` equal to `123`

## References

https://github.com/screwdriver-cd/screwdriver/issues/2765

## License


tkyi:
Can the same format be used for adding pipeline as deleting?

Adding: PUT /collections/{id}/pipelines?ids[]={pipelineId}

sagar1312:
> Can the same format be used for adding pipeline as deleting?
> 
> Adding: PUT /collections/{id}/pipelines?ids[]={pipelineId}

Thought of keeping it similar to DELETE. We currently don't have use case in the UI where we would add more than one pipeline to a collection.

We can definitely extend the API in the future to allow addition of multiple pipelines. Need more thoughts on how to handle certain scenarios. For Ex: the existing API does not error out when some includes an invalid pipeline Ids when saving the complete collection, giving a false impression that all the pipelines were added to the collection. Did not intend to carry similar approach for the new API to support addition of multiple pipelines.

Deferring it for now to focus on other features.

jithine:
> > Can the same format be used for adding pipeline as deleting?
> > Adding: PUT /collections/{id}/pipelines?ids[]={pipelineId}
> 
> Thought of keeping it similar to DELETE. But, we currently don't have use case in the UI where we would add more than one pipeline to a collection.
> 
> We can definitely extend the API in the future to allow addition of multiple pipelines. Need more thoughts on how to handle certain scenarios. For Ex: the existing API does not error out when a user includes invalid pipeline Ids when saving the complete collection, giving a false impression that all the pipelines were added to the collection. Did not intend to carry similar approach for the new API to support addition of multiple pipelines.
> 
> Deferring it for now to focus on other features.

I would say let's keep it consistent with delete. Perhaps can be a search use case.

sagar1312:
> Can the same format be used for adding pipeline as deleting?
> 
> Adding: PUT /collections/{id}/pipelines?ids[]={pipelineId}

Found a use case (copying pipelines from one collection to another) in the UI which needs adding multiple pipelines to a collection. API has been modified to accept multiple pipelines to add.

tkyi (approved):

plugins/collections/addPipelines.js
```diff
@@ -0,0 +1,82 @@
...
+                        throw boom.notFound(`Collection ${id} does not exist`);
+                    }
+                    if (!user) {
+                        throw boom.notFound(`User ${username} does not exist`);
+                    }
+                    if (collection.userId !== user.id) {
+                        throw boom.forbidden(`User ${username} does not own collection`);
+                    }
+
+                    const pipelineIdsQueryParam = request.query[IDS_KEY];
```
adong, line 44: if `IDS_KEY` is `id` instead of `ids[]`, then this code can be 

```diff
---const pipelineIdsQueryParam = request.query[IDS_KEY];
+++const pipelineIdsQueryParam = request.query.id;  

// pipelineIdsQueryParam would be be a single value of 10, if 'id=10' 
// and will convert be pipelineIdsQueryParam to an array of [10, 23] when `id=10&id=23`
```

Fun fact, while I was researching `req.query`. 

I found out that https://github.com/hapijs/hapi/blob/d00293248dd68e9c3e634f5a076c3d8fc19a2049/lib/request.js#L218 

`.query` is from node native query string, https://nodejs.org/api/url.html#class-urlsearchparams.

which uses `URLSearchParams` under the hood, https://nodejs.org/api/url.html#class-urlsearchparams

plugins/collections/removePipelines.js
```diff
@@ -0,0 +1,70 @@
...
+                        throw boom.notFound(`User ${username} does not exist`);
+                    }
+                    if (collection.userId !== user.id) {
+                        throw boom.forbidden(`User ${username} does not own collection`);
+                    }
+
+                    // Only return specific pipelines
+                    const pipelineIdsQueryParam = request.query[IDS_KEY];
+
+                    if (pipelineIdsQueryParam) {
```
adong, line 47: if `IDS_KEY` is `id` instead of `ids[]`, then this code can be 

```diff
---const pipelineIdsQueryParam = request.query[IDS_KEY];
+++const pipelineIdsQueryParam = request.query.id;  

// pipelineIdsQueryParam would be be a single value of 10, if 'id=10' 
// and will convert be pipelineIdsQueryParam to an array of [10, 23] when `id=10&id=23`
```

Fun fact, while I was researching `req.query`. 

I found out that https://github.com/hapijs/hapi/blob/d00293248dd68e9c3e634f5a076c3d8fc19a2049/lib/request.js#L218 

`.query` is from node native query string, https://nodejs.org/api/url.html#class-urlsearchparams.

which uses `URLSearchParams` under the hood, https://nodejs.org/api/url.html#class-urlsearchparams

adong: Just want to mention here, that's why I used [`URLSearchParams`](https://github.com/screwdriver-cd/ui/pull/875/files#diff-684f2ffd6b15b53b3f34d55dd8bdf12113c7ccc172f32c94321f6ee600f64d91R360) in PR  [875](https://github.com/screwdriver-cd/ui/pull/875)

PR #2826, opened 2023-01-30
fix: use node 18.13 image

klu909:
## License


jithine:
we shouldn't pin to a specific version,

PR #2827, opened 2023-01-30
fix: overwrite cpu/ram annotation to high

klu909:
## License

### DIFF
--- a/screwdriver.yaml
+++ b/screwdriver.yaml
@@ -44,8 +44,6 @@ jobs:
             - GIT_KEY
     docker-publish:
         annotations:
-            screwdriver.cd/cpu: HIGH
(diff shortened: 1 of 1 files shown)


screwdriver.yaml
```diff
@@ -43,6 +43,12 @@ jobs:
             # Pushing tags to Git
             - GIT_KEY
     docker-publish:
+        annotations:
+            screwdriver.cd/cpu: HIGH
+            screwdriver.cd/ram: HIGH
+            screwdriver.cd/dockerEnabled: true
+            screwdriver.cd/dockerCpu: HIGH
```
jithine, line 50: only dockerCpu & DockerRam are needed

jithine (approved):

Issue #2853, opened 2023-04-05
Attackers can take admin permissions if admin changed his username


sagar1312:
**What happened**:
Username (/login) in SCM (Ex: Github) is mutable. In screwdriver, admin privileges are configured and determined based on SCM username of the logged in user.

If a user is configured as an admin in Screwdriver and the user changes the username in SCM, the user will lose admin privileges.
If a new account is created in the SCM with the old username, the new user will acquire admin privileges.

**What you expected to happen**:
Determine admin privileges based on SCM user ID which is immutable instead of SCM username.

**How to reproduce it**:
1. Create an account in github (with username `john`)
2. Add the user to the admin list in the Screwdriver API installation (`admins:  [github:john]`)
3. Now change the username in git (`john74`)
   a. The user will lose admin privileges in the screwdriver  
4. Now create a new account in git with old username of the previous account (`john`)
   a. New account user has admin privileges in the screwdriver

PR #2854, opened 2023-04-05
feat(2853): Configure and perform auth checks (admins, whitelist) based on both SCM user ID and name

sagar1312:
## Context
Username (/login) in SCM (Ex: Github) is mutable. In screwdriver, admin privileges are configured and determined based on SCM username of the logged in user.

If a user is configured as admin in Screwdriver and the user changes the username in SCM, the user will lose admin privileges.
If a new account is created in the SCM with the old username, the new user will acquire admin privileges.

## Objective
This PR deliver below changes
- Configure and grant admin privileges based on both SCM user ID and name

- Configure and authenticate allowed users based on both SCM user ID and name

**Old config (Now deprecated)**
```
# list of users able to authenticate against the system
# Ex: ['github:john', 'bitbucket:john']
  whitelist: []

# List of users who should be given screwdriver admin privileges
# Ex: ['github:john', 'bitbucket:john']
  admins: []
```

**New config**
```
# list of users able to authenticate against the system
# Ex: ['github:john:12345', 'bitbucket:john:{98fsa1ba-0b91-4e3c-95ee-55899e933b0}']
  allowList: []

# List of users who should be given screwdriver admin privileges
# Ex: ['github:john:12345', 'bitbucket:john:{98fsa1ba-0b91-4e3c-95ee-55899e933b0}']
  sdAdmins: []
```

**P.S.**
To continue to use old approach, you need set `authCheckById` to `false` in the config.
Support for old approach will soon be removed in the future release.

## References

https://github.com/screwdriver-cd/screwdriver/issues/2853

## License

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -32,9 +32,17 @@ auth:
   https: false
   # A flag to set if you want guests to browse your pipelines
   allowGuestAccess: false
-  # Whitelist of users able to authenticate against the system
+  # Deprecated. Instead, use allowList which is more secure.
+  # List of users able to authenticate against the system
   # if empty, it allows everyone
+  # Values should follow '{scmDisplayName:scmUsername}' format
+  # Ex: ['github:john', 'bitbucket:john']
   whitelist: []
+  # list of users able to authenticate against the system
+  # if empty, it allows everyone
+  # Values should follow '{scmDisplayName:scmUsername:scmUserId}' format
+  # Ex: ['github:john:12345', 'bitbucket:john:{98fsa1ba-0b91-4e3c-95ee-55899e933b0}']
+  allowList: []
   # Deprecated. Instead, use sdAdmins which is more secure.
   # List of users who should be given screwdriver admin privileges
   # Values should follow '{scmDisplayName:scmUsername}' format
@@ -44,9 +52,13 @@ auth:
   # Values should follow '{scmDisplayName:scmUsername:scmUserId}' format
   # Ex: ['github:john:12345', 'bitbucket:john:{98fsa1ba-0b91-4e3c-95ee-55899e933b0}']
   sdAdmins: []
-  # When set to true, performs admin check using the users listed in 'sdAdmins'.
-  # When set to false, performs admin check using the users listed in 'admins'
-  adminCheckById: true
+  # When set to true
+  #  - grant admin privileges to the users listed in 'sdAdmins'
+  #  - only authenticate the users listed in 'allowList'
+  # When set to false, performs
+  #  - grant admin privileges to the users listed in 'admins'
+  #  - only authenticate the users listed in 'whitelist'
+  authCheckById: true
   # Default session timeout (in minutes)
   sessionTimeout: 120
   # SameSite Cookie Option
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -35,7 +35,18 @@ auth:
   # Whitelist of users able to authenticate against the system
   # if empty, it allows everyone
   whitelist: []
+  # Deprecated. Instead, use sdAdmins which is more secure.
+  # List of users who should be given screwdriver admin privileges
+  # Values should follow '{scmDisplayName:scmUsername}' format
+  # Ex: ['github:john', 'bitbucket:john']
   admins: []
+  # List of users who should be given screwdriver admin privileges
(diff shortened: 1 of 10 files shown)


jithine (approved):

config/default.yaml
```diff
@@ -35,7 +35,18 @@ auth:
...
+  # Values should follow '{scmDisplayName:scmUsername}' format
+  # Ex: ['github:john', 'bitbucket:john']
   admins: []
+  # List of users who should be given screwdriver admin privileges
+  # Values should follow '{scmDisplayName:scmUsername:scmUserId}' format
+  # Ex: ['github:john:12345', 'bitbucket:john:{98fsa1ba-0b91-4e3c-95ee-55899e933b0}']
+  sdAdmins: []
+  # When set to true, performs admin check using the users listed in 'sdAdmins'.
+  # When set to false, performs admin check using the users listed in 'admins'
+  adminCheckById: true
```
tkyi, line 49: should probably default false for backwards compatibility atm?

tkyi:
Are we going to make changes to this file too?
https://github.com/screwdriver-cd/queue-service/blob/master/plugins/auth/index.js#L64

plugins/auth/index.js
```diff
@@ -95,12 +97,13 @@ const authPlugin = {
...
          * @method generateProfile
          * @param  {String}        username   Username of the person
+         * @param  {String}        scmUserId  User ID in the SCM
          * @param  {String}        scmContext Scm to which the person logged in belongs
          * @param  {Array}         scope      Scope for this profile (usually build or user)
          * @param  {Object}        metadata   Additonal information to tag along with the login
          * @return {Object}                   The profile to be stored in jwt and/or cookie
          */
-        server.expose('generateProfile', (username, scmContext, scope, metadata) => {
+        server.expose('generateProfile', (username, scmUserId, scmContext, scope, metadata) => {
```
tkyi, line 106: Would prefer something like below so order no longer matters (especially since you're inserting in the middle of the params)
```suggestion
        server.expose('generateProfile', ({ username, scmUserId, scmContext, scope, metadata }) => {
```

DekusDenial (changes requested):
This PR doesn't seem to enforce the same scm user id matching on "whitelist" users, who are also subject to cybersquatting and who are also have write access to pipelines too.

plugins/auth/login.js
```diff
@@ -82,8 +88,15 @@ function addOAuthRoutes(config) {
                 const { collectionFactory } = request.server.app;
                 const accessToken = request.auth.credentials.token;
                 const { username } = request.auth.credentials.profile;
-
-                const profile = request.server.plugins.auth.generateProfile(username, scmContext, ['user'], {});
+                const scmUserId = request.auth.credentials.profile.id;
```
tkyi, line 91: where is `id` set?

sagar1312: It is set by https://github.com/hapijs/bell plugin which is being used as the auth strategy for this route

sagar1312:
> Are we going to make similar changes to this repo too? https://github.com/screwdriver-cd/queue-service/blob/master/plugins/auth/index.js#L64

Thank you for pointing it out. Wasn't aware of this config being used outside the API.
Will make the change

sagar1312:
> This PR doesn't seem to enforce the same scm user id matching on "whitelist" users, who are also subject to cybersquatting and who are also have write access to pipelines too.

@DekusDenial I have pushed the changes to perform auth check on `whitelist` users using SCM username and ID. Please take a look.

plugins/auth/login.js
tkyi: ah I see
https://github.com/hapijs/bell/blob/master/lib/providers/github.js#L26

plugins/auth/login.js
```diff
@@ -82,13 +88,23 @@ function addOAuthRoutes(config) {
...
```
tkyi, line 91: can replace with
```javascript
const { username, id: scmUserId } = request.auth.credentials.profile;
```

sagar1312: Done

DekusDenial (approved):

DekusDenial (approved):

plugins/auth/index.js
sagar1312: https://github.com/screwdriver-cd/screwdriver/pull/2859

Issue #2862, opened 2023-04-21
Template detail view enhancements


sagar1312:
**What happened**:
Template detail view in the UI have below limitations

1. Cannot delete a specific version of the template
2. Unable to get overall usage metrics for a template across all the versions
3. Unable to know how many pipelines are currently using the template
4. Unable to view usage metrics for a particular date/time range

**What you expected to happen**:

- [x] 1. Ability to filter template usage metrics for selected date range (last X months). Set default range to last **1 year**.
- [ ] 2. New metric (`Pipelines`) to show how many distinct pipelines are using the template
- [x] 3. Aggregated usage metrics (`Pipelines`, `Jobs` and `Builds`) across all the versions
- [x] 4. Usage metrics for all the individual versions
- [x] 5. Move trust status button position
- [x] 6. Date when the selected version was published
- [x] 7. Ability to deleted a specific version of the template

**How to reproduce it**:
N/A

itleigns:
Hi, I would like to report a bug and ask a question as YJ member.

First, in my environment, the second newest template does not have the `latest` tag after the template with the `latest` tag is deleted from the UI, which seems not to match the description in the [PR](https://github.com/screwdriver-cd/screwdriver/pull/2863).

Secondly, deleting a template by version can only be performed by the Screwdriver administrator. Is that as expected? If so, shouldn't the ACTIONS tab be invisible to the standard user?

Lastly, It is easier to see if a template is `Trusted` or not by version. However, only the latest version can be changed from the UI. Also, when filtering by `Trusted` in the template list page, only the latest version is checked, and if the latest version is deleted, the result can change. We would like to rethink these specifications.

jithine:
@sagar1312 ^^

sagar1312:
>First, in my environment, the second newest template does not have the latest tag after the template with the latest tag is deleted from the UI, which seems not to match the description in the https://github.com/screwdriver-cd/screwdriver/pull/2863.

When version is deleted, we delete all the tags associated it. Template owner needs make a decision to create/update tags (including `latest`) if needed to point to other versions available.
For Ex: if a version with `stable` tag is deleted, Screwdriver cannot make a decision on which available version can be marked as `stable`. We are extending the same treatment for all the tags (including `latest`).

>Secondly, deleting a template by version can only be performed by the Screwdriver administrator. Is that as expected? If so, shouldn't the ACTIONS tab be invisible to the standard user?

[...]

PR #2863, opened 2023-04-21
feat(2862): Add a new API to delete a specific version of a template

sagar1312:
## Context

The existing API to remove a template, removes all the versions of the specified template and their associated tags.

There could be an use case where template owner publish a new version and later realized that there are some issues with it and prevent any jobs from using it.
To support this, we need an API to just delete a specific version of the template and its associated tags.

## Objective
Add a new API to delete the specified version of the template
`DELETE /templates/{name}/versions/{version}`

Example:
`DELETE /templates/nodejs%2Flib/versions/2.0.12` removes version `2.0.12` of template `lib` under the namespace `nodejs`.

**Note**
If the deleted version was the last published, the API would set the `latest` attribute of the previous version to `true`.

## References

https://github.com/screwdriver-cd/screwdriver/issues/2862

## License

### DIFF
--- a/plugins/templates/removeVersion.js
+++ b/plugins/templates/removeVersion.js
@@ -37,13 +37,18 @@ module.exports = () => ({
                     return canRemove(credentials, template, 'admin', request.server.app)
                         .then(() => {
                             shouldUpdateLatest = template.latest;
-                            const templatePromises = [template.remove()];
-                            const tagPromises = tags.map(tag => tag.remove());
+                            const removeTemplatePromise = template.remove();
+                            const removeTagPromises = tags.map(tag => tag.remove());
 
-                            return Promise.all(templatePromises.concat(tagPromises)).then(() => {
+                            return Promise.all([removeTemplatePromise, ...removeTagPromises]).then(() => {
                                 if (shouldUpdateLatest) {
                                     return templateFactory
-                                        .list({ params: { name }, sort: 'descending', sortBy: 'createTime' })
+                                        .list({
+                                            params: { name },
+                                            sort: 'descending',
+                                            sortBy: 'createTime',
+                                            paginate: { count: 1 }
+                                        })
                                         .then(templates => {
                                             if (templates.length > 0) {
                                                 const newLatestTemplate = templates[0];
--- a/plugins/templates/removeVersion.js
+++ b/plugins/templates/removeVersion.js
@@ -4,7 +4,7 @@ const boom = require('@hapi/boom');
 const joi = require('joi');
 const schema = require('screwdriver-data-schema');
 const nameSchema = schema.models.template.base.extract('name');
-const versionSchema = schema.models.template.base.extract('version');
+const exactVersionSchema = schema.config.template.exactVersion;
 
 module.exports = () => ({
     method: 'DELETE',
@@ -74,7 +74,7 @@ module.exports = () => ({
         validate: {
             params: joi.object({
                 name: nameSchema,
-                version: versionSchema
+                version: exactVersionSchema
             })
         }
     }
--- a/plugins/templates/removeVersion.js
+++ b/plugins/templates/removeVersion.js
@@ -1,0 +1,76 @@
+'use strict';
+
+const boom = require('@hapi/boom');
+const joi = require('joi');
+const schema = require('screwdriver-data-schema');
+const nameSchema = schema.models.template.base.extract('name');
+const versionSchema = schema.models.template.base.extract('version');
+
+module.exports = () => ({
+    method: 'DELETE',
+    path: '/templates/{name}/versions/{version}',
+    options: {
+        description: 'Delete the specified version of template and the tags associated with it',
+        notes: 'Returns null if successful',
+        tags: ['api', 'templates'],
+        auth: {
+            strategies: ['token'],
+            scope: ['build', 'user', '!guest']
+        },
+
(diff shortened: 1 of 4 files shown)


plugins/templates/removeVersion.js
```diff
@@ -0,0 +1,76 @@
...
+                        throw boom.notFound(`Template ${name} with version ${version} does not exist`);
+                    }
+
+                    return canRemove(credentials, template, 'admin', request.server.app)
+                        .then(() => {
+                            shouldUpdateLatest = template.latest;
+                            const templatePromises = [template.remove()];
+                            const tagPromises = tags.map(tag => tag.remove());
+
+                            return Promise.all(templatePromises.concat(tagPromises)).then(() => {
```
adong, line 43: Just easy to read: 
`templatePromises.concat(tagPromises)` => `[...templatePromises, ...tagPromises]`

plugins/templates/removeVersion.js
```diff
@@ -0,0 +1,76 @@
...
+            ])
+                .then(([template, tags]) => {
+                    if (!template) {
+                        throw boom.notFound(`Template ${name} with version ${version} does not exist`);
+                    }
+
+                    return canRemove(credentials, template, 'admin', request.server.app)
+                        .then(() => {
+                            shouldUpdateLatest = template.latest;
+                            const templatePromises = [template.remove()];
```
adong, line 40: `templatePromises` => `removeTemplatePromise`
`tagPromises` => `removeTagsPromises`

plugins/templates/removeVersion.js
```diff
@@ -0,0 +1,76 @@
...
+                            const tagPromises = tags.map(tag => tag.remove());
+
+                            return Promise.all(templatePromises.concat(tagPromises)).then(() => {
+                                if (shouldUpdateLatest) {
+                                    return templateFactory
+                                        .list({ params: { name }, sort: 'descending', sortBy: 'createTime' })
```
adong, line 46: I would say, just fetch the first one in descending order? 

https://github.com/screwdriver-cd/models/blob/master/lib/baseFactory.js#L184

from: 
```
.list({ params: { name }, sort: 'descending', sortBy: 'createTime' })
```

to
```
.list({ params: { name }, sort: 'descending', sortBy: 'createTime', paginate: { count: 1 }})
```

plugins/templates/removeVersion.js
tkyi: removeTagPromises* to be consistent with the other one

plugins/templates/removeVersion.js
tkyi, line 1: do we need to implement soft delete for templates?

tkyi (changes requested):
Can you update README: https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/templates/README.md

Also can you add new functionality to the template npm package: https://github.com/screwdriver-cd/template-main/blob/master/index.js

plugins/templates/removeVersion.js
sagar1312: This needs to be addressed across the board. I believe, we only support soft deletion of jobs. We need extend the same to other entities like `pipeline`, `template` which have come up in earlier discussions.
Supporting soft delete for templates would be outside the scope https://github.com/screwdriver-cd/screwdriver/issues/2862

plugins/templates/removeVersion.js
sagar1312: Done

plugins/templates/removeVersion.js
sagar1312: Done

plugins/templates/removeVersion.js
sagar1312: Done

sagar1312:
> Also can you add new functionality to the template npm package: https://github.com/screwdriver-cd/template-main/blob/master/index.js

https://github.com/screwdriver-cd/template-main/pull/37

sagar1312:
@adong @tkyi 
Addressed all the suggestions. Please take a look.

plugins/templates/removeVersion.js
```diff
@@ -0,0 +1,81 @@
+'use strict';
+
+const boom = require('@hapi/boom');
+const joi = require('joi');
+const schema = require('screwdriver-data-schema');
+const nameSchema = schema.models.template.base.extract('name');
+const versionSchema = schema.models.template.base.extract('version');
```
tkyi, line 7: just thought of something
would it be better to use `exactVersion` schema? https://github.com/screwdriver-cd/data-schema/blob/master/config/template.js#L80

what happens if a user puts `12.0` vs `12.0.0` if we use `version` schema?

sagar1312: We look for exact match in DB while fetching the template. 

When version `12.0` is specified, 404 is returned with error message saying template with version `12.0` does not exist. Version `12.0.0` is not removed when `12.0` is specified in the request.
**Sample Error**
```javascript
{
    "statusCode": 404,
    "error": "Not Found",
    "message": "Template sagar1312/sd-template-parameters with version 12.0 does not exist"
}

```


With `exactVersion`, it is fail-fast with a generic error
```
{
    "statusCode": 400,
    "error": "Bad Request",
    "message": "Invalid request params input"
}
```

Made changes to use `exactVersion` while validating the input.

tkyi (approved):

Issue #2767, opened 2022-09-09
Setup & Teardown actions for Pipeline stages


jithine:
**What happened**:

[Pipeline stages ](https://github.com/screwdriver-cd/screwdriver/issues/2669)introduces logical grouping of various jobs, for example group all your `production` jobs as one stage of your build pipeline. It would be beneficial to provide setup and teardown actions for a Pipeline stage as a whole so that any initialization or cleanups needed before jobs in a stage runs can be executed. Idea is similar to the setup & teardown steps for an individual job

**What you expected to happen**:

Users should be able to define pre & post actions for a stage. 
UI should provide interface where users can see the logs for pre & post actions.
Pre actions should run exactly once before any job in the stage runs.
Post actions should run exactly once when a stage ends due to either success or failure of individual jobs.
Pre & Post actions could be fashioned as jobs

Sub-Tasks:

- [x] https://github.com/screwdriver-cd/screwdriver/issues/3022
- [x] https://github.com/screwdriver-cd/screwdriver/issues/3026
- [x] https://github.com/screwdriver-cd/screwdriver/issues/3045

**References**:

Mocks - TBD

jithine:
cc @adong @tkyi 

Alan pls share mocks with the actions mocked up

tkyi:
Some things for discussion:

# Pre/postaction in stage
- is this just another SD job, but it will always run (like how a setup/teardown step is just another SD step)?
- can you rerun a pre/postaction job manually?
- can you enable/disable a pre/postaction job?
- job name ideas: `pre/post{stage}`, `pre/postaction`, `setup/teardown{stage}`, `sd-setup-{stage}/sd-teardown-{stage}`
- what image will the job use? generic image, or will we let the user configure?
- will we skip `sd-setup-scm`(git clone) step?

## SD Config
Current stage config:
```
jobs:
  job1:
    requires: [~pr, ~commit]
    image: node:14
    steps:
      - init: echo 'init'
  job2:
    requires: [~pr, ~commit]
    image: node:14
    steps:
      - init: echo 'init'
  job3:
    requires: [~pr, ~commit]
    image: node:14
    steps:
      - init: echo 'init'
  job4:
    requires: [~pr, ~commit]
    image: node:14
    steps:
      - init: echo 'init'

stages:
  seahorse:
    jobs: [job3, job4]
    description: "seahorse starfish satarfish deploy jobs"
  dolphin:
    jobs: [job2, job1]
[...]

PR #2870, opened 2023-05-04
feat(2767): Handle stage logic in build workflow [5]

tkyi:
## Context

Properties might want to fetch stageBuilds for a stage.

## Objective

This PR adds an endpoint to get stageBuilds.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/2767

## License

I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.


plugins/builds/index.js
```diff
@@ -785,18 +820,86 @@ const buildsPlugin = {
...
+                    const workflowStageBuilds = [];
+
+                    current.flattenedWorkflowGraph.nodes.forEach(n => {
+                        if (n.stageName === currentNode.stageName) {
+                            workflowStageBuilds.push(n.name);
+                        }
+                    });
+
+                    // get all builds in current stage
+                    const buildParentBuilds = current.build.parentBuilds;
```
sagar1312, line 871: Relying on `parentBuilds` may not work for all scenarios.
For example: Below stage has 6 jobs on two separate paths running in parallel
image
Parent builds for `prod-blue-certify` will only include `prod-blue-deploy` and `prod-blue-test`.
So the `teardown` will not be triggered when the all the jobs in the stage are successfully completed.

plugins/builds/index.js
```diff
@@ -785,18 +820,86 @@ const buildsPlugin = {
...
+                } else {
+                    //  else get all stage builds in workflow, compare number of stageBuilds to stage builds with status
+                    // Get job names from stage workflow
+                    const workflowStageBuilds = [];
+
+                    current.flattenedWorkflowGraph.nodes.forEach(n => {
+                        if (n.stageName === currentNode.stageName) {
+                            workflowStageBuilds.push(n.name);
+                        }
+                    });
```
sagar1312, line 868:
```suggestion
                    const workflowStageBuilds = currentStageBuild.workflowGraph.nodes.map(n => return n.name);
```

plugins/builds/index.js
sagar1312: Another scenario could be a stage with only one job.

plugins/builds/index.js
```diff
@@ -785,18 +820,86 @@ const buildsPlugin = {
...
+                                const buildId = buildParentBuilds[pipeline.id].jobs[b];
+                                const buildForStage = eventBuilds.filter(eb => eb.id === buildId);
+
+                                stageDict[b] = buildForStage.length > 0 ? buildForStage[0] : undefined;
+                            }
+                        });
+                    }
+
+                    // Make sure number of builds is same as expected number of stage jobs
+                    if (Object.values(stageDict).every(b => b !== undefined)) {
```
sagar1312, line 887: `teardown` may not get triggered in some scenarios.
For Ex: Below stage has 3 jobs
image
When `prod-blue-test` fails, `teardown` should get triggered. But this logic expect build for all the jobs within a stage. Since `prod-blue-certify` does not have a build, `teardown` will not be triggered.

plugins/builds/index.js
```diff
@@ -785,18 +820,86 @@ const buildsPlugin = {
...
+                        // Check build statuses
+                        const stageIsDone = builds.every(b => isNotActiveBuild(b.status, b.endTime));
+                        const shouldRunStageTeardown = builds.some(b => hasTerminalBuild(b.status, b.endTime));
+
+                        // if any have terminal status or all no longer active, destroy created stage builds, set next job to teardown
+                        if (shouldRunStageTeardown || stageIsDone) {
+                            builds.forEach(async b => deleteBuild(b, buildFactory));
+                            nextJobsTrigger = [`stage@${currentNode.stageName}:teardown`];
+                        }
+                        // TODO: set final stage status
```
sagar1312, line 899: Note to myself. To review this once implemented.

sagar1312 (changes requested):

plugins/stages/stageBuilds/list.js
```diff
@@ -0,0 +1,68 @@
...
+const boom = require('@hapi/boom');
+const joi = require('joi');
+const schema = require('screwdriver-data-schema');
+const listSchema = joi.array().items(schema.models.stageBuild.get).label('List of stage builds for a stage');
+const idSchema = schema.models.stage.base.extract('id');
+const eventIdSchema = schema.models.stageBuild.base.extract('eventId');
+
+module.exports = () => ({
+    method: 'GET',
+    path: '/stages/{id}/stageBuilds',
```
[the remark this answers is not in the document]
tkyi: added

plugins/builds/index.js
```diff
@@ -765,7 +920,40 @@ const buildsPlugin = {
...
+                                    pipelineId: current.pipeline.id
+                                });
+
+                                if (nextStage) {
+                                    const nextStageBuild = await stageBuildFactory.get({
+                                        stageId: nextStage.id,
+                                        eventId: current.event.id
+                                    });
+
+                                    await nextStageBuild.remove();
```
sagar1312, line 953: This should be abstracted into `build` model
https://github.com/screwdriver-cd/models/blob/master/lib/build.js#L652
```suggestion
                            const nextJobIsStageSetup = STAGE_SETUP_PATTERN.test(nextJobName);

                            // Delete next stageBuild if next job is stage setup
                            if (nextJobIsStageSetup) {
                                const [, nextStageName] = nextJobName.match(STAGE_SETUP_PATTERN);
                                const nextStage = await stageFactory.get({
                                    name: nextStageName,
                                    pipelineId: current.pipeline.id
                                });

                                if (nextStage) {
                                    const nextStageBuild = await stageBuildFactory.get({
                                        stageId: nextStage.id,
                                        eventId: current.event.id
                                    });

                                    await nextStageBuild.remove();
```

plugins/builds/update.js
```diff
@@ -245,18 +268,58 @@ module.exports = () => ({
...
+                stageBuildHasFailure = ['FAILURE', 'ABORTED', 'UNSTABLE'].includes(stageBuild.status);
+            }
 
+            // Guard against triggering non-successful or unstable builds
+            // Don't further trigger pipeline if intend to skip further jobs
             if (newBuild.status !== 'SUCCESS' || skipFurther) {
                 // Check for failed jobs and remove any child jobs in created state
                 if (newBuild.status === 'FAILURE') {
-                    await removeJoinBuilds({ pipeline, job, build: newBuild }, request.server.app);
+                    await removeJoinBuilds(
```
sagar1312, line 303: `removeJoinBuild()` needed to be invoked when the current build is `teardown` and has successfully completed, but the `stageBuild` status is a negative terminal state

plugins/builds/index.js
```diff
@@ -870,19 +1058,77 @@ const buildsPlugin = {
...
+                            // If next build is stage setup, create stageBuild and set to running
+                            if (nextBuildStage && nextJobName === stageSetupName) {
+                                await stageBuildFactory.create({
+                                    stageId: nextBuildStage.id,
+                                    eventId: current.event.id,
+                                    status: 'CREATED'
+                                });
+                            }
+                        }
+
```
sagar1312, line 1086: We should abstract this into `buildFactory`. This would ensure, `stageBuild` will always gets created when `setup` build is created irrespective of from where the API is called.

plugins/builds/index.js
```diff
@@ -870,19 +1058,77 @@ const buildsPlugin = {
...
+                                await existNextBuild.remove();
+
+                                return null;
+                            }
+
+                            return existNextBuild;
+                        }
+                    }
+
+                    // Current build is not part of stage
```
sagar1312, line 1122: With the latest changes in `index.js` update build endpoint, `triggerNextJobs()` will not be invoked when `teardown` has successfully completed, but `stageBuild` has failure. Hence, this block of code will not clean up the downstream builds of `teardown`.

Instead, invoke `removeJoinBuilds()` in `index.js` update build endpoint to cleanup downstream builds of `teardown` when `stageBuild` has failure.
```suggestion
                    if (!['CREATED', null, undefined].includes(existNextBuild.status)) {
                        return existNextBuild;
                    }
                    
```

plugins/builds/index.js
```diff
@@ -515,6 +518,83 @@ async function getParentBuildStatus({ newBuild, joinListNames, pipelineId, build
...
+            start: false
+        });
+
+        return stageTeardownBuild;
+    }
+
+    // Create stageBuild if newBuild is stage setup
+    if (jobName === stageSetupName) {
+        await stageBuildFactory.create({ stageId: stage.id, eventId: event.id, status: 'CREATED' });
+    }
```
sagar1312, line 593: This is a common step for both `AND` and `OR`. As suggested, if we abstract the logic to create `stageBuild` when `setup` build is created into `buildFactory`, we can remove this duplicate code.

plugins/builds/index.js
```diff
@@ -515,6 +518,83 @@ async function getParentBuildStatus({ newBuild, joinListNames, pipelineId, build
...
+            jobName: stageTeardownName,
+            username,
+            scmContext,
+            event, // this is the parentBuild for the next build
+            baseBranch: event.baseBranch || null,
+            start: false
+        });
+
+        return stageTeardownBuild;
+    }
```
sagar1312, line 588: If a stage job fails, `teardown` build would be created via `removeJoinBuilds` logic.
Why do we need to create `teardown` build here?

sagar1312 (changes requested):

plugins/builds/index.js
```diff
@@ -515,6 +518,83 @@ async function getParentBuildStatus({ newBuild, joinListNames, pipelineId, build
...
+                params: { eventId: event.id, jobId: stageTeardownJob[0].id }
+            });
+
+            if (existingStageTeardownBuild && existingStageTeardownBuild[0]) {
+                return existingStageTeardownBuild[0];
+            }
+        }
+
+        // Doesn't exist, create stage teardown job and return as next job
+        const stageTeardownBuild = await createInternalBuild({
```
sagar1312, line 575: How would `teardown` build created here would get started/executed?

plugins/builds/index.js
tkyi: Opened a PR: https://github.com/screwdriver-cd/models/pull/600

plugins/builds/index.js
tkyi: Opened a PR: https://github.com/screwdriver-cd/models/pull/600

plugins/builds/index.js
```diff
@@ -709,6 +764,67 @@ function getParentBuildIds({ currentBuildId, parentBuilds, joinListNames, pipeli
...
+
+    if (nextJobName === stageTeardownName) {
+        return buildDeletePromises;
+    }
+    if (buildConfig.eventId) {
+        // Remove next build
+        buildDeletePromises.push(deleteBuild(buildConfig, buildFactory));
+
+        // Check if stage teardown build already exists
+        const stageTeardownJob = await jobFactory.get({
```
sagar1312, line 798: We should separate the concerns. Let us not interleave the responsibility of creating `teardown` build along side cleaning up of join builds.

Can we move the logic to create `teardown` build to separate function `checkAndCreateStageTeardownBuild()` and invoke this when a stage job fails?

plugins/builds/index.js
```diff
@@ -873,16 +1020,44 @@ const buildsPlugin = {
                         return createInternalBuild(internalBuildConfig);
                     }
 
-                    if (!['CREATED', null, undefined].includes(existNextBuild.status)) {
+                    if (!['CREATED', null, undefined].includes(existNextBuild.status) && !current.stage) {
```
sagar1312, line 1023: As suggested here
https://github.com/screwdriver-cd/screwdriver/pull/2870/files/9f679937feaa4e278d80b3b1ef245984c9b03993#r1448111687, we need to revert this change.

```suggestion
                    if (!['CREATED', null, undefined].includes(existNextBuild.status)) {
```

plugins/builds/index.js
```diff
@@ -873,16 +1020,44 @@ const buildsPlugin = {
...
+                                await existNextBuild.remove();
+
+                                return null;
+                            }
+
+                            return existNextBuild;
+                        }
+                    }
+
+                    // Current build is not part of stage
```
sagar1312, line 1053: As suggested here https://github.com/screwdriver-cd/screwdriver/pull/2870/files/9f679937feaa4e278d80b3b1ef245984c9b03993#r1448111687, this block of code is no longer needed.

```suggestion
```

plugins/builds/update.js
```diff
@@ -245,16 +270,99 @@ module.exports = () => ({
...
+            // (if stage teardown build exists, and stageBuild.status is negative,
+            // and there are no active stage builds, and teardown build is not started)
+            if (stage) {
+                const stageTeardownName = getFullStageJobName({ stageName: stage.name, jobName: 'teardown' });
+                const stageTeardownJob = await jobFactory.get({
+                    params: { pipelineId: pipeline.id, name: stageTeardownName }
+                });
+                let stageTeardownBuild;
+
+                if (stageTeardownJob) {
```
sagar1312, line 333: This check is not needed. A stage must always have `teardown` job
```suggestion
               
```

plugins/builds/update.js
```diff
@@ -245,16 +270,99 @@ module.exports = () => ({
...
+                await triggerNextJobs(
+                    { pipeline, job, build: newBuild, username, scmContext, event: newEvent, stage },
+                    request.server.app
+                );
+            }
+
+            // Determine if stage teardown build should start
+            // (if stage teardown build exists, and stageBuild.status is negative,
+            // and there are no active stage builds, and teardown build is not started)
+            if (stage) {
```
sagar1312, line 326: We can skip this block, if the `newBuild` status is not a terminal status
```suggestion
            if (stage && FINISHED_STATUSES.includes(newBuild.status)) {
```

plugins/builds/update.js
```diff
@@ -245,16 +270,99 @@ module.exports = () => ({
...
+                    if (
+                        stageTeardownBuild &&
+                        stageTeardownBuild.status === 'CREATED' &&
+                        stageBuildHasFailure &&
+                        stageIsDone
+                    ) {
+                        stageTeardownBuild.status = 'QUEUED';
+                        await stageTeardownBuild.update();
+                        await stageTeardownBuild.start();
+                    }
```
sagar1312, line 364: If the `teardown` build is already started, we don't need to query for other stage job builds and determine if the stage is done.

Also move the logic to determine whether the stage is done to. a separate function.
```suggestion
if(stageTeardownBuild.status === 'CREATED' && isStageDone(stage)) {
    stageTeardownBuild.status = 'QUEUED';
    await stageTeardownBuild.update();
    await stageTeardownBuild.start();
}
```


```
isStageDone(stage){
// Get all jobIds for jobs in the stage
                    const stageJobIds = stage.jobIds;

                    stageJobIds.push(stage.setup);

                    // Get all builds in a stage
                    const stageJobBuilds = await buildFactory.list({
                        params: { jobId: stageJobIds, eventId: newEvent.id }
                    });
                    let stageIsDone = false;

                    if (stageJobBuilds && stageJobBuilds.length !== 0) {
                        stageIsDone = !stageJobBuilds.some(b => !FINISHED_STATUSES.includes(b.status));
                    }

return stageIsDone;
}
```

sagar1312 (changes requested):

plugins/builds/index.js
tkyi: https://github.com/screwdriver-cd/models/pull/606

plugins/builds/index.js
```diff
@@ -709,6 +721,82 @@ function getParentBuildIds({ currentBuildId, parentBuilds, joinListNames, pipeli
...
+/**
+ * Create stage teardown build if it doesn't already exist
+ * @param  {Factory}    jobFactory                      Job factory
+ * @param  {Factory}    buildFactory                    Build factory
+ * @param  {Object}     current                         Current object
+ * @param  {String}     stageTeardownName               Stage teardown name
+ * @param  {String}     username                        Username
+ * @param  {String}     scmContext                      SCM context
+ */
+async function checkStageTeardownBuild({ jobFactory, buildFactory, current, stageTeardownName, username, scmContext }) {
```
sagar1312, line 733:
```suggestion
async function checkAndCreateStageTeardownBuild({ jobFactory, buildFactory, current, stageTeardownName, username, scmContext }) {
```

sagar1312 (approved):

sagar1312 (approved):

sagar1312 (approved):

PR #2908, opened 2023-08-04
fix: Fix job triggers not working properly depending on conditions

yakanechi:
## Context
Currently, it has been observed that the trigger is not functioning correctly under certain conditions. The conditions are as follows: The numbers indicate the order of execution. S = Succeeded, F = Failed

1. requires [S1, ~S2, F3]
It seems that the next build starts due to the success of S2 without waiting for the result of F3. However, since the build has already been created at S1, S2 fails to create the next build and cannot start the job.
スクリーンショット 2023-08-04 16 23 17

2. requires [~S1, F2, S3] 
After the success of S1, the build is created and started. However, the failure of F2 and S3 trigger results in the removal of the build created by S1. Therefore, causing the next build will be deleted.
スクリーンショット 2023-08-04 16 23 06

3. requires [~S1, ~S2]
In this case, the operation works, but an error occurs when trying to create a build with a duplicate key in the database upon the success of S2.

## Objective
Fixes the above problems.
The specific fixes are as follows.
1. Prevent multiple calls to the create build function by checking in advance whether the next build exists before the trigger's build creation process.
2. If the trigger meets the execution conditions when the next build already exists, the existing build is executed as is.
3. When deleting a build, add a condition that it is before execution.
4. Fix the test according to the above modifications.

## License

### DIFF
--- a/plugins/builds/index.js
+++ b/plugins/builds/index.js
@@ -851,6 +851,8 @@ const buildsPlugin = {
 
                             return (await existNextBuild.update()).start();
                         }
+
+                        return existNextBuild;
                     }
 
                     return createInternalBuild(internalBuildConfig);
--- a/plugins/builds/index.js
+++ b/plugins/builds/index.js
@@ -528,27 +528,31 @@ async function getParentBuildStatus({ newBuild, joinListNames, pipelineId, build
  * @return {Promise}                The newly updated/created build
  */
 async function handleNewBuild({ done, hasFailure, newBuild, jobName, pipelineId }) {
-    if (done) {
-        // Delete new build since previous build failed
-        if (hasFailure) {
-            logger.info(
-                `Failure occurred in upstream job, removing new build - build:${newBuild.id} pipeline:${pipelineId}-${jobName} event:${newBuild.eventId} `
-            );
-            await newBuild.remove();
-
-            return null;
-        }
-
-        // If all join builds finished successfully and it's clear that a new build has not been started before, start new build
-        if (['CREATED', null, undefined].includes(newBuild.status)) {
-            newBuild.status = 'QUEUED';
-            const queuedBuild = await newBuild.update();
-
-            return queuedBuild.start();
-        }
+    if (!done) {
+        return null;
+    }
+    if (!['CREATED', null, undefined].includes(newBuild.status)) {
+        // Possible build status group
+        // ['RUNNING', 'QUEUED', 'BLOCKED', 'UNSTABLE', 'FROZEN', 'COLLAPSED', 'SUCCESS', 'FAILURE', 'ABORTED']
+        return null;
     }
 
-    return null;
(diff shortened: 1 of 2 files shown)


plugins/builds/index.js
```diff
@@ -528,27 +528,29 @@ async function getParentBuildStatus({ newBuild, joinListNames, pipelineId, build
...
+        return null;
     }
 
-    return null;
+    // All join builds finished successfully and it's clear that a new build has not been started before.
+    // Start new build.
+    newBuild.status = 'QUEUED';
+    await newBuild.update();
+
+    return newBuild.start();
```
itleigns, line 553: Elsewhere, the return value of `update` is used. Wouldn't it be better to follow that?

jithine:
Also another issue related to job trigger, not sure if these are related https://github.com/screwdriver-cd/screwdriver/issues/2909

plugins/builds/index.js
yakanechi: The update function just returns the instance itself, so there seems no need to reassign the return value and use it.

In addition, I considered that it would be more dangerous to cause the misunderstanding that it was returning a copy of the instance by returning.

itleigns: OK.

itleigns (approved):

yakanechi:
@jithine 
> Also another issue related to job trigger, not sure if these are related #2909

I did the test and it wasn't blocked.
Unfortunately, the fixes in this PR do not seems to be related #2909.

jithine (approved):
would need another merge/rebase with master branch

jithine:
> @jithine
> 
> > Also another issue related to job trigger, not sure if these are related #2909
> 
> I did the test and it wasn't blocked. Unfortunately, the fixes in this PR do not seems to be related #2909.

Thank you for testing @yakanechi Another join job related error our internal Yahoo users are reported are around Remote-Join scenarios, which is not covered in this code path.

yakanechi:
@jithine 
Thank you for your comments.
Probably the problem is in a completely different part of this PR.
We think it would be better to make a new PR for the problem.

jithine:
@yakanechi @itleigns We are noticing something weird, multiple builds are now being created in a join scenario with OR condition and they extra builds end up changing to Collapsed builds.

This event https://cd.screwdriver.cd/pipelines/10293/events/765417 has a COLLAPSED status. Because the join job `join` in the OR condition is getting created multiple times in same event.

jithine:
So we need to do a release to fix a DB migration issue, so we are reverting this for now, we can introduce this back after the release

jithine:
So we did more digging and it seems that this PR never made it to prod and duplicate builds are still happening. So I am reverting the revert. https://github.com/screwdriver-cd/screwdriver/pull/2915

But something is causing duplicate builds for the same job to be create in the same event

PR #2917, opened 2023-08-15
feat: add exhaustive health check

pritamstyz4ever:
## Context

Health check does not include checking for queue and store

## Objective

This PR adds a check for queue and store health if request has exhaustive in query params

## License

### DIFF
--- a/plugins/status.js
+++ b/plugins/status.js
@@ -1,6 +1,40 @@
 'use strict';
 
+const logger = require('screwdriver-logger');
 const schema = require('screwdriver-data-schema');
+const requestRetry = require('screwdriver-request');
+
+const RETRY_LIMIT = 2;
+const RETRY_DELAY = 5; // in seconds
+
+/**
+ * Makes api call to the url endpoint
+ * @async invoke
+ * @param {String} url
+ * @return Promise.resolve
+ */
+async function invoke(url) {
+    logger.info(`GET ${url}`);
+
+    const options = {
+        url,
+        retry: {
+            limit: RETRY_LIMIT,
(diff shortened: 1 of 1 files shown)


plugins/status.js
```diff
@@ -14,7 +50,21 @@ const statusPlugin = {
...
             method: 'GET',
             path: '/status',
-            handler: (_, h) => h.response('OK').code(200),
+            handler: async (request, h) => {
+                const { exhaustive } = request.query;
+
+                if (exhaustive) {
+                    const queueResponseCode = await invoke(`${request.server.app.ecosystem.queue}/v1/status`);
+                    if (queueResponseCode !== 200) {
+                        return h.response('SD Queue Service Unavailable').code(queueResponseCode);
```
jithine, line 59: why not fetch store response as well and return both in case of error

plugins/status.js
```diff
@@ -14,7 +51,29 @@ const statusPlugin = {
         server.route({
             method: 'GET',
             path: '/status',
-            handler: (_, h) => h.response('OK').code(200),
+            handler: async (request, h) => {
+                const { exhaustive } = request.query;
```
tkyi, line 55: Wondering if there's a different term we can use vs `exhaustive`. Seems easy to misspell.
No strong preference though

jithine: extended ?

tkyi: I guess either `complete` for ease of use or just pick the word that best describes the intent (which could be `exhaustive` or `extensive` or whatever)

jithine: let's stick with `exhaustive` for now

jithine (approved):

Issue #2135, opened 2020-06-30
Pipeline templates


hk-vmg:
**What happened**:
Many teams cut-and-paste screwdriver.yaml and also have difficulty building correct, standards-compliant, maintainable pipelines.

**What you expected to happen**:
I would love to see a reusable "pipeline" concept implemented in some way.  Ideally screwdriver.yaml embodies an instance of a "pipeline" which can be configured, templatized (a "type" of pipeline), and distributed. A missing screwdriver.yaml (see #2104 ) is just a pipeline of type "generic" with generic defaults.

A pipeline could include a default set of jobs, which could be redefined, and a set of required or optional configuration variables.

A brute-force way to understand the idea is to "include" a shared canonical screwdriver.yaml, and then de-dupe all yaml entries, though there are cleaner/more robust ways to support it.

jithine:
We are seeing more our Yahoo users requesting this feature. Copying from an internal ask.

At the first look parent pipelines might provide this functionality but it does not, parent pipelines are a centralised way to execute the same jobs across multiple repositories but it doesn’t allow versioning, changes in the parent pipeline affects all the child pipelines that is why this proposal is to allow a pipeline that is a template that can be versioned and instantiate on each repository and the screwdriver.yaml might have a simple reference to the pipeline definition like this example your screwdriver.yaml only has a reference to the pipeline
 
This might be the pipeline that deploys Java Kubernetes workloads 

environment:
  DO_SOMETHING_DIFFERENT: yes
pipelines:
  name: pipeline/java-k8s@1.0.0

And this might be the pipeline that deploys nodejs workloads on ECS

pipelines
  name: pipeline/nodejs-ecs@1.1.1

If changes are made in the pipeline the version is bumped to the new desired version.

quike:
This is one of the most exciting enhancements for re usability and standardization. If following the same concept as current jobs templates or commands we use `stable` and `latest` tags, even better:
```
template: pipeline/nodejs-ecs@1.1.1
# or
template: pipeline/nodejs-ecs@stable
# or 
template: pipeline/nodejs-ecs@latest
```
Thanks for this feature!

VonnyJap:
This feature has been released on MVP version. And now the enhancement of pipeline template is tracked under this issue.
https://github.com/screwdriver-cd/screwdriver/issues/3059

PR #2923, opened 2023-08-24
fix(2135): Set templateType filter while getting template tag from DB for all the workflows

sagar1312:
## Context
Changes were made in PR https://github.com/screwdriver-cd/screwdriver/pull/2922 to set `templateType` filter while getting template tag from DB.
But this got applied only the workflow to create/update a template tag.

But we noticed, there are more workflows which needs similar fix.
Ex: Synchronizing the pipeline to reflect the latest `screwdriver.yaml` configuration during event creation.
More info is here https://github.com/screwdriver-cd/models/pull/586

## Objective
Undo PR https://github.com/screwdriver-cd/screwdriver/pull/2922
Upgrade to the latest `screwdriver-models`(v29.7.0) to apply the fix made in https://github.com/screwdriver-cd/models/pull/586

## References
https://github.com/screwdriver-cd/screwdriver/issues/2135

## License


screwdriver.yaml
```diff
@@ -48,7 +48,7 @@ jobs:
             screwdriver.cd/dockerCpu: HIGH
             screwdriver.cd/dockerRam: HIGH
         requires: publish
-        template: sd/dind@latest
```
sagar1312, line 51: Temporarily specifying the version instead of the tag as OSSD pipeline to build and deploy screwdriver is broken.

adong (approved):

PR #2933, opened 2023-09-21
feat(2767): Remove 'workflowGraph' from stageBuilds

sagar1312:
## Context
Pipeline workflow graph and stage workflow graph were getting merged to generate unified workflow graph which was used by the API while processing an event.

Workflow parser has been modified to eliminate stage workflows and always generate unified workflow graph which include node/edges for all the jobs irrespective whether the job belongs to a stage or not.
https://github.com/screwdriver-cd/workflow-parser/pull/4

## Objective
Update API to use unified workflow graph.

## References

https://github.com/screwdriver-cd/screwdriver/issues/2767

## License

### DIFF
--- a/plugins/builds/index.js
+++ b/plugins/builds/index.js
@@ -608,7 +608,7 @@ function fillParentBuilds(parentBuilds, current, builds, nextEvent) {
 
                 // parentBuild is in current event
                 if (+pid === current.pipeline.id) {
-                    workflowGraph = current.flattenedWorkflowGraph;
+                    workflowGraph = current.event.workflowGraph;
                 } else if (nextEvent) {
                     if (+pid !== nextEvent.pipelineId) {
                         // parentBuild is remote triggered from external event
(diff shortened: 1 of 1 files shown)


jithine (approved):

plugins/builds/index.js
```diff
@@ -675,10 +675,10 @@ async function createJoinObject(nextJobs, current, eventFactory) {
 
             if (externalEvent) {
                 pipelineObj.event = externalEvent;
-                jobs = workflowParser.getSrcForJoin(externalEvent.workflowGraph, { jobName: jName });
+                jobs = workflowParser.getSrcForJoin(event.workflowGraph, { jobName: jName });
```
tkyi, line 678: why aren't we using `externalEvent` here?

sagar1312: Good catch. Fixed it.

tkyi (approved):

Issue #2725, opened 2022-06-25
SD_PARENT_BUILD_ID is not always accurate


DekusDenial:
**What happened**:
The environment variable `SD_PARENT_BUILD_ID` created at the beginning of a build is sometimes missing parent build ID if the build itself has multiple joined parents

**What you expected to happen**:
`SD_PARENT_BUILD_ID` must be accurate and include all parent build IDs

**How to reproduce it**:
Not all the times, but if you have two parent builds completed closely following one another, the next build may be missing the build ID for the parent which completed last.

jithine:
Should we fixed with the linked pR

PR #2937, opened 2023-10-16
fix: Remote Join double build

yk634:
## Context

1. In the following Remote Join workflow, when the build of `a2` and the build of `a3` are finished almost simultaneously, the build of `a4` may be executed twice.

image

2. If there are jobs waiting for the same job with Remote Join and Remote Or triggers, both of them (`b4` and `b5`) will be executed twice, as follows.

image

## Objective

1. In some cases, when jobs waiting for Remote Join finish almost simultaneously, the parentBuildId is not set correctly for the triggered build and [it becomes the target of Restart](https://github.com/screwdriver-cd/screwdriver/blob/cc915df162b3aa8117ad8f2294ced64c7f75513d/plugins/builds/index.js#L943-L953), resulting in double execution.
So, when a build is created by a remote trigger, modify parentBuildId to be set to all builds of the job that condition the Join.

2. Add to the Restart condition that the eventId of the already created build and the triggering event do not match.

## References

[#2725](https://github.com/screwdriver-cd/screwdriver/issues/2725)

## License

### DIFF
--- a/plugins/builds/index.js
+++ b/plugins/builds/index.js
@@ -454,7 +454,7 @@ async function updateParentBuilds({ joinParentBuilds, nextBuild, build }) {
     );
 
     nextBuild.parentBuilds = newParentBuilds;
-    nextBuild.parentBuildId = [build.id].concat(nextBuild.parentBuildId || []);
+    nextBuild.parentBuildId = Array.from(new Set([build.id, ...(nextBuild.parentBuildId || [])]));
 
     // FIXME: Is this needed ? Why not update once in handleNewBuild()
     return nextBuild.update();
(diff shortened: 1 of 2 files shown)


ibu1224 (approved):

plugins/builds/index.js
```diff
@@ -683,6 +683,31 @@ async function createJoinObject(nextJobs, current, eventFactory) {
...
+
+        if (
+            parentBuilds[joinInfo.externalPipelineId] &&
+            parentBuilds[joinInfo.externalPipelineId].jobs[joinInfo.externalJobName]
+        ) {
+            parentBuildIds.push(parentBuilds[joinInfo.externalPipelineId].jobs[joinInfo.externalJobName]);
+        }
+    }
+
+    return Array.from(new Set([currentBuildId, ...parentBuildIds]));
```
adong, line 708: Okay, so using `Set` prevents the duplicated ids.

adong (approved):

kumada626 (approved):

PR #2947, opened 2023-10-31
fix: integrate auto populate sonar projectUrl in pipeline badge [3]

adong:
## Context

1. Sonar Coverage will populate the sonar `projectUrl` only when the scope is pipeline,
2. On the backend, we will only set the pipeline sonar badge if `pipeline` itself exists and `projectUrl` is available

## Objective
Only populate if `pipelineSonarBadge.defaultName` and `pipelineSonarBadge.defaultUri` were not generated, under assumption that `defaultName` and `defaultUri` does not change once populated. 

Users can override the `pipelineSonarBadge.name` and `pipelineSonarBadge.uri` however, just not the defaults

## References

https://github.com/screwdriver-cd/data-schema/pull/540
https://github.com/screwdriver-cd/coverage-sonar/pull/73
## License
I confirm that this contribution is made under a BSD license and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #2955, opened 2023-11-17
fix: update projectUrl and add more logs

adong:
## Context

1) Use `{ projectUrl }` instead
2) Only update `projectUrl` if not using self host sonar
3) Add more logs to track which lines of code are failing

## License

### DIFF
--- a/plugins/coverage/token.js
+++ b/plugins/coverage/token.js
@@ -64,6 +64,8 @@ module.exports = config => ({
                     throw boom.notFound(`Pipeline ${pipelineId} does not exist`);
                 }
 
+                logger.info(`looking up pipeline:${pipelineId}, and found pipeline:${pipeline}`);
+
                 tokenConfig.pipelineName = pipeline.name;
             }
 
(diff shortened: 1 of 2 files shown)


plugins/coverage/token.js
```diff
@@ -82,50 +84,57 @@ module.exports = config => ({
...
+            if (!pipeline && pipelineId) {
+                pipeline = await pipelineFactory.get(pipelineId);
+
+                logger.info(`looking up again, pipeline:${pipelineId}, and found pipeline: ${pipeline}`);
+            }
+
+            if (pipeline && projectUrl) {
+                try {
+                    const pipelineSonarBadge = {
+                        defaultName: `${pipelineId}`, // ensure pipelineId is stored as String instead of Integer
```
adong, line 103: https://github.com/screwdriver-cd/data-schema/blob/2c89a71805b5700cff322a87984e2e94f5f0b61e/models/pipeline.js

klu909 (approved):

plugins/coverage/token.js
```diff
@@ -82,50 +84,57 @@ module.exports = config => ({
...
+                        defaultUri: projectUrl
+                    };
+                    let shouldPipelineUpdate = true;
+
+                    if (
```
klu909, line 108: can we combine this if branch with one at the bottom?

adong: Can you give me one example? This is what I can come up with, only update if truly needed.

adong: Look in PR https://github.com/screwdriver-cd/screwdriver/pull/2947, I have made multiple attempts to optimize regards to if logic.

image

adong: If there's a better way, I will make another PR to optimize further

Issue #2368, opened 2021-03-09
Correct the description given in the swagger


wahapo:
**What happened**:
The description given in the [swagger](https://api.screwdriver.cd/v4/documentation/) is wrong.

**What you expected to happen**:
I have found the points below. There may be others.

* [`GET /v4/pipelines/{id}/events`](https://api.screwdriver.cd/v4/documentation#/v4/getV4PipelinesIdEvents)

  The parameters `search`, `sort`, `sortBy`, and `getCount` don't seem to be used in the actual application.
  Other APIs (e.g. `GET /v4/pipelines/{id}/jobs`) seem to have the same issue.

* [`GET /v4/auth/token`](https://api.screwdriver.cd/v4/documentation#/v4/getV4AuthTokenBuildid)

  Only `/v4/auth/token/{buildId}` is listed.

* [`GET /v4/events/{id}`](https://api.screwdriver.cd/v4/documentation#/v4/getV4EventsId)

  <img width="200" src="https://user-images.githubusercontent.com/24538326/110429123-cc697580-80ed-11eb-8701-7535582276cb.png">

  The response `workflowGraph.nodes[]` actually contains the jobId as `id`.
  Other APIs (e.g. `GET /v4/pipelimes/{id}/events`) seem to have the same issue.

  `Actually`:
  ```
   "workflowGraph": {
     "nodes": [
       {
         "name": "~commit"
       },
       {
         "name": "main",
         "id": 527
       }
     ]
   },
  ```

* [`GET /v4/builds/{id}`](https://api.screwdriver.cd/v4/documentation#/v4/getV4BuildsId)

  <img width="150" src="https://user-images.githubusercontent.com/24538326/110429020-a6dc6c00-80ed-11eb-8d64-f899a72fdd19.png">

  In the case where there is only one value for `parentBuildId`, it seems that it is not actually an array.
  If it is always returned in an array, it will be sufficient.

  `Actually`:
  ```
  "parentBuildId": 5,

  OR

  "parentBuildId": [ 6, 7 ],
  ```

**How to reproduce it**:
See [swagger](https://api.screwdriver.cd/v4/documentation/).

kumada626:
I have found some more wrong points.
- [`GET /v4/builds/{id}/steps`](https://api.screwdriver.cd/v4/documentation#/v4/getV4BuildsIdSteps)
We cannot specify {id} as a parameter in swagger UI.
[...]

PR #3050, opened 2024-03-04
fix(2368): fix api validation

itleigns:
## Context
There were discrepancies between the display in Swagger and the actual specifications.

## Objective
I made several minor adjustments to ensure it is displayed correctly in Swagger. This is in response to the following comment on the issue.
```
GET /v4/builds/statuses
Default jobIds are strange and the type expected integer but got string.

GET /v4/jobs/{id}/builds
readonly is expected to be ON by default, but now is OFF.

Unnecessary getCount

sortBy does not work in these endpoint, so should not be available from the UI.

Also Not a bug, etc., but there is not enough information on examples. More detailed examples would be good.
```

## References
https://github.com/screwdriver-cd/screwdriver/issues/2368#issuecomment-1953634802

## License


test/plugins/jobs.test.js
```diff
@@ -330,20 +330,20 @@ describe('job plugin test', () => {
...
         it('returns 200 for getting builds with query params', () => {
-            options.url = `/jobs/${id}/builds?fetchSteps=false&readOnly=true`;
+            options.url = `/jobs/${id}/builds?fetchSteps=false&readOnly=false`;
 
             return server.inject(options).then(reply => {
                 assert.equal(reply.statusCode, 200);
                 assert.calledWith(job.getBuilds, {
                     sort: 'descending',
-                    sortBy: 'createTime',
-                    readOnly: true
```
yakanechi, line 346: Is there any reason not to explicitly state readOnly:false here?

itleigns: It is because `getBuilds` is called that way.
https://github.com/screwdriver-cd/screwdriver/blob/2049371e4a22adb0aa82a172c3f604914ea907b6/plugins/jobs/listBuilds.js#L33-L35

yakanechi: Thank you. I understood it.

yakanechi (approved):

plugins/templates/listVersionsWithMetric.js
```diff
@@ -66,7 +66,9 @@ module.exports = () => ({
                 joi.object({
                     search: joi.forbidden(), // we don't support search for Template list versions with metrics,
                     startTime: joi.string().isoDate(),
-                    endTime: joi.string().isoDate()
+                    endTime: joi.string().isoDate(),
```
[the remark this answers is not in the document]
itleigns: Thank you. I added it.

tkyi (approved):

Issue #3056, opened 2024-03-08
Backend change to support marking pipeline template as trusted


VonnyJap:
**What happened**:
There is no API support as of now to mark a pipeline template as trusted

**What you expected to happen**:
Required change on the backend to be able to mark a pipeline template as trusted

PR #3060, opened 2024-03-14
feat(3056): add endpoint to mark pipeline template as trusted

klu909:
## Context

add endpoint to mark pipeline template as trusted
## Objective

add endpoint to mark pipeline template as trusted
## References
https://github.com/screwdriver-cd/screwdriver/issues/3056

## License

### DIFF
--- a/plugins/pipelines/templates/updateTrusted.js
+++ b/plugins/pipelines/templates/updateTrusted.js
@@ -1,0 +1,55 @@
+'use strict';
+
+const boom = require('@hapi/boom');
+const joi = require('joi');
+const schema = require('screwdriver-data-schema');
+const metaSchema = schema.models.templateMeta.base;
+
+module.exports = () => ({
+    method: 'PUT',
+    path: '/pipeline/templates/{namespace}/{name}/trusted',
+    options: {
+        description: "Update a pipeline template's trusted property",
+        notes: 'Returns null if successful',
+        tags: ['api', 'pipeline', 'templates', 'trusted'],
+        auth: {
+            strategies: ['token'],
+            scope: ['admin', '!guest']
+        },
+
+        handler: async (request, h) => {
+            const { namespace, name } = request.params;
+            const { pipelineTemplateFactory } = request.server.app;
+            const { trusted } = request.payload;
+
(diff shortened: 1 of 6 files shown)


tkyi (changes requested):
Can you update README: https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/pipelines/README.md

plugins/pipelines/templates/updateTrusted.js
```diff
@@ -0,0 +1,55 @@
...
+        tags: ['api', 'pipeline', 'templates', 'trusted'],
+        auth: {
+            strategies: ['token'],
+            scope: ['admin', '!guest']
+        },
+
+        handler: async (request, h) => {
+            const { namespace, name } = request.params;
+            const { pipelineTemplateFactory } = request.server.app;
+            const { trusted } = request.payload;
```
pritamstyz4ever, line 23: the trusted API should also take `version` in the payload, to mark specific versions and lower as trusted

plugins/pipelines/templates/updateTrusted.js
```diff
@@ -0,0 +1,55 @@
...
+            const pipelineTemplateMeta = await pipelineTemplateFactory.get({
+                name,
+                namespace
+            });
+
+            if (!pipelineTemplateMeta) {
+                throw boom.notFound(`Pipeline template ${namespace}/${name} does not exist`);
+            }
+
+            if (!trusted) {
```
pritamstyz4ever, line 34: if no `version` is specified then use the latest, else update with the version from payload

pritamstyz4ever (changes requested):

plugins/pipelines/templates/updateTrusted.js
klu909: resolving this as per discussion in slack

pritamstyz4ever (approved):
Let's stick to marking latest as trusted as it will be done by SD Admin

tkyi (approved):

tkyi (changes requested):
Can you add a test for when `trusted: false`

tkyi (approved):

PR #3095, opened 2024-04-23
fix: 

VonnyJap:
## Context

The feature of [subscribed-scm-notifications](https://docs.screwdriver.cd/user-guide/configuration/workflow#subscribed-scm-notifications) is defective as such that: 

- when a repo is being subscribed by another downstream pipeline, the PR job never start
- a downstream pipeline is triggered regardless of event types of the upstream repo, for example supposed this is the config
```
subscribe:
    scmUrls:
        - git@github.com:supra08/functional-workflow.git: ['~pr']
 jobs:
    A:
        steps:
            - echo: echo test
        requires: [~pr, ~commit, ~subscribe]
```
With the current implementation, job A is triggered even during the commit event of `git@github.com:supra08/functional-workflow.git`. 

## License


plugins/webhooks/helper.js
```diff
@@ -365,7 +364,43 @@ async function triggeredPipelines(
...
+    // process the pipelinesWithSubscribedRepos only when the pipelinesOnCommitBranch is not empty
+    // pipelinesOnCommitBranch has the information to determine the triggering event of downstream subscribing repo
+    pipelinesWithSubscribedRepos.forEach(p => {
+        if (!Array.isArray(p.subscribedScmUrlsWithActions)) {
+            return;
+        }
+        p.subscribedScmUrlsWithActions.forEach(subscribedScmUriWithAction => {
+            const { scmUri: subscribedScmUri, actions: subscribedActions } = subscribedScmUriWithAction;
+
+            if (pipelinesOnCommitBranch[0].scmUri === subscribedScmUri) {
```
pritamstyz4ever, line 380: should check length of arr when indexing

plugins/webhooks/helper.js
```diff
@@ -462,13 +498,11 @@ async function createPREvents(options, request) {
...
                 let eventConfig = {};
 
-                // Check if the webhook event is from a subscribed repo and
-                // and fetch the source repo commit sha and save the subscribed sha
                 if (uriTrimmer(scmConfig.scmUri) !== uriTrimmer(p.scmUri)) {
                     subscribedConfigSha = sha;
 
                     try {
-                        sha = await pipelineFactory.scm.getCommitSha({
+                        configPipelineSha = await pipelineFactory.scm.getCommitSha({
```
pritamstyz4ever, line 505: `configPipelineSha` is not defined

pritamstyz4ever (commented):
some comments

plugins/webhooks/helper.js
VonnyJap: it is initialized https://github.com/screwdriver-cd/screwdriver/blob/c3d7ce4a9a22a2e98233fcf85c3245d774f4ede2/plugins/webhooks/helper.js#L497

plugins/webhooks/helper.js
VonnyJap: these lines should be sufficient to ensure that `pipelinesOnCommitBranch` at least is a length of 1. 
https://github.com/screwdriver-cd/screwdriver/blob/c3d7ce4a9a22a2e98233fcf85c3245d774f4ede2/plugins/webhooks/helper.js#L367-L369

VonnyJap: and I have also added a comment that the subsequent block will only proceed when `pipelinesOnCommitBranch` is a length of at least 1. 

https://github.com/screwdriver-cd/screwdriver/blob/c3d7ce4a9a22a2e98233fcf85c3245d774f4ede2/plugins/webhooks/helper.js#L371-L372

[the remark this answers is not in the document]
VonnyJap: I think as it stands, there is no reusable purpose for this section that can bring a benefit of putting into a function. So I would propose to keep it as is.

VonnyJap:
@pritamstyz4ever - responded to your comments.

test/plugins/webhooks.helper.test.js
```diff
@@ -1966,35 +1997,68 @@ describe('startHookEvent test', () => {
...
+                            scmUri: 'github.com:789123:master',
+                            annotations: {},
+                            admins: {
+                                baxterthehacker: false
+                            },
+                            workflowGraph,
+                            branch: Promise.resolve('master')
+                        })
+                    ]);
+                const pipelineMock2 = _.cloneDeep(pipelineMock);
```
pritamstyz4ever, line 2012: you can write a simple deep clone instead of adding the entire lodash pkg I think

package.json
```diff
@@ -93,6 +93,7 @@
     "js-yaml": "^3.14.1",
     "jsonwebtoken": "^9.0.0",
     "license-checker": "^25.0.1",
+    "lodash": "^4.17.21",
```
pritamstyz4ever, line 96: used only in tests, move to dev-dependencies

pritamstyz4ever (approved):

PR #3111, opened 2024-05-04
fix: no lodash for cloning

VonnyJap:
## Context

Removing lodash because deep clone can be natively done in JS

## Objective

Implement deep clone natively instead of using lodash 

## References
https://github.com/screwdriver-cd/screwdriver/pull/3095#discussion_r1589653340

## License

### DIFF
--- a/test/plugins/webhooks.helper.test.js
+++ b/test/plugins/webhooks.helper.test.js
@@ -1435,7 +1435,7 @@ describe('startHookEvent test', () => {
                         branch: Promise.resolve('master')
                     })
                 ]);
-            const pipelineMock2 = { ...pipelineMock };
+            const pipelineMock2 = JSON.parse(JSON.stringify(pipelineMock));
 
             pipelineMock2.subscribedScmUrlsWithActions = [{ scmUri: 'github.com:789123:master', actions: ['commit'] }];
             pipelineFactoryMock.list
@@ -2008,7 +2008,7 @@ describe('startHookEvent test', () => {
                             branch: Promise.resolve('master')
                         })
                     ]);
-                const pipelineMock2 = { ...pipelineMock };
+                const pipelineMock2 = JSON.parse(JSON.stringify(pipelineMock));
 
                 pipelineMock2.subscribedScmUrlsWithActions = [{ scmUri: 'github.com:789123:master', actions: ['pr'] }];
                 pipelineFactoryMock.list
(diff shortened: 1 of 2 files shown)


tanoda:
Note: It's not a *deep* clone but a *shallow* clone. it may not be a problem if you want to shallow clone only, but please change title/description.

VonnyJap:
> Note: It's not a _deep_ clone but a _shallow_ clone. it may not be a problem if you want to shallow clone only, but please change title/description. deep cloning in native way: https://developer.mozilla.org/en-US/docs/Web/API/structuredClone

@tanoda - thank you for the feedback. The next commit implements deep cloning.

test/plugins/webhooks.helper.test.js
```diff
@@ -1436,7 +1435,7 @@ describe('startHookEvent test', () => {
                         branch: Promise.resolve('master')
                     })
                 ]);
-            const pipelineMock2 = _.cloneDeep(pipelineMock);
+            const pipelineMock2 = JSON.parse(JSON.stringify(pipelineMock));
```
minghay, line 1438: Using `structuredClone()` would be cleaner than having to force everything to be JSON compatible.
```suggestion
            const pipelineMock2 = structuredClone(pipelineMock);
```

VonnyJap: @minghay - not sure if we need to consider older browser versions. 
Screenshot 2024-05-06 at 3 25 20 PM

test/plugins/webhooks.helper.test.js
```diff
@@ -2009,7 +2008,7 @@ describe('startHookEvent test', () => {
                             branch: Promise.resolve('master')
                         })
                     ]);
-                const pipelineMock2 = _.cloneDeep(pipelineMock);
+                const pipelineMock2 = JSON.parse(JSON.stringify(pipelineMock));
```
tkyi, line 2011: Also should use structured clone here

tkyi (changes requested):

minghay:
@VonnyJap Looks like the object that's getting cloned has some `Promises` stubbed out via `sinon`, which is why the `structuredClone` isn't working.  In this case, you're probably better off keeping it as a shallow clone like:
```js
const pipelineMock2 = Object.assign({}, pipelineMock);
```

VonnyJap:
@minghay @tkyi please check.

minghay (approved):

Issue #3027, opened 2024-02-12
Skip execution of implicit jobs created for a Stage


VonnyJap:
**What happened**:
Currently the implicit `setup` and `teardown` jobs created for a stage will invoke execution and hence running unnecessary builds.  

**What you expected to happen**:
Skip the execution of implicit `setup` and `teardown` jobs created for a stage to save runtime and build resources. 

**Issues with similar purpose**:
https://github.com/screwdriver-cd/screwdriver/issues/1449
https://github.com/screwdriver-cd/screwdriver/issues/2470

VonnyJap:
closing this as it will be solved when https://github.com/screwdriver-cd/screwdriver/issues/2470 is completed.

PR #3131, opened 2024-06-04
feat(2470): Skip execution of virtual jobs

sagar1312:
## Context

A job can be marked as virtual by adding the annotation `screwdriver.cd/virtualJob: true`. Such jobs should not be executed while processing the workflow for an event.

## Objective

Skip queueing and execution of jobs marked as virtual with annotation `screwdriver.cd/virtualJob: true`.
Status of builds of such jobs should be set to "SUCCESS" when the build is created.
Builds of its downstream jobs must be created to continue the processing of the workflow. 

## References

https://github.com/screwdriver-cd/screwdriver/issues/3027

## License


pritamstyz4ever (commented):
Looks fine, but we need to add more tests, coverage has gone down.

plugins/builds/triggers/joinBase.js
```diff
@@ -103,9 +104,10 @@ class JoinBase {
             done,
             hasFailure,
             newBuild,
-            nextJobName,
+            jobName: nextJobName,
```
tkyi, line 107: good catch

tkyi:
Agree with Pritam. LGTM, probably just need to add a test case for `// Bypass execution of the build if the job is virtual` in orBase.js

tkyi (approved):

PR #3122, opened 2024-05-14
fix(3012 ): Refactoring builds triggers

y-oksaku:
## Context

The trigger processing is complex, making it difficult to modify some aspects of the trigger processing independently from others.
As a result, fixing bugs and unexpected behavior becomes challenging.

## Objective

This PR refactors the build trigger processing as follows: https://github.com/screwdriver-cd/screwdriver/issues/3012

- Separate each trigger into a class: OR trigger, AND trigger, RemoteTrigger, and RemoteJoin.
- Simplify these trigger processes.
- Move functions to a helper file.
- Add more tests for triggers.

Now, we can trigger a single remote join in the downstream (e.g. `requires: [ sd@1:a ]`).
However, some existing strange behaviors were found, and these have not been fixed. (Refer to the comments of https://github.com/screwdriver-cd/screwdriver/issues/3012)

## References

- Issue: https://github.com/screwdriver-cd/screwdriver/issues/3012
  - There are sub PRs that have been refactoring.

If something happened, please mention to us ([kumada626](https://github.com/kumada626), [yk634](https://github.com/yk634), [yakanechi](https://github.com/yakanechi), [y-oksaku](https://github.com/y-oksaku)).

## License
I confirm that this contribution is made under a BSD license and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #3132, opened 2024-06-06
fix(3012): Trigger remote join job in restart when the last remote join job is successful

yakanechi:
## Context

Followup this PR
https://github.com/screwdriver-cd/screwdriver/pull/3122

Remote Join job should be triggered on the new event on upstream when restart from an event downstream of the event in which Remote Join is successful.
However, the event and build are created but not triggered.

## Objective

Fix the function call location so that the build can be fetched and triggered from the new event that was created.
In addition, add a test for this scenario. (In other words, make sure that it is triggered twice in total by start and restart)

## License

### DIFF
--- a/test/plugins/trigger.test.js
+++ b/test/plugins/trigger.test.js
@@ -1652,7 +1652,7 @@ describe('trigger tests', () => {
 
         const downstreamRestartEvent = await downstreamEvent.restartFrom('b');
 
-        assert.equal(await downstreamRestartEvent.getBuildOf('a'), null);
+        assert.isNull(downstreamRestartEvent.getBuildOf('a'));
         await downstreamRestartEvent.getBuildOf('b').complete('SUCCESS');
 
         const upstreamRestartEvent = upstreamPipeline.getLatestEvent();
--- a/test/plugins/trigger.test.js
+++ b/test/plugins/trigger.test.js
@@ -2596,7 +2596,7 @@ describe('trigger tests', () => {
         assert.equal(upstreamPipeline.getBuildsOf('target').length, 1);
     });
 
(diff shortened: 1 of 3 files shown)


test/plugins/trigger.test.js
```diff
@@ -1634,6 +1634,34 @@ describe('trigger tests', () => {
...
+        await upstreamEvent.run();
+
+        const downstreamEvent = downstreamPipeline.getLatestEvent();
+
+        await downstreamEvent.getBuildOf('a').complete('SUCCESS');
+        await downstreamEvent.getBuildOf('b').complete('SUCCESS');
+
+        const downstreamRestartEvent = await downstreamEvent.restartFrom('b');
+
+        assert.equal(await downstreamRestartEvent.getBuildOf('a'), null);
```
y-oksaku, line 1655:
```suggestion
        assert.isNull(downstreamRestartEvent.getBuildOf('a'));
```

test/plugins/trigger.test.js
```diff
@@ -2061,6 +2142,35 @@ describe('trigger tests', () => {
...
+
+        const downstreamEvent = downstreamPipeline.getLatestEvent();
+
+        await downstreamEvent.getBuildOf('a').complete('SUCCESS');
+        await downstreamEvent.getBuildOf('b').complete('SUCCESS');
+        await downstreamEvent.getBuildOf('c').complete('SUCCESS');
+
+        const downstreamRestartEvent = await downstreamEvent.restartFrom('c');
+
+        assert.equal(await downstreamRestartEvent.getBuildOf('b'), null);
```
y-oksaku, line 2164:
```suggestion
        assert.isNull(downstreamRestartEvent.getBuildOf('b'));
```

y-oksaku (approved):

tkyi (approved):

PR #3139, opened 2024-06-25
Debloated pkg

SandZn:
## Context

Add a debloated version of package.json file, excluding unneeded dependencies that could bring extra maintenance costs and security risks.

## Objective

During the test runtime, through our dynamic analysis, we tracked the OS system and found that 18 direct dependencies are installed, however, unused during the testing runtime.

So we figured out these dependencies, and try to remove them.

## References

No references

## License


tanoda:
Some parts of this change will break [the container uploaded to DockerHub](https://github.com/screwdriver-cd/screwdriver/blob/master/Dockerfile). and [sd-in-a-bos](https://github.com/screwdriver-cd/screwdriver?tab=readme-ov-file#using-in-a-box).
So, all (or most) `screwdriver-*` dependencies should be kept for installation.

SandZn:
Thanks for the response!

> Some parts of this change will break [the container uploaded to DockerHub] (https://github.com/screwdriver-cd/screwdriver/blob/master/Dockerfile). and [sd-in-a-bos](https://github.com/screwdriver-cd/screwdriver?tab=readme-ov-file#using-in-a-box). 

I see. It seems that these dependencies are environment-related. Why not move these corresponding dependencies to `devDependencies`?


> So, all (or most) `screwdriver-*` dependencies should be kept for installation.
Maybe a few of them are never needed, but are not removed from `package.json`.

PR #3140, opened 2024-07-03
fix: The first admin's GitHub rate limit is consumed excessively

y-oksaku:
## Context

Currently, SD API uses pipeline admin's token due to access GitHub when start builds.
[The first of `pipeline.admins` is used as pipeline admin](https://github.com/screwdriver-cd/models/blob/4d6d9e7dfa5425a9ddd2e73625de623d370c1377/lib/pipeline.js#L1181), but admins list is not sorted anywhere.
As a result, the first admin's token is always used even if other admin started builds.

There is no problem to check user's permission, but the first admins's GitHub API rate limit is consumed excessively.

## Objective

Put current user at the head of pipeline admins when update pipeline admins.
This sort update is applied when start builds, thus other admin's token will not be used.

## License

### DIFF
--- a/plugins/events/index.js
+++ b/plugins/events/index.js
@@ -27,10 +27,10 @@ const eventsPlugin = {
          * @return {Promise}                Updates the pipeline admins and throws an error if not an admin
          */
         server.expose('updateAdmins', ({ permissions, pipeline, username }) => {
-            const newAdmins = pipeline.admins;
-
             // Delete user from admin list if bad permissions
             if (!permissions.push) {
+                const newAdmins = pipeline.admins;
+
                 delete newAdmins[username];
                 // This is needed to make admins dirty and update db
                 pipeline.admins = newAdmins;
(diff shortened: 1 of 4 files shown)


plugins/events/index.js
```diff
@@ -40,16 +40,18 @@ const eventsPlugin = {
...
-                pipeline.admins = newAdmins;
+            // Put current user at the head of admins to use its SCM token after this
+            // SCM token is got from the first pipeline admin
+            const newAdminNames = [username, ...Object.keys(pipeline.admins)];
+            const newAdmins = {};
 
-                return pipeline.update();
-            }
+            newAdminNames.forEach(name => {
+                newAdmins[name] = true;
```
y-oksaku, line 49: `username` is string, so `Object.keys` order is insertion order.
https://262.ecma-international.org/6.0/#sec-ordinary-object-internal-methods-and-internal-slots-ownpropertykeys

yakanechi (approved):

tkyi (approved):

Issue #2957, opened 2023-11-21
Strange behavior of build when AND condition is satisfied


ibu1224:
**What happened**:
As depicted in the image below, when the AND condition is met and executed, the build is displayed inaccurately.

`Parent event`
<img width="289" alt="スクリーンショット 2023-11-21 11 40 35" src="https://github.com/screwdriver-cd/screwdriver/assets/17828065/51b0d3e3-6776-44ca-89a5-3afe3407eb55">

`Child event`
<img width="292" alt="スクリーンショット 2023-11-21 11 40 44" src="https://github.com/screwdriver-cd/screwdriver/assets/17828065/a65bb855-ba22-4c9b-b713-3a791eeeddd0">

I thought the event that TARGET runs was the event that restarted the D build. I think that the matter shown in the parent event is not the intended behavior, as the status appears to be failing on the UI.

`Event tab`
<img width="333" alt="スクリーンショット 2023-11-21 14 25 22" src="https://github.com/screwdriver-cd/screwdriver/assets/17828065/d00fbd56-23c8-447b-940d-68fa46d10733">

This is not a UI rendering issue.
It is probably a problem with the EventId associated with the build creation.

Is this expected behavior?  Do you think this is a bug?

**What you expected to happen**:
TARGET builds should be displayed on
child events, which can be confusing on the user interface as it is linked to failed events.

**How to reproduce it**:
pipeline: https://cd.screwdriver.cd/pipelines/12604/events/778589

1. Create a TARGET pipeline by joining jobs A, B, C, and D. 
2. Set a delay for Job A before execution.
3. Deliberately trigger failure for Job D. Then, restart Job D's build.
4. Ensure successful completion of Job D before proceeding with Job A, and guarantee the success of Job A as well.
5.  draw the TARGET job to the parent event.

y-oksaku:
The most recently completed build within the JOIN list triggers the next jobs.
Therefore, the last completed build `A` triggers the `TARGET` job.
Subsequently, the `TARGET` build starts because all statuses of the latest build within the same event group are `SUCCESS`.
As the `TARGET` build is executed within the same event, the final state will appear as depicted in the image.

This behavior is not a bug but is based on the specification where the last completed build within the JOIN list triggers the next jobs.
The same kind of display occurs when using OR conditions. (ex. `requires: [ failure1, failure2, ~success ]`)

While the UI display may indeed appear peculiar, to facilitate user understanding of This specification, I think it would be better to implement the following:

[...]

PR #3141, opened 2024-07-05
fix(2957): Fix event where a restarted join build is run

kumada626:
## Context

This PR fixes some cases of the [issue](https://github.com/screwdriver-cd/screwdriver/issues/2957).

It corrects the join build to be properly restarted and executed on the event in the following two test cases that have been added:
- In the case where `[a, b]` are in the requires, `a` fails but succeeds after `a` restart, and then `b` succeeds.
- In the case where `[a, b, c]` are in the requires, `a` fails but succeeds after `a` restart, and then `b` and `c` succeed.

However, the following test case, which is also exemplified in the original issue, still requires fixing:
- In the case where `[a, b, c]` are in the requires, `a` fails, `a` is restarted, then `b` succeeds, `a` succeeds, and finally `c` succeeds.

## Objective

Find trigger build in child event after current event has no trigger build.

## References

#2957 

## License

### DIFF
--- a/test/plugins/trigger.test.js
+++ b/test/plugins/trigger.test.js
@@ -931,6 +931,7 @@ describe('trigger tests', () => {
         await event.getBuildOf('b').complete('SUCCESS');
         await event.getBuildOf('c').complete('SUCCESS');
 
+        assert.isNull(event.getBuildOf('target'));
         assert.equal(restartEvent.getBuildOf('target').status, 'RUNNING');
     });
 
@@ -950,6 +951,7 @@ describe('trigger tests', () => {
         await restartEvent.getBuildOf('a').complete('SUCCESS');
         await event.getBuildOf('c').complete('SUCCESS');
 
+        assert.isNull(event.getBuildOf('target'));
         assert.equal(restartEvent.getBuildOf('target').status, 'RUNNING');
     });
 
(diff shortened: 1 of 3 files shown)


y-oksaku (approved):

VonnyJap (dismissed):

test/plugins/trigger.test.js
```diff
@@ -889,6 +915,44 @@ describe('trigger tests', () => {
...
+        await event.getBuildOf('hub').complete('SUCCESS');
+        await event.getBuildOf('a').complete('FAILURE');
+
+        const restartEvent = await event.restartFrom('a');
+
+        await restartEvent.getBuildOf('a').complete('SUCCESS');
+        await event.getBuildOf('b').complete('SUCCESS');
+        await event.getBuildOf('c').complete('SUCCESS');
+
+        assert.equal(restartEvent.getBuildOf('target').status, 'RUNNING');
```
yakanechi, line 934: In `[ a, b ] is triggered~` test above there is the following assert, is it not needed here? (Same for the following xit test)
assert.isNull(event.getBuildOf('target'));

kumada626: Add assert.

yakanechi: thank you for your fix.

yakanechi (approved):

tkyi (approved):

Issue #3012, opened 2024-02-04
Need to refactor regarding build triggers


yakanechi:
**What happened**:
The trigger processing has become more complex, and different trigger processes are intertwined in the code, making it very expensive to identify and fix problems that occur around triggers.
Also, when a bug related to a trigger is fixed, a new bug is likely to occur. (The latest example is #2980.)

These problems are especially likely to occur in complex conditions such as scenarios that combine external triggers and restarts.
However, since both the restart process and the external trigger process are written together with the normal trigger process, it is difficult to perform stand-alone tests.

**What you expected to happen**:
The code should be highly readable and easy to maintain.
In addition, it is desirable to have completely separate flows for each trigger.
Especially this file: [plugins/builds/index.js](https://github.com/screwdriver-cd/screwdriver/blob/1d215ecf5b2428b482a7cb300433f361b0daf846/plugins/builds/index.js)

**How to reproduce it**:

yakanechi:
LY team is considering refactoring the [plugins/builds/index.js](https://github.com/screwdriver-cd/screwdriver/blob/1d215ecf5b2428b482a7cb300433f361b0daf846/plugins/builds/index.js) file over the coming weeks.
The following is what we are considering at this time

1. Creating a triggers directory in the plugins/builds directory and consolidating all trigger processing in that directory.
2. Split the processes such as And, OR, Remote Join, Remote trigger, etc. into separate files for each function.
3. Split a huge function into several smaller functions.
4. Add tests for changed/added functions.

We do not plan to change any other files in this refactoring.
Any comments on the task and timing of the work? Thank you.

yakanechi:
LY team is considering refactoring the [plugins/builds/index.js](https://github.com/screwdriver-cd/screwdriver/blob/1d215ecf5b2428b482a7cb300433f361b0daf846/plugins/builds/index.js) file over the coming weeks.
The following is what we are considering at this time

1. Creating a triggers directory in the plugins/builds directory and consolidating all trigger processing in that directory.
2. Split the processes such as And, OR, Remote Join, Remote trigger, etc. into separate files for each function.
3. Split a huge function into several smaller functions.
4. Simplify complex functions to the extent possible (Shallow nesting, etc.)
5. Add tests for changed/added functions.

We do not plan to change any other files in this refactoring.
Any comments on the task and timing of the work? Thank you.

yakanechi:
@tkyi @yakanechi 
[...]

PR #3142, opened 2024-07-05
fix(3012): Fixed error handling of external trigger 

yakanechi:
## Context

There is a problem that remote trigger is not triggered when changed #3122
When a pipeline is deleted, the job and trigger information associated with that pipeline may remain in rare cases.
The error occurs if this deleted pipeline has jobs triggered from other pipelines like `requires: [ ~sd@1:build ]`.

Specifically, in the figure below, the 34 pipelines have already been deleted from the DB, but the job information remains in the DB and is displayed as such.
In this case, the build job 23 is triggered, but the build job 35, which was added after 34, is not triggered.

スクリーンショット 2024-07-05 16 30 47

## Objective

As before refactoring, other remote triggers are triggered successfully, even if they contain remote triggers for which no pipeline already exists
This means that in the above example, not only 23 but also 35 will be triggered.

## References
#3012

## License


test/plugins/trigger.test.js
```diff
@@ -2034,6 +2036,88 @@ describe('trigger tests', () => {
         assert.equal(upstreamPipeline.getBuildsOf('target').length, 2);
     });
 
+    it('[ ~sd@2:a ] and [ ~sd@3:a ], [ ~sd@2:a ] is triggered and [ ~sd@3:a ] is not triggerd when sd@3 pipeline is deleted from DB', async () => {
```
y-oksaku, line 2039: I think these tests should be contained in `builds.test.js` since it seems better to throw exception explicitly.

Ex.
https://github.com/screwdriver-cd/screwdriver/blob/ec99cc8b486bf71a85676322715c8559df52eb9a/test/plugins/builds.test.js#L5320

I do not think it is good to put any tests in `trigger.test.js`.

plugins/builds/index.js
```diff
@@ -224,76 +224,87 @@ async function triggerNextJobs(config, app) {
...
+                    err
+                );
+            }
         }
 
-        for (const [nextJobName, nextJob] of Object.entries(joinedPipeline.jobs)) {
-            const nextJobId = nextJob.id || (await getJobId(nextJobName, currentPipeline.id, jobFactory));
-            const { isVirtual: isNextJobVirtual, stageName: nextJobStageName } = nextJob;
+        // Skip trigger process if createExternalEvent fails
+        if (externalEvent) {
```
yakanechi, line 239: Below here, only the indentation is changed by this `if`

test/plugins/trigger.test.js
yakanechi: I fixed it!

y-oksaku (approved):

tkyi (approved):

Issue #3143, opened 2024-07-10
Provide a feature to allow job in the middle of a stage to be restarted


ThehamzaIftikhar:
**What happened**:

Users are not allowed to start/restart the event from a job that is in the middle of a stage.

**What you expected to happen**:

Users are able to start/restart the event from a job that is in the middle of a stage.

**How to reproduce it**:

Try starting/restarting an event from a job that is in a stage and not a stage setup.

PR #3145, opened 2024-07-11
feat(3143): Prevent event creation when started from a stage teardown

ThehamzaIftikhar:
## Context

Not allow users to start an event form a stage teardown

## Objective

throw an error if the user tries to start an event from a stage teardown

## References

[https://github.com/screwdriver-cd/screwdriver/issues/3143](url)

## License

### DIFF
--- a/plugins/events/create.js
+++ b/plugins/events/create.js
@@ -4,7 +4,7 @@ const urlLib = require('url');
 const boom = require('@hapi/boom');
 const validationSchema = require('screwdriver-data-schema');
 const ANNOT_RESTRICT_PR = 'screwdriver.cd/restrictPR';
-const { getScmUri } = require('../helper');
+const { getScmUri, isStageTeardown } = require('../helper');
 
 module.exports = () => ({
     method: 'POST',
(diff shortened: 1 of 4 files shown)


plugins/helper.js
```diff
@@ -3,6 +3,7 @@
 const boom = require('@hapi/boom');
 const dayjs = require('dayjs');
 const STAGE_PREFIX = 'stage@';
+const STAGE_TEARDOWN_PATTERN = /^stage@([\w-]+)(?::teardown)$/;
```
tkyi, line 6: I think this would also work, but no strong preference
```suggestion
const STAGE_TEARDOWN_PATTERN = /^stage@([\w-]+):teardown$/;
```

plugins/events/create.js
```diff
@@ -29,6 +29,11 @@ module.exports = () => ({
             let { pipelineId, startFrom, parentBuildId, parentBuilds, groupEventId, parentEventId, prNum } =
                 request.payload;
 
+            // Validation: Prevent event creation of startFrom is a stage teardown and parentEventID does not exist (start case)
```
VonnyJap, line 32:
```suggestion
            // Validation: Prevent event creation if startFrom is a stage teardown and parentEventID does not exist (start case)
```

VonnyJap: nitpick

tkyi:
need to fix tests

test/plugins/events.test.js
```diff
@@ -632,6 +632,58 @@ describe('event plugin test', () => {
             });
         });
 
+        it('returns 400 bad request error when startFrom is teardown', () => {
```
sagar1312, line 635: minor suggestion

```suggestion
        it('returns 400 bad request error when startFrom is stage teardown and parent event is not specified', () => {
```

sagar1312 (approved):

Issue #3156, opened 2024-07-31
Create a new endpoint to allow deletion of specific version for a command


VonnyJap:
**What happened**:

While there is API endpoints to [delete a command](https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/commands/remove.js) and [a tag from a command](https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/commands/removeTag.js), there is no endpoint to allow the deletion of specific version of a command. 

**What you expected to happen**:

There should be an endpoint to allow deletion of specific version of a command, just like the [job template](https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/templates/removeVersion.js) and [pipeline template](https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/pipelines/templates/removeVersion.js). 

**How to reproduce it**:
This is a new implementation hence no way to reproduce.

PR #3159, opened 2024-08-01
feat(3156): New endpoint to delete specific version of a command

VonnyJap:
## Context

While there is API endpoints to [delete a command](https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/commands/remove.js) and [a tag from a command](https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/commands/removeTag.js), there is no endpoint to allow the deletion of specific version of a command. And it will be useful to allow specific command and its associated tag to be deleted because bad commits happen sometimes. 

## Objective

This PR implements a new endpoint to allow specific version of a command to be deleted. 

## References

https://github.com/screwdriver-cd/screwdriver/issues/3156

## License

### DIFF
--- a/plugins/commands/README.md
+++ b/plugins/commands/README.md
@@ -1,4 +1,5 @@
 # Commands Plugin
+
 > Hapi commands plugin for the Screwdriver API
 
 ## Usage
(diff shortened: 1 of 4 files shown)


tkyi (changes requested):
LGTM but can you update the README
https://github.com/screwdriver-cd/screwdriver/tree/master/plugins/commands#readme

tkyi (approved):

PR #3162, opened 2024-08-05
Skip execution of jobs between stage setup and event startFrom

ThehamzaIftikhar:
## Context

The job after the stage setup was being executed regardless of the startFrom being from the middle of stage or not

## Objective

The startFrom job will be triggered next if the current job is stage setup and the startFrom is a non setup job in that stage

## References

https://github.com/screwdriver-cd/screwdriver/issues/3143

## License

### DIFF
--- a/plugins/builds/triggers/helpers.js
+++ b/plugins/builds/triggers/helpers.js
@@ -6,6 +6,7 @@ const merge = require('lodash.mergewith');
 const schema = require('screwdriver-data-schema');
 const { EXTERNAL_TRIGGER_ALL } = schema.config.regex;
 const { getFullStageJobName } = require('../../helper');
+const STAGE_SETUP_PATTERN = /^stage@([\w-]+)(?::setup)$/;
 
 /**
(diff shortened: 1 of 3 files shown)


tkyi (approved):

plugins/builds/triggers/helpers.js
```diff
@@ -1063,6 +1064,38 @@ function buildsToRestartFilter(joinPipeline, groupEventBuilds, currentEvent, cur
...
+}
+
+/**
+ * Check if the current job is a stage setup and the next job is a non-setup stage job
+ * @param {String} triggerJob                Current job
+ * @param {String} startFrom                 Next job
+ * @param {Object} workflowGraph             Workflow Graph
+ * @return {Boolean}
+ */
+function isStartFromMiddleOfStage(triggerJob, startFrom, workflowGraph) {
```
sagar1312, line 1095: We also need to ensure `triggerJob` and `startFrom` belongs to the same stage.
Refer https://github.com/screwdriver-cd/workflow-parser/pull/53#discussion_r1704565605

PR #3166, opened 2024-08-07
fix(3143): skip jobs between stage setup and startFrom stage job only if they belong to the same stage

ThehamzaIftikhar:
## Context

The isStartFromMiddleOfStage function does not check if the startFrom job and the stage setup belong to the same stage

## Objective

the function will check that both the current job and startFrom belong to the same stage if the ortrigger is to be executed for this scenario

## References

https://github.com/screwdriver-cd/screwdriver/issues/3143

## License


plugins/builds/triggers/helpers.js
```diff
@@ -1074,26 +1073,35 @@ function isStageSetup(jobName) {
...
 
 /**
- * Check if the current job is a stage setup and the next job is a non-setup stage job
+ * Check if the current job is a stage setup and the next job is a non-setup job in the same stage
  * @param {String} triggerJob                Current job
  * @param {String} startFrom                 Next job
  * @param {Object} workflowGraph             Workflow Graph
  * @return {Boolean}
  */
 function isStartFromMiddleOfStage(triggerJob, startFrom, workflowGraph) {
```
sagar1312, line 1094: Can we rename the function name and args/variables
`isStartFromMiddleOfStage` -> `isStartFromMiddleOfCurrentStage`
`triggerJob` -> `currentJobName`
and may be `startFrom` -> `eventStartFrom`

plugins/builds/triggers/helpers.js
```diff
@@ -1074,26 +1073,35 @@ function isStageSetup(jobName) {
...
-    return isStageSetup(triggerJob) && !isStageSetup(startFrom) && isStageJob(workflowGraph, startFrom);
+    if (isStageSetup(triggerJob)) {
+        const startFromStageName = getStageName(workflowGraph, triggerJob);
+        const triggerStageName = getStageName(workflowGraph, triggerJob);
```
sagar1312, line 1097:
```suggestion
        const currentStageName = getStageName(workflowGraph, currentJobName);
```

sagar1312 (changes requested):
Add a test to cover the scenario where `event.startFrom` is from a different stage that the stage of the current job (stage setup)

sagar1312 (approved):

Issue #3174, opened 2024-08-21
Stages - metadata passed to stage teardown is incomplete when there is failure in any stage job


sagar1312:
**What happened**:
When a build for any stage job fails, build is created for stage `teardown` if it is not already created. But `parentBuildId` and `parentBuilds` attributes are not set/updated for the `teardown` build. Due to the missing of reference to parent builds, when metadata is computed for `teardown` build it is incomplete.

**What you expected to happen**:
Set `parentBuildId` and `parentBuilds` attributes when the build for stage `teardown` is created on stage failure.
Update `parentBuildId` and `parentBuilds as needed when the build for stage `teardown` already exists and builds for other stage jobs complete.

**How to reproduce it**:
Refer to this pipeline
https://cd.screwdriver.cd/pipelines/14896/events/816743
<img width="1212" alt="image" src="https://github.com/user-attachments/assets/4e497ae6-f36f-4463-bbe8-ba5a5ec25eb2">

`parentBuildId` and `parentBuilds` should include reference to both `ci-test-functional` and `ci-test-integration`.
But only reference to `ci-test-functional` exists.

<img width="826" alt="image" src="https://github.com/user-attachments/assets/6063637d-dc17-4b69-9ae8-037aaf49bb0c">

PR #3175, opened 2024-08-21
fix(3174): Stages - metadata passed to stage teardown is incomplete when there is failure in any stage job

sagar1312:
## Context

When a build for any stage job fails, build is created for stage `teardown` if it is not already created. But `parentBuildId` and `parentBuilds` attributes are not set/updated for the `teardown` build. Due to the missing of reference to parent builds, when metadata is computed for `teardown` build it is incomplete.

## Objective
Set `parentBuildId` and `parentBuilds` attributes when the build for stage `teardown` is created on stage failure.
Update `parentBuildId` and `parentBuilds as needed when the build for stage `teardown` already exists and builds for other stage jobs complete.

**Before:**
image

**After:**
image

## References

https://github.com/screwdriver-cd/screwdriver/issues/3174

## License


VonnyJap:
What is the key in the parentBuilds object referring to?

VonnyJap (approved):

sagar1312:
> LGTM. But wondering what is the key in the parentBuilds object referring to?

The key is the pipeline id.

Issue #3177, opened 2024-08-27
Remote join jobs are not triggered correctly on restart


yakanechi:
**What happened**:
Remote join may not be triggered correctly when restarting an event containing a remote join job.
Specifically, it is triggered as a remote trigger and executed on a different event than the group event on restart.
Also, an empty event is created for the number of downstream jobs in the trigger condition.

This especially occurs when all trigger conditions are remote jobs.
On the other hand, if there is at least one job in the same pipeline in the trigger condition, there is no problem.
(See second job in the screenshot below).

**What you expected to happen**:
Remote join job is triggered successfully on restart event.

**How to reproduce it**:
1. Run with the following configuration
2. Restart from first job

```
jobs:
  first:
    requires: [ ~commit, ~pr ]

  first2:
    requires: [ ~first ]

  second:
    requires: [ first2, sd@23:component, sd@35:component, sd@36:component ]

  third:
    requires: [ sd@23:first, sd@23:second, sd@23:third, sd@35:first, sd@35:second, sd@35:third, sd@36:first, sd@36:second, sd@36:third ]
```

Start
![start](https://github.com/user-attachments/assets/b93ba277-3a41-4d1f-bc33-29b00a46b963)
Restart event
![restart1](https://github.com/user-attachments/assets/d06e58fb-e39d-41de-86fc-90ad742c4727)
Last remote join job event
![restart2](https://github.com/user-attachments/assets/96d08b12-2c49-457f-9809-9ed3e688cebd)

PR #3178, opened 2024-08-27
fix(2981): remote join wait on downsream jobs during restart

### LINKED ISSUE #2981: Remote-join job doesn't wait on downstream jobs during restart
**What happened**:

When an upstream job of the remote-join job is restarted, remote-join job immediately runs reusing build status from previous event.

https://cd.screwdriver.cd/pipelines/12907/events/784132
Screenshot 2024-01-11 at 3 33 48 PM

Job first was restarted from previous event, where remote-join in job `second` worked as expected, however in the new event, job `second` ran immediately without waiting for any of the downstream to be triggered or executed. 

**What you expected to happen**:

Job `second` in the restarted event should depend for status from newly triggered remote downstream jobs. 

**How to reproduce it**:

yakanechi:
## Context

fix #2981
When an upstream job of the remote-join job is restarted, remote-join job immediately runs reusing build status from previous event.

## Objective

When a remote join job's upstream jobs are restarted, the remote join job will use the post-restart build status as the trigger for jobs that may run from the restart job.

Enumerate jobs started after the restart job, referring to [the subgraph filter in the UI](https://github.com/screwdriver-cd/ui/blob/8a764ea1a6db38e29e85588a8cc7fcc56d770869/app/utils/graph-tools.js#L454), and do not use the recent build status of the group event.

- About the test
We found another bug related to remote join while working on this. (#3177)
Therefore, a test has been added to the “Tests for behavior not ideal” section
Also, with this fix, the test that was originally in “Tests for behavior not ideal” works as expected so that it will be removed and the test that was xit will be executed.

## References
#2981
#3177

## License

### DIFF
--- a/plugins/builds/triggers/helpers.js
+++ b/plugins/builds/triggers/helpers.js
@@ -1049,12 +1049,8 @@ function subsequentJobFilter(workflowGraph, startNode) {
         return [];
     }
 
-    let start = startNode;
-
     // startNode can be a PR job in PR events, so trim PR prefix from node name
-    if (startNode.match(/^PR-[0-9]+:/)) {
-        start = startNode.split(':')[1];
-    }
+    const start = trimJobName(startNode);
 
     const visiting = [start];
 
@@ -1062,7 +1058,7 @@ function subsequentJobFilter(workflowGraph, startNode) {
 
     if (edges.length) {
         while (visiting.length) {
-            const cur = visiting.shift();
+            const cur = visiting.pop();
 
             edges.forEach(e => {
                 if (e.src === cur && !visited.has(e.dest)) {
--- a/plugins/builds/triggers/helpers.js
+++ b/plugins/builds/triggers/helpers.js
@@ -673,8 +673,8 @@ async function getParallelBuilds({ eventFactory, parentEventId, pipelineId }) {
  * @param {ParentBuilds} parentBuilds parent builds
  * @param {Build[]} relatedBuilds Related builds which is used to fill parentBuilds data
  * @param {Event} currentEvent Current event
- * @param {Event} nextEvent Next triggered event (Remote trigger or Same pipeline event triggered as external)
- * @returns {ParentBuilds} Merged parent builds { "${pipelineId}": { jobs: { "${jobName}": ${buildId} }, eventId: 123 }  }
+ * @param {string[]} ignoreJobs Names of job that may be run on current event
+ * @param {Event} nextEvent Next triggered event (Remote trigger or Same pipeline event triggered as external) * @returns {ParentBuilds} Merged parent builds { "${pipelineId}": { jobs: { "${jobName}": ${buildId} }, eventId: 123 }  }
  *
  * @example
  * >>> mergeParentBuilds(...)
@@ -689,7 +689,7 @@ async function getParallelBuilds({ eventFactory, parentEventId, pipelineId }) {
  *     },
  * }
  */
-function mergeParentBuilds(parentBuilds, relatedBuilds, currentEvent, nextEvent) {
+function mergeParentBuilds(parentBuilds, relatedBuilds, currentEvent, nextEvent, ignoreJobs) {
     const newParentBuilds = {};
 
     Object.entries(parentBuilds).forEach(([pipelineId, { jobs, eventId }]) => {
@@ -718,7 +718,6 @@ function mergeParentBuilds(parentBuilds, relatedBuilds, currentEvent, nextEvent)
                     nodeName = `sd@${pipelineId}:${nodeName}`;
                 }
             }
-
             const targetJob = workflowGraph.nodes.find(node => node.name === nodeName);
 
             if (!targetJob) {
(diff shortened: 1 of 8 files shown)


plugins/builds/triggers/helpers.js
```diff
@@ -1032,6 +1033,48 @@ function buildsToRestartFilter(joinPipeline, groupEventBuilds, currentEvent, cur
...
+    if (!startNode || !nodes.length) {
+        return [];
+    }
+
+    let start = startNode;
+
+    // startNode can be a PR job in PR events, so trim PR prefix from node name
+    if (startNode.match(/^PR-[0-9]+:/)) {
+        start = startNode.split(':')[1];
+    }
```
y-oksaku, line 1056: You can use these functions.

https://github.com/screwdriver-cd/screwdriver/blob/919680ba9b286704f211adc6caff347e028543c5/plugins/builds/triggers/helpers.js#L362-L364

https://github.com/screwdriver-cd/screwdriver/blob/919680ba9b286704f211adc6caff347e028543c5/plugins/builds/triggers/helpers.js#L371-L377

plugins/builds/triggers/helpers.js
```diff
@@ -1032,6 +1033,48 @@ function buildsToRestartFilter(joinPipeline, groupEventBuilds, currentEvent, cur
...
+
+    const visiting = [start];
+
+    const visited = new Set(visiting);
+
+    if (edges.length) {
+        while (visiting.length) {
+            const cur = visiting.shift();
```
y-oksaku, line 1064: `Array.shift` is slower than `Array.pop`.
So using `visiting.pop()` would be better if we can use `pop` method here.

plugins/builds/triggers/helpers.js
```diff
@@ -673,8 +673,8 @@ async function getParallelBuilds({ eventFactory, parentEventId, pipelineId }) {
  * @param {ParentBuilds} parentBuilds parent builds
  * @param {Build[]} relatedBuilds Related builds which is used to fill parentBuilds data
  * @param {Event} currentEvent Current event
- * @param {Event} nextEvent Next triggered event (Remote trigger or Same pipeline event triggered as external)
- * @returns {ParentBuilds} Merged parent builds { "${pipelineId}": { jobs: { "${jobName}": ${buildId} }, eventId: 123 }  }
+ * @param {Event} nextEvent Next triggered event (Remote trigger or Same pipeline event triggered as external) * @returns {ParentBuilds} Merged parent builds { "${pipelineId}": { jobs: { "${jobName}": ${buildId} }, eventId: 123 }  }
+ * @param {string[]} ignoreJobs Names of job that may be run on current event
```
y-oksaku, line 677:
```suggestion
 * @param {Event} nextEvent Next triggered event (Remote trigger or Same pipeline event triggered as external)
 * @param {string[]} ignoreJobs Names of job that may be run on current event
 * @returns {ParentBuilds} Merged parent builds { "${pipelineId}": { jobs: { "${jobName}": ${buildId} }, eventId: 123 }  }
```

plugins/builds/triggers/helpers.js
yakanechi: I fixed it!

plugins/builds/triggers/helpers.js
yakanechi: I fixed it!

plugins/builds/triggers/helpers.js
```diff
@@ -1032,6 +1034,44 @@ function buildsToRestartFilter(joinPipeline, groupEventBuilds, currentEvent, cur
...
+
+    const visiting = [start];
+
+    const visited = new Set(visiting);
+
+    if (edges.length) {
+        while (visiting.length) {
+            const cur = visiting.pop();
+
+            edges.forEach(e => {
```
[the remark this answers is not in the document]
yakanechi: As you pointed out, there were many unnecessary loops, so I reduced them.

y-oksaku (approved):
We might need to consider excluding a disabled job from subsequent jobs.
(Some users disable a job to skip executing in a restart event.)

yakanechi:
All fixes have been done.
The following fixes will be made in a new PR

> We might need to consider excluding a disabled job from subsequent jobs.
> (Some users disable a job to skip executing in a restart event.)

plugins/builds/triggers/helpers.js
```diff
@@ -668,6 +668,65 @@ async function getParallelBuilds({ eventFactory, parentEventId, pipelineId }) {
...
 
+/**
+ * Get subsequent job names which the root is the start from node
+ * @param   {Array}   [workflowGraph]         Array of graph vertices
+ * @param   {Array}   [workflowGraph.nodes]   Array of graph vertices
+ * @param   {Array}   [workflowGraph.edges]   Array of graph edges
+ * @param   {String}  [startNode]            Starting/trigger node
+ * @returns {Array<String>}                   subsequent job names
+ */
+function subsequentJobFilter(workflowGraph, startNode) {
```
tkyi, line 679: I don't feel strongly but would something like `getDownstreamJobs` or `getSubsequentJobs` be more clear?

plugins/builds/triggers/helpers.js
```diff
@@ -668,6 +668,65 @@ async function getParallelBuilds({ eventFactory, parentEventId, pipelineId }) {
...
+    const { nodes, edges } = workflowGraph;
+
+    if (!nodes.length) {
+        return [];
+    }
+
```
tkyi, line 685: can probably remove these lines since the next if statement covers the same scenario
```suggestion
```

plugins/builds/triggers/helpers.js
yakanechi: @tkyi 
Thanks for your comment.
Renamed to"getSubsequentJobs" because it is used not only for dosntream but also for normal events.

PR #3199, opened 2024-09-18
feat(1749): Add endpoint to download all zipped artifacts

tkyi:
## Context

Would be nice to be able to download all SD artifacts instead of each individual file.

## Objective

This PR adds an endpoint to download all zipped SD artifacts.

## References
N/A

## License


plugins/builds/artifacts/getAll.js
```diff
@@ -0,0 +1,102 @@
...
+                    }
+
+                    return canAccessPipeline(credentials, event.pipelineId, 'pull', req.server.app);
+                })
+                .then(async () => {
+                    const token = jwt.sign({
+                        buildId, artifact, scope: ['user']
+                    }, config.authConfig.jwtPrivateKey, {
+                        algorithm: 'RS256',
+                        expiresIn: '5s',
```
pritamstyz4ever, line 51: the expiry of 5 sec may be too short when downloading large files, consider changing to may be 10m

plugins/builds/artifacts/getAll.js
```diff
@@ -0,0 +1,102 @@
...
+                        jwtid: uuidv4()
+                    });
+                    // Create zip file and add to it
+                    const zip = new AdmZip();
+                    const baseUrl = `${config.ecosystem.store}/v1/builds/${buildId}/ARTIFACTS`;
+
+                    try {
+                        // Get manifest, figure out list of files to download
+                        const manifest = await request({
```
pritamstyz4ever, line 60: is passing the token in query param and also in header intentional?

plugins/builds/artifacts/getAll.js
```diff
@@ -0,0 +1,102 @@
...
+                            url: `${baseUrl}/manifest.txt?token=${token}`,
+                            method: 'GET',
+                            context: {
+                                token
+                            }
+                        }).text();
+                        const manifestArray = manifest.split('\n');
```
pritamstyz4ever, line 67: may be also remove any empty lines from manifestArray

plugins/builds/artifacts/getAll.js
```diff
@@ -0,0 +1,102 @@
...
+                            if (file) { // Could be an empty string
+                                const content = await request({
+                                    url: `${baseUrl}/${file}?token=${token}&type=download`,
+                                    method: 'GET',
+                                    context: {
+                                        token
+                                    }
+                                }).buffer();
+
+                                zip.addFile(file, Buffer.from(content, 'utf8'));
```
pritamstyz4ever, line 79: artifacts may also contain binary files, utf8 may not work

plugins/builds/artifacts/getAll.js
```diff
@@ -0,0 +1,102 @@
...
+                        const manifest = await request({
+                            url: `${baseUrl}/manifest.txt?token=${token}`,
+                            method: 'GET',
+                            context: {
+                                token
+                            }
+                        }).text();
+                        const manifestArray = manifest.split('\n');
+
+                        await Promise.all(manifestArray.map(async file => {
```
pritamstyz4ever, line 69: instead of reading all files in memory using zip.buffer() , we should be streaming of the files, not sure if adm-zip has streaming functionality, but you can checkout [archiver](https://www.npmjs.com/package/archiver) pkg
and use `fs.createWriteStream(zipPath);`

pritamstyz4ever (changes requested):

plugins/builds/artifacts/getAll.js
tkyi: no, good catch

plugins/builds/artifacts/getAll.js
```diff
@@ -0,0 +1,104 @@
...
+                        url: `${baseUrl}/manifest.txt?token=${token}`,
+                        method: 'GET'
+                    }).text();
+                    const manifestArray = manifest.trim().split('\n');
+
+                    // Create a stream and set up archiver
+                    const archive = archiver('zip', { zlib: { level: 9 } });
+                    // PassThrough stream to make archiver readable by Hapi
+                    const passThrough = new PassThrough();
+
```
pritamstyz4ever, line 69:
```suggestion
archive.on('error', (err) => {
    logger.error('Archiver error:', err);
    passThrough.emit('error', err); // Propagate the error to the PassThrough stream
});

passThrough.on('error', (err) => {
    logger.error('PassThrough stream error:', err);
});

```

plugins/builds/artifacts/getAll.js
```diff
@@ -0,0 +1,104 @@
...
+                    // Pipe the archive to PassThrough so it can be sent as a response
+                    archive.pipe(passThrough);
+
+                    // Fetch the artifact files and append to the archive
+                    try {
+                        for (const file of manifestArray) {
+                            if (file) {
+                                const fileStream = request.stream(`${baseUrl}/${file}?token=${token}&type=download`);
+
```
pritamstyz4ever, line 78:
```suggestion
fileStream.on('error', (err) => {
                logger.error(`Error downloading file: ${file}`, err);
                archive.emit('error', err);
            });
```

pritamstyz4ever (approved):
minor suggestions

VonnyJap (approved):

PR #3202, opened 2024-09-21
fix: need to init a value when key does not exist in an object

VonnyJap:
## Context

Observed this scenario of workflowgraph
```
const workflowGraph = {
    nodes: [
        {
            name: '~pr'
        },
        {
            name: '~commit'
        },
        {
            name: '~sd@123:deploy-db',
            id: 573500
        },
        {
            name: 'sd@123:deploy-api',
            id: 573501
        }
    ],
    edges: [
        {
            src: '~pr',
            dest: 'pr'
        },
        {
            src: '~commit',
            dest: 'component'
        },
        {
            src: '~sd@123:deploy-db',
            dest: 'certify-db'
        },
        {
            src: 'certify-db',
            dest: 'sd@123:deploy-api'
        },
        ### entry below comes from two remote pipelines
        ### is this possible?
        {
            src: '~sd@123:deploy-api',
            dest: 'sd@456:mysql'
        }
    ]
};
```

## Objective

When edges have src entry that slightly differ from the nodes entry, in this case with `~` then the key won't be found in the object causing to be undefined and hence resulting in error. 

## References

https://github.com/screwdriver-cd/screwdriver/pull/3178

## License

### DIFF
--- a/plugins/builds/triggers/helpers.js
+++ b/plugins/builds/triggers/helpers.js
@@ -686,8 +686,8 @@ function getSubsequentJobs(workflowGraph, startNode) {
     const nodeToEdgeDestsMap = Object.fromEntries(nodes.map(node => [node.name, []]));
 
     let start = trimJobName(startNode);
-    // In rare cases, WorkflowGraph and startNode may have different start tildes
 
+    // In rare cases, WorkflowGraph and startNode may have different start tildes
     if (!(start in nodeToEdgeDestsMap)) {
         if (start.startsWith('~')) {
             start = start.slice(1);
@@ -700,11 +700,18 @@ function getSubsequentJobs(workflowGraph, startNode) {
         return [];
     }
 
+    console.log('edges: ', edges);
+
     const visiting = [start];
 
     const visited = new Set(visiting);
 
-    edges.forEach(edge => nodeToEdgeDestsMap[edge.src].push(edge.dest));
+    edges.forEach(edge => {
+        if (!nodeToEdgeDestsMap[edge.src]) {
+            nodeToEdgeDestsMap[edge.src] = [];
+        }
+        nodeToEdgeDestsMap[edge.src].push(edge.dest);
+    });
     if (edges.length) {
         while (visiting.length) {
             const currentNode = visiting.pop();
--- a/plugins/builds/triggers/helpers.js
+++ b/plugins/builds/triggers/helpers.js
@@ -700,8 +700,6 @@ function getSubsequentJobs(workflowGraph, startNode) {
         return [];
     }
 
-    console.log('edges: ', edges);
-
     const visiting = [start];
 
     const visited = new Set(visiting);
--- a/plugins/builds/triggers/helpers.js
+++ b/plugins/builds/triggers/helpers.js
@@ -705,6 +705,8 @@ function getSubsequentJobs(workflowGraph, startNode) {
     const visited = new Set(visiting);
 
     edges.forEach(edge => {
+        // this is a temporary fix for the issue where the edge.src is not in the nodes array
+        // TODO: https://github.com/screwdriver-cd/screwdriver/issues/3206
         if (!nodeToEdgeDestsMap[edge.src]) {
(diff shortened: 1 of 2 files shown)


plugins/builds/triggers/helpers.js
```diff
@@ -700,11 +700,18 @@ function getSubsequentJobs(workflowGraph, startNode) {
...
     const visiting = [start];
 
     const visited = new Set(visiting);
 
-    edges.forEach(edge => nodeToEdgeDestsMap[edge.src].push(edge.dest));
+    edges.forEach(edge => {
+        if (!nodeToEdgeDestsMap[edge.src]) {
+            nodeToEdgeDestsMap[edge.src] = [];
+        }
+        nodeToEdgeDestsMap[edge.src].push(edge.dest);
```
sagar1312, line 713: This looks like a bug in the workflow graph. `edges` in the workflow graph needs to be fixed which building the workflow graph not to include `~` prefix in the `src`/`dest`.

sagar1312 (changes requested):

plugins/builds/triggers/helpers.js
sagar1312: Per discussion, either the `src` or `dest` of the edge should be a job in the current pipeline.
Long term fix would be to find/fix why we have edges with neither `src` nor `dest` is part of the pipeline.

For short term, we can have this temporary fix. Can we add a TODO/note to come back to this?

plugins/builds/triggers/helpers.js
```diff
@@ -700,11 +700,18 @@ function getSubsequentJobs(workflowGraph, startNode) {
         return [];
     }
 
+    console.log('edges: ', edges);
```
jithine, line 703: remove console log ?

VonnyJap: removed

plugins/builds/triggers/helpers.js
```diff
@@ -686,8 +686,8 @@ function getSubsequentJobs(workflowGraph, startNode) {
     const nodeToEdgeDestsMap = Object.fromEntries(nodes.map(node => [node.name, []]));
 
     let start = trimJobName(startNode);
-    // In rare cases, WorkflowGraph and startNode may have different start tildes
 
+    // In rare cases, WorkflowGraph and startNode may have different start tildes
```
yakanechi, line 690: Thanks for finding the error.
As other members have said, it looks like a bug to have job names in `name`, `src`, `dest` with `~` prefixes other than `~commit` and `~pr`.
(I have actually encountered such a case during my last PR, as this comment.)

Also, I think it is a bug that there is a`src` or `dest` with a name that is not in the `nodes` like `certify-db`, apart from the presence or absence of `~` prefix.

yakanechi, line 690: Thanks for finding the error.
As other members have said, it looks like a bug to have job names in `name`, `src`, `dest` with `~` prefixes other than `~commit` and `~pr`.
(I have actually encountered such a case during my last PR, as this comment.)

Also, I think it is a bug that there is a`src` or `dest` with a name that is not in the `nodes` like `certify-db`, apart from the presence or absence of `~` prefix.

yakanechi: >         ### entry below comes from two remote pipelines
>         ### is this possible?
>         {
>             src: '~sd@123:deploy-api',
>             dest: 'sd@456:mysql'
>         }
Regarding the above comment in the PR context, I think that it is possible if 123 or 456 is the same as the current pipeline ID.
Conversely, if neither is the same as the current pipeline ID, it would be strange for them to exist in the this workflowGraph, because they are not related to the current pipeline at all.

plugins/builds/triggers/helpers.js
VonnyJap: I put a `todo` https://github.com/screwdriver-cd/screwdriver/pull/3202/commits/aa29c13d7fe617672f742efb2380dfff8042dfd0

sagar1312 (approved):

Issue #3209, opened 2024-09-26
Virtual jobs does not honor freeze windows


sagar1312:
**What happened**:
Changes were made in this PR https://github.com/screwdriver-cd/screwdriver/pull/3131 to skip execution of job with the annotation `screwdriver.cd/virtualJob: true`.

But this change did not take freeze windows in to the consideration. Due to this, the virtual jobs are skipped execution and marked as 'SUCCESS' with in the freeze window.

**What you expected to happen**:
Queue the virtual jobs with freeze windows for execution.
Queue Service should pick this up outside the freeze window, skip the execution and mark the status as 'SUCCESS'

**How to reproduce it**:
Refer to this pipeline https://github.com/sagar1312/screwdriver-config/tree/main/virtual-job/with-freeze-window

PR #3210, opened 2024-09-26
fix(3209): Honor freeze windows for virtual jobs

sagar1312:
## Context

Changes were made in this PR https://github.com/screwdriver-cd/screwdriver/pull/3131 to skip execution of job with the annotation `screwdriver.cd/virtualJob: true`.

But this change did not take freeze windows in to the consideration. Due to this, the virtual jobs are skipped execution and marked as 'SUCCESS' with in the freeze window.

## Objective
Do not skip the execution of virtual jobs with freeze windows. Instead add it to the execution queue.
Let Queue Service skip the execution when the job is picked up from the queue outside the freeze windows.

## References

https://github.com/screwdriver-cd/screwdriver/issues/3209

## License

### DIFF
--- a/plugins/builds/index.js
+++ b/plugins/builds/index.js
@@ -28,7 +28,7 @@ const {
     createEvent,
     parseJobInfo,
     ensureStageTeardownBuildExists,
-    getJobId,
+    getJob,
     isOrTrigger,
     extractExternalJoinData,
     extractCurrentPipelineJoinData,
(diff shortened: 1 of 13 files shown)


tkyi (dismissed):

plugins/builds/index.js
```diff
@@ -244,9 +242,9 @@ async function triggerNextJobs(config, app) {
 
         // Skip trigger process if createExternalEvent fails
         if (externalEvent) {
-            for (const [nextJobName, nextJob] of Object.entries(joinedPipeline.jobs)) {
-                const nextJobId = nextJob.id || (await getJobId(nextJobName, currentPipeline.id, jobFactory));
-                const { isVirtual: isNextJobVirtual, stageName: nextJobStageName } = nextJob;
+            for (const [nextJobName, nextJobInfo] of Object.entries(joinedPipeline.jobs)) {
+                const nextJob = await getJob(nextJobName, joinedPipelineId, jobFactory);
```
sagar1312, line 246: I believe there was a long living bug which used `currentPipeline.id` instead of `joinedPipelineId` to get the job associated with external pipeline.
Probably the execution control never reached the bug as the `nextJob.id` would be always defined.

tkyi (approved):

Issue #1749, opened 2019-09-18
Can not download artifacts by dir


sugarnaoming:
**What happened**:
Can't download artifacts by dir.
It's too much hassle to manually download file one by one.

**What you expected to happen**:
Can download artifacts by dir

**How to reproduce it**:

yuichi10:
I also want a function that is able to download all artifacts at once.

PR #3215, opened 2024-10-02
feat(1749): Download directory artifacts as zip file

tkyi:
## Context

Would be nice if users could download directories of artifacts as a zip file.

## Objective

This PR checks if the user is trying to download an artifact directory and returns them.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/1749

## License


pritamstyz4ever (dismissed):
overall LGTM, but please add a few more tests to cover failure scenarios, large file downloads, timeouts

pritamstyz4ever (approved):

Issue #3176, opened 2024-08-23
Stages - teardown job is triggered even if all other jobs are not finished


kumada626:
**What happened**:
Stage teardown job is triggered soon after one of a last job in the stage is finished even if there is a running job in that stage.
In the following pipeline, `stage@sandbox:teardown` job is triggered after `sandbox-e2e-test` finishes while `sandbox-test` is still running.
![スクリーンショット 2024-08-23 14 06 28](https://github.com/user-attachments/assets/0e57c13e-2f4e-40cd-a1de-fef8a3261b26)

**What you expected to happen**:
Teardown job should be triggered after all other jobs in the stage are finished.

**How to reproduce it**:
1. provide a pipeline that has following screwdriver.yaml.
```
shared:
  image: node:20
  steps:
    - test: echo 'test'
 
stages:
  sandbox:
    requires: [ ~commit ]
    description: Deploy sandbox and test
    jobs: [ sandbox-deploy, sandbox-e2e-test, sandbox-test ]
    teardown:
      steps:
        - init: echo 'sandbox teardown'
 
jobs:
  sandbox-deploy:
    description: empty
  sandbox-e2e-test:
    requires: [ sandbox-deploy ]
  sandbox-test:
    requires: [ sandbox-deploy ]
    steps:
      - sleep: sleep 60
```
2. start this pipeline.
3. `stage@sandbox:teardown` job is triggered before `sandbox-test` job is finished.

kumada626:
@VonnyJap I check this behavior on current OSS-SD(API v7.0.188) and still occurs this bug. Like below screenshot, `stage@sandbox:teardown` job is triggered after `stage@sandbox:sandbox-e2e-test` is finished even if  `stage@sandbox:sandbox-test` still running.
<img width="440" alt="スクリーンショット 2024-09-06 15 38 23" src="https://github.com/user-attachments/assets/f920e3d7-f4f1-4464-af61-1fd68fbaab71">
If `stage@sandbox:sandbox-e2e-test` is failed, `stage@sandbox:teardown` correctly triggered after `stage@sandbox:sandbox-test` is finished. So I think this issue is caused when a job triggering teardown job is success.

VonnyJap:
@tkyi - Could you help investigate why the teardown is being triggered even though not all upstream jobs have completed?

PR #3216, opened 2024-10-02
fix(3176): Stage teardown should not run when other stage jobs are not finished

tkyi:
## Context

Stage teardown sometimes runs prematurely when a job has completed successfully even if a parallel job has not completed running.
 
## Objective

This PR checks that all builds in the stage are done running before running the stage teardown job.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/3176

## License

### DIFF
--- a/plugins/builds/update.js
+++ b/plugins/builds/update.js
@@ -182,21 +182,26 @@ async function getStage({ stageFactory, workflowGraph, jobName, pipelineId }) {
  * Checks if all builds in stage are done running
  * @param  {Object}     stage                     Stage
  * @param  {Object}     event                     Event
+ * @param  {String}     buildStatus               Current build status
  * @return {Boolean}              Flag if stage is done
  */
-async function isStageDone({ stage, event }) {
-    // Get all jobIds for jobs in the stage
-    const stageJobIds = stage.jobIds;
-
-    stageJobIds.push(stage.setup);
-
-    // Get all builds in a stage for this event
+async function isStageDone({ stage, event, buildStatus }) {
+    // Combine job IDs in the stage and the stage setup ID
+    const stageJobIds = [...stage.jobIds, stage.setup];
+    // Get all builds for jobs in the stage for this event
     const stageJobBuilds = await event.getBuilds({ params: { jobId: stageJobIds } });
-    let stageIsDone = false;
 
-    // Make sure all builds in stage have run
-    if (stageJobBuilds && stageJobBuilds.length === stageJobIds.length) {
-        stageIsDone = !stageJobBuilds.some(b => !FINISHED_STATUSES.includes(b.status));
+    // If no builds exist, return false
+    if (!stageJobBuilds || stageJobBuilds.length === 0) {
+        return false;
+    }
+
+    // Check if all builds in the stage are finished
+    const stageIsDone = !stageJobBuilds.some(b => !FINISHED_STATUSES.includes(b.status));
+
+    // If the build status is 'SUCCESS', ensure all jobs are completed
+    if (buildStatus === 'SUCCESS') {
+        return stageJobBuilds.length === stageJobIds.length && stageIsDone;
     }
 
     return stageIsDone;
@@ -301,6 +306,7 @@ module.exports = () => ({
                 jobName: job.name,
                 pipelineId: pipeline.id
             });
+
             const isStageTeardown = STAGE_TEARDOWN_PATTERN.test(job.name);
             let stageBuildHasFailure = false;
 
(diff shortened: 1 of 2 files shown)


VonnyJap (dismissed):

plugins/builds/update.js
```diff
@@ -197,7 +194,8 @@ async function isStageDone({ stage, event }) {
     const stageJobBuilds = await event.getBuilds({ params: { jobId: stageJobIds } });
     let stageIsDone = false;
 
-    if (stageJobBuilds && stageJobBuilds.length !== 0) {
+    // Make sure all builds in stage have run
+    if (stageJobBuilds && stageJobBuilds.length === stageJobIds.length) {
```
sagar1312, line 198: Number of stage job builds may not always equal to the number of jobs within the stage.
In case of a failure of a stage job, downstream stage jobs will not be executed. In such case, with the changes made teardown may never get executed

sagar1312 (changes requested):
We may have to revert this change

plugins/builds/update.js
tkyi: updated the logic

VonnyJap (dismissed):

plugins/builds/update.js
```diff
@@ -359,7 +361,7 @@ module.exports = () => ({
 
                 // Start stage teardown build if stage is done
                 if (stageTeardownBuild && stageTeardownBuild.status === 'CREATED') {
-                    const stageIsDone = await isStageDone({ stage, event: newEvent });
+                    const stageIsDone = await isStageDone({ stage, event: newEvent, buildStatus: newBuild.status });
```
sagar1312, line 364: Should we use stage build status instead of the job build status?
```suggestion
                    const stageIsDone = await isStageDone({ stage, event: newEvent, stageBuildStatus: stageBuild.status });
```

plugins/builds/update.js
```diff
@@ -197,7 +195,10 @@ async function isStageDone({ stage, event }) {
     const stageJobBuilds = await event.getBuilds({ params: { jobId: stageJobIds } });
     let stageIsDone = false;
 
-    if (stageJobBuilds && stageJobBuilds.length !== 0) {
+    // Make sure all builds in stage have run successfully in success case
+    if (stageJobBuilds && buildStatus === 'SUCCESS') {
+        stageIsDone = stageJobBuilds.length === stageJobIds.length;
```
sagar1312, line 200: We should not conclude the stage as completed when the `stageJobBuilds.length` match `stageJobIds.length`. We also need to ensure that all those builds have reached terminal status.

sagar1312: I think below logic to update the status of stage build is buggy
https://github.com/screwdriver-cd/screwdriver/pull/3216/files#diff-2d039e9e4eec4a54c7c110debffb3a176c9774982fc9f9d783a5a750defaa2d2R317-R322

When the stage job build is successful, we might be prematurely marking the stage build as `SUCCESS` without checking if other stage job builds are pending execution.

sagar1312 (changes requested):

plugins/builds/update.js
```diff
@@ -185,20 +182,26 @@ async function getStage({ stageFactory, workflowGraph, jobName, pipelineId }) {
...
-    const stageJobBuilds = await event.getBuilds({ params: { jobId: stageJobIds } });
-    let stageIsDone = false;
+    // Check if all builds in the stage are finished
+    const stageIsDone = !stageJobBuilds.some(b => !FINISHED_STATUSES.includes(b.status));
 
-    if (stageJobBuilds && stageJobBuilds.length !== 0) {
-        stageIsDone = !stageJobBuilds.some(b => !FINISHED_STATUSES.includes(b.status));
+    // If the build status is 'SUCCESS', ensure all jobs are completed
+    if (buildStatus === 'SUCCESS') {
+        return stageJobBuilds.length === stageJobIds.length && stageIsDone;
```
sagar1312, line 204: I believe we should not be relying on the count of stage job builds.

For Ex: Let us consider a stage which has 5 jobs `jobA`, `jobB`, `jobC1`, `jobC2` and `jobD`. `jobA`, `jobB`, `jobC1`, `jobD` are triggered in parallel after stage setup is completed. `jobC2` is downstream of `jobC1`. Let us assume `jobA`, `jobB` and `jobC1` have completed with statuses shown below and the current build is for `jobD` with status `SUCCESS`.
| Job | Build Status |
| --- | --- |
| setup | SUCCESS |
| jobA | SUCCESS |
| jobB | SUCCESS |
| jobC1 | FAILURE |
| jobD | SUCCESS |
There will no active builds (There is be no build for `jobC2`). Per this logic, `stageJobBuilds.length` (5) would not match `stageJobIds.length` (6). Stage will never be marked as done and `teardown` may not get triggered.

tkyi:
Closing in favor of https://github.com/screwdriver-cd/workflow-parser/pull/56

Issue #3227, opened 2024-10-18
[Feature Request] Feature flag for Download Zipped Artifacts


kumada626:
**What happened**:
In current implementation, screwdriver zips artifacts on its API and no restriction on its download file size. Because of this, if users try to download a large file size artifacts might crushing API.
So we want to disable this feature until size restriction is implemented.

**What you expected to happen**:
If feature flag for Download Zipped Artifacts is set false on [UI config](https://github.com/screwdriver-cd/ui/blob/master/config/environment.js#L47), users only can download each artifacts and cannot download zipped directory or all artifacts(same as before).

**How to reproduce it**:

VonnyJap:
@kumada626 - We can close this issue now?

kumada626:
@VonnyJap Yes, I think we can close this issue. 
However, we have not implemented foolproof solution that is described in [design doc](https://github.com/screwdriver-cd/screwdriver/blob/master/design/sd-download-artifacts.md).
So it is good to create another issue to improve download artifacts feature.

kumada626:
@VonnyJap Yes, we can close this issue.

PR #3232, opened 2024-10-29
fix: Add maxDownloadSize for artifact directory download

tkyi:
## Context

Would be nice if we could check for max file download size before downloading.

## Objective

This PR adds config for `maxDownloadSize` for downloading artifact directories.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/3227

## License

### DIFF
--- a/plugins/builds/artifacts/get.js
+++ b/plugins/builds/artifacts/get.js
@@ -31,6 +31,7 @@ module.exports = config => ({
             const { credentials } = req.auth;
             const { canAccessPipeline } = req.server.plugins.pipelines;
             const { buildFactory, eventFactory } = req.server.app;
+            const { maxDownloadSize } = buildFactory;
 
             return buildFactory.get(buildId)
                 .then(build => {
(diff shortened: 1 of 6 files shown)


plugins/builds/artifacts/get.js
```diff
@@ -70,6 +71,23 @@ module.exports = config => ({
...
+                                if (file) {
+                                    const fileMetaResponse = await request.head(`${baseUrl}/${file}?token=${token}&type=download`);
+                                    const fileSize = parseInt(fileMetaResponse.headers['content-length'], 10);
+
+                                    // Accumulate total size
+                                    totalSize += fileSize;
+
+                                    // If total size exceeds allowed limit, stop further processing
+                                    if (totalSize > maxDownloadSize) {
+                                        throw new Error('Total size of files exceeds the allowed limit of 2GB.');
```
kumada626, line 87: Limit size in error message should not be hardcoded because limit download size is configurable.

plugins/builds/artifacts/get.js
```diff
@@ -70,6 +71,23 @@ module.exports = config => ({
                             }).text();
                             const manifestArray = manifest.trim().split('\n');
                             const directoryArray = manifestArray.filter(f => f.startsWith(`./${artifact}/`));
+                            let totalSize = 0;
+
+                            // Check file sizes by fetching metadata
+                            for (const file of directoryArray) {
+                                if (file) {
+                                    const fileMetaResponse = await request.head(`${baseUrl}/${file}?token=${token}&type=download`);
```
pritamstyz4ever, line 79: not pressing: for large no of files fetching metadata can be time consuming, we can run in parallel with a small batch size.

pritamstyz4ever: also, `req.query.dir && req.query.type === 'download'` , we may need schema validation for query params

pritamstyz4ever (dismissed):
minor comments, nothing urgent can be incorporated later.

plugins/builds/artifacts/get.js
tkyi: updated

plugins/builds/artifacts/get.js
```diff
@@ -70,6 +71,23 @@ module.exports = config => ({
...
+                                    const fileSize = parseInt(fileMetaResponse.headers['content-length'], 10);
+
+                                    // Accumulate total size
+                                    totalSize += fileSize;
+
+                                    // If total size exceeds allowed limit, stop further processing
+                                    if (totalSize > maxDownloadSize) {
+                                        throw new Error(`Total size of files exceeds the allowed limit of ${maxDownloadSize}GB.`);
```
kumada626, line 87: Isn't maxDownloadSize in bytes, not GB? I think it is needed to change the logging unit or convert it when outputting to the log.

kumada626 (approved):

Issue #3181, opened 2024-08-30
screwdriver.cd/virtualJob seems to run the full job when triggered by hand


scr-oath:
**What happened**:

Trigger via "restart from here" a job that is annotated with `screwdriver.cd/virtualJob: true` and watch it go through all steps such as checkout, cache pulling, etc.

**What you expected to happen**:

`screwdriver.cd/virtualJob: true` should do nothing no matter how it's triggered

**How to reproduce it**:

- Create a job annotated with `screwdriver.cd/virtualJob: true` and trigger it by hand
- Watch every step run.

kumada626:
I also found a case where a job annotated with `screwdriver.cd/virtualJob: true` executes all steps. If a job annotated with `screwdriver.cd/virtualJob: true` has `~commit` or `~pr` in its `requires` property and is triggered by scm webhook, the job will run every step.

VonnyJap:
As of now, this has been missed out for the implementation of virtual job and we are prioritizing this as the next improvement to release.

PR #3252, opened 2024-12-12
feat(3181): Skip execution of a virtual job when an event is started from that job

sagar1312:
## Context

Per changes made in https://github.com/screwdriver-cd/models/pull/635, virtual job builds at the beginning in the workflow for an event, may not be queued for execution. Status of such build will be set to `CREATED`.

## Objective

Identify such build immediately after creating the event to mark its status as 'SUCCESS' and trigger its downstream jobs.

## References

https://github.com/screwdriver-cd/screwdriver/issues/3181

## License

### DIFF
--- a/plugins/builds/helper/updateBuild.js
+++ b/plugins/builds/helper/updateBuild.js
@@ -11,14 +11,15 @@ const FINISHED_STATUSES = ['FAILURE', 'SUCCESS', 'ABORTED', 'UNSTABLE', 'COLLAPS
 /**
  * @typedef {import('screwdriver-models/lib/build')} Build
  * @typedef {import('screwdriver-models/lib/event')} Event
+ * @typedef {import('screwdriver-models/lib/step')} Step
  */
 
 /**
  * Identify whether this build resulted in a previously failed job to become successful.
  *
  * @method isFixedBuild
- * @param  build         Build Object
- * @param  jobFactory    Job Factory instance
+ * @param  {Build}          build       Build Object
+ * @param  {JobFactory}     jobFactory  Job Factory instance
  */
 async function isFixedBuild(build, jobFactory) {
     if (build.status !== 'SUCCESS') {
@@ -29,18 +30,15 @@ async function isFixedBuild(build, jobFactory) {
     const failureBuild = await job.getLatestBuild({ status: 'FAILURE' });
     const successBuild = await job.getLatestBuild({ status: 'SUCCESS' });
 
-    if ((failureBuild && !successBuild) || failureBuild.id > successBuild.id) {
-        return true;
-    }
-
-    return false;
+    return !!((failureBuild && !successBuild) || failureBuild.id > successBuild.id);
 }
 
 /**
  * Stops a frozen build from executing
+ *
  * @method stopFrozenBuild
- * @param  {Object} build         Build Object
- * @param  {String} previousStatus    Prevous build status
+ * @param  {Build}  build           Build Object
+ * @param  {String} previousStatus  Previous build status
  */
 async function stopFrozenBuild(build, previousStatus) {
     if (previousStatus !== 'FROZEN') {
@@ -52,9 +50,11 @@ async function stopFrozenBuild(build, previousStatus) {
 
 /**
  * Updates execution details for init step
- * @method stopFrozenBuild
- * @param  {Object} build       Build Object
- * @param  {Object} app         Hapi app Object
+ *
+ * @method  stopFrozenBuild
+ * @param   {Build}         build   Build Object
+ * @param   {Object}        app     Hapi app Object
+ * @returns {Promise<Step>}         Updated step
  */
 async function updateInitStep(build, app) {
     const step = await app.stepFactory.get({ buildId: build.id, name: 'sd-setup-init' });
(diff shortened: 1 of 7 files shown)


plugins/events/create.js
```diff
@@ -244,6 +246,23 @@ module.exports = () => ({
             if (event.builds === null) {
                 return boom.notFound('No jobs to start.');
             }
+
+            const virtualJobBuilds = event.builds.filter(b => b.status === 'CREATED');
```
VonnyJap, line 250: is that enough to assume that all builds with `status === 'CREATED'` is a virtual job then?

plugins/builds/helper/updateBuild.js
```diff
@@ -0,0 +1,304 @@
...
+ * @method stopFrozenBuild
+ * @param  {Object} build         Build Object
+ * @param  {String} previousStatus    Prevous build status
+ */
+async function stopFrozenBuild(build, previousStatus) {
+    if (previousStatus !== 'FROZEN') {
+        return Promise.resolve();
+    }
+
+    return build.stopFrozen(previousStatus);
```
VonnyJap, line 50:
```suggestion
    return await build.stopFrozen(previousStatus);
```

plugins/builds/helper/updateBuild.js
```diff
@@ -0,0 +1,304 @@
...
+}
+
+/**
+ * Updates execution details for init step
+ * @method stopFrozenBuild
+ * @param  {Object} build       Build Object
+ * @param  {Object} app         Hapi app Object
+ */
```
VonnyJap, line 58:
```suggestion
/**
 * Updates execution details for init step
 * @method updateInitStep
 * @param  {Object} build       Build Object
 * @param  {Object} app         Hapi app Object
 * @return {Promise}            Promise resolving when the init step is updated
 */
 ```

File: plugins/builds/helper/updateBuild.js [hunk 4]
```diff
@@ -0,0 +1,304 @@
...
+    // If there is no init step, do nothing
+
+    if (!step) {
+        return null;
+    }
+
+    step.endTime = build.startTime || new Date().toISOString();
+    step.code = 0;
+
+    return step.update();
```
VonnyJap, line 70:
```suggestion
    return await step.update();
```

plugins/builds/helper/updateBuild.js
```diff
@@ -0,0 +1,304 @@
...
+        } else if (build.status === 'FAILURE' || build.status === 'SUCCESS') {
+            if (statusMessage) {
+                build.statusMessage = statusMessage;
+                build.statusMessageType = statusMessageType || null;
+            }
+        } else {
+            build.statusMessage = statusMessage || null;
+            build.statusMessageType = statusMessageType || null;
+        }
+    }
```
VonnyJap, line 105:
```suggestion
    // UNSTABLE -> SUCCESS needs to update meta and endtime.
    // However, the status itself cannot be updated to SUCCESS
    if (currentStatus === 'UNSTABLE') {
        return;
    }

    // Update the build status if desiredStatus is provided
    if (desiredStatus !== undefined) {
        build.status = desiredStatus;
    }

    // Handle different statuses and set the appropriate status message
    switch (build.status) {
        case 'ABORTED':
            build.statusMessage = currentStatus === 'FROZEN'
                ? `Frozen build aborted by ${username}`
                : `Aborted by ${username}`;
            break;
        case 'FAILURE':
        case 'SUCCESS':
            if (statusMessage) {
                build.statusMessage = statusMessage;
                build.statusMessageType = statusMessageType || null;
            }
            break;
        default:
            build.statusMessage = statusMessage || null;
            build.statusMessageType = statusMessageType || null;
            break;
    }
```
to improve readability: 
- use early return
- use switch instead of nested if else.

plugins/builds/helper/updateBuild.js
```diff
@@ -0,0 +1,304 @@
...
+ * Checks if all builds in stage are done running
+ * @param  {Object}     stage                     Stage
+ * @param  {Object}     event                     Event
+ * @return {Boolean}              Flag if stage is done
+ */
+async function isStageDone({ stage, event }) {
+    // Get all jobIds for jobs in the stage
+    const stageJobIds = stage.jobIds;
+
+    stageJobIds.push(stage.setup);
```
VonnyJap, line 140:
```suggestion
    const stageJobIds = [...stage.jobIds, stage.setup];
```

plugins/events/create.js
sagar1312: This could be plural, when we have an event started from `~pr` or `~commit`, and there are more than one virtual job listening to these triggers.

sagar1312: yes, builds created immediately after the event is created would be in `CREATED` status only if the job is virtual.

plugins/builds/helper/updateBuild.js
sagar1312: This is an existing code. Just moved it to a common module. Avoid any additional refactoring as part of this PR>

Also `await` on individual promise is not needed here.
It is being done here https://github.com/screwdriver-cd/screwdriver/pull/3252/files/d12977ba7d6469c2ca79437eb26f62eb580b3c5b#diff-2c3decb1d714adeb8f265dd68e51a120caa8ada38d91b21d9dffc66974f3276dR212

plugins/builds/helper/updateBuild.js
sagar1312: Same here we don't to `await` here. It is done by the caller 
https://github.com/screwdriver-cd/screwdriver/pull/3252/files/d12977ba7d6469c2ca79437eb26f62eb580b3c5b#diff-2c3decb1d714adeb8f265dd68e51a120caa8ada38d91b21d9dffc66974f3276dR206

plugins/builds/helper/updateBuild.js
```diff
@@ -0,0 +1,304 @@
...
+        build.endTime = new Date().toISOString();
+    } else if (desiredStatus === 'RUNNING') {
+        build.startTime = new Date().toISOString();
+    } else if (desiredStatus === 'BLOCKED' && !hoek.reach(build, 'stats.blockedStartTime')) {
+        build.stats = Object.assign(build.stats, {
+            blockedStartTime: new Date().toISOString()
+        });
+    } else if (desiredStatus === 'QUEUED' && currentStatus !== 'QUEUED') {
+        throw boom.badRequest(`Cannot update builds to ${desiredStatus}`);
+    } else if (desiredStatus === 'BLOCKED' && currentStatus === 'BLOCKED') {
```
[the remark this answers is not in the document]
sagar1312: No change needed. This is an exiting logic. Just moved it to a common module.
Per the comment "Queue-Service can call BLOCKED status update multiple times", both `desiredStatus` and `currentStatus` can be `BLOCKED`.

plugins/builds/helper/updateBuild.js
sagar1312: Done

plugins/builds/helper/updateBuild.js
sagar1312: Done

plugins/builds/helper/updateBuild.js
sagar1312: Done

test/plugins/events.test.js
```diff
@@ -377,8 +400,12 @@ describe('event plugin test', () => {
                 meta
             };
 
-            eventFactoryMock.get.withArgs(parentEventId).resolves(getEventMock(testEvent));
-            eventFactoryMock.create.resolves(getEventMock(testEvent));
+            eventMock = getEventMock(testEvent);
+            // eventFactoryMock.get.withArgs(parentEventId).resolves(getEventMock(testEvent));
```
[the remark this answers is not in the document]
sagar1312: Done

sagar1312:
@minghay @VonnyJap I have addressed the suggestions. Please take a look.

minghay (approved):

plugins/events/create.js
VonnyJap: What about builds that are waiting for other joined parent downstream builds to complete? Aren't those builds normally created and then park with "CREATED" status?

sagar1312: There are the builds at the beginning of the workflow for an event. I believe, these wouldn't be waiting for any upstream.

Issue #3225, opened 2024-10-17
Webhook returns 204 even when a parsing error occurs


kumada626:
**What happened**:
When a parsing error occurs during the webhook parsing process, the response still returns a 204 status, making it difficult to notice that an error has occurred.

**What you expected to happen**:
The system should correctly return a 500 or 400-series error, making it easier to notice when an error occurs.

**How to reproduce it**:
1. Send a webhook that will cause a parsing error.
2. Observe that the response returns a 204 status instead of an error status.

PR #3256, opened 2024-12-17
fix(3235): Fetch parentEvent meta when create Event

kumada626:
## Context

When restart build was failed in `sd-setup-init` step and restart build from that build, meta is missing at that time.
So we fix this behavior by fetching parent event meta when creating event.

## Objective

Fetch parent event meta when creating event.

## References

#3225 

## License

### DIFF
--- a/plugins/events/create.js
+++ b/plugins/events/create.js
@@ -220,15 +220,24 @@ module.exports = () => ({
 
             payload.sha = sha;
 
-            // If there is parentEvent, pass workflowGraph and sha to payload
+            // If there is parentEvent, pass workflowGraph, meta and sha to payload
             // Skip PR, for PR builds, we should always start from latest commit
             if (payload.parentEventId) {
                 const parentEvent = await eventFactory.get(parentEventId);
+                let mergedParameters = {};
 
                 payload.baseBranch = parentEvent.baseBranch || null;
 
                 if (!payload.meta.parameters && parentEvent.meta && parentEvent.meta.parameters) {
-                    payload.meta.parameters = parentEvent.meta.parameters;
+                    mergedParameters = parentEvent.meta.parameters;
+                }
+
+                if (Object.keys(payload.meta).length === 0) {
+                    payload.meta = { ...parentEvent.meta };
+                }
+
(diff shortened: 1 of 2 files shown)


y-oksaku (approved):

yakanechi (approved):

plugins/events/create.js
```diff
@@ -220,15 +220,24 @@ module.exports = () => ({
...
+                    mergedParameters = parentEvent.meta.parameters;
+                }
+
+                if (Object.keys(payload.meta).length === 0) {
+                    payload.meta = { ...parentEvent.meta };
+                }
+
+                if (Object.keys(mergedParameters).length > 0) {
+                    payload.meta.parameters = mergedParameters;
                 }
```
VonnyJap, line 241: what about make this more simplified?
```suggestion
                // Merge parameters if they exist in the parent event and not in the payload
                if (!payload.meta.parameters && parentEvent.meta && parentEvent.meta.parameters) {
                    payload.meta.parameters = parentEvent.meta.parameters;
                }
                
                // Copy meta from parent event if payload.meta is empty
                if (Object.keys(payload.meta).length === 0) {
                    payload.meta = { ...parentEvent.meta };
                }
```

VonnyJap (changes requested):
please check my suggestion

plugins/events/create.js
kumada626: If payload(RestartedEevnt) only has parameters in meta, I think your suggested method will not fetch parentEvent meta (and neither will my method).
I will have fixed this bug.

VonnyJap:
@kumada626 - can you also add couple of functional tests 
https://github.com/screwdriver-cd/screwdriver/blob/master/features/metadata.feature
Changes to metadata behavior can be crucial and we should add some functional tests to make sure that it is working as intended if possible.

kumada626:
@VonnyJap Since this change does not alter normal behavior, it would be difficult to add it as a functional test.
This change is to fix the case of a restart from a build that failed in setup-init of a build due to some temporary environmental factors, so if we can guarantee that the normal behavior has not changed, there should be no problem.

VonnyJap (approved):

PR #3257, opened 2024-12-17
fix: Show error details and change status code in case of webhook error [2]

yakanechi:
## Context

If a user specifies a non-existent buildcluster for a job, models will return an error, but the API will return 500 because no status code is specified.

Also, even though models returns a detailed error, it does not output the error to the webhook, so the user cannot determine the cause of the error by looking at the webhook response.

## Objective

Outputs a detailed error log to the webhook response
In addition, the status code set in the event of an error is specified in the webhook response.

## References

Related: https://github.com/screwdriver-cd/models/pull/636

## License

### DIFF
--- a/plugins/webhooks/helper.js
+++ b/plugins/webhooks/helper.js
@@ -447,23 +447,21 @@ async function startEvents(eventConfigs, eventFactory) {
 
     if (errorCount > 0 && errorCount === eventsCount) {
         // preserve current behavior of returning 500 on error
-        const errorMessages = [];
+        const errorMessages = new Set();
         let statusCode;
 
         errors.forEach(err => {
-            if (err.message !== undefined) {
-                errorMessages.push(err.message);
-            }
+            errorMessages.add(`"${err.message}"`);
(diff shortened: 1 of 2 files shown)


plugins/webhooks/helper.js
```diff
@@ -433,18 +433,38 @@ async function startEvents(eventConfigs, eventFactory) {
...
+    if (errorCount > 0 && errorCount === eventsCount) {
         // preserve current behavior of returning 500 on error
-        throw new Error('Failed to start any events');
+        const errorMessages = new Set();
+        let statusCode;
+
+        errors.forEach(err => {
+            errorMessages.add(`"${err.message}"`);
+            if (err.statusCode !== undefined) {
+                statusCode = err.statusCode;
```
y-oksaku, line 456: I think that returning 500 is better if the error has no statusCode.

yakanechi: Fixed default value to 500

y-oksaku (approved):

VonnyJap (approved):

Issue #3266, opened 2025-01-15
Extend banner schema to have scope level information


VonnyJap:
**What happened**:

The [current banner implementation](https://github.com/screwdriver-cd/screwdriver/tree/master/plugins/banners) operates at the global level, but sometimes pipeline or build specific notifications are required to inform only affected pipelines or builds based on actions taken by the cluster admins or other automations.

**What you expected to happen**:
The proposed improvements include extending [the banner schema](https://github.com/screwdriver-cd/data-schema/blob/master/models/banner.js) to integrate scope level metadata and a reference to pipeline or build IDs, depending on the scope. 

**How to reproduce it**:
New feature, no way to reproduce.

PR #3278, opened 2025-02-05
feat(3266): add authn for banner list and get endpoints

VonnyJap:
## Context

The current banner endpoints for [list](https://github.com/screwdriver-cd/screwdriver/blob/87dfd38e62a2357b230b484a3f2b6fb971986743/plugins/banners/list.js) and [get](https://github.com/screwdriver-cd/screwdriver/blob/87dfd38e62a2357b230b484a3f2b6fb971986743/plugins/banners/get.js) do not require user authentication, which is unintended behavior.

## Objective

The change aims to enforce token-based authentication for both API endpoints.

## References

https://github.com/screwdriver-cd/screwdriver/issues/3266

## License


plugins/banners/get.js
```diff
@@ -13,6 +13,10 @@ module.exports = () => ({
         description: 'Get a single banner',
         notes: 'Return a banner record',
         tags: ['api', 'banners'],
+        auth: {
+            strategies: ['token'],
+            scope: ['user', 'build', 'pipeline']
```
sagar1312, line 18:
```suggestion
            scope: ['user', '!guest']
```

plugins/banners/list.js
```diff
@@ -10,6 +10,10 @@ module.exports = () => ({
         description: 'Get banners',
         notes: 'Returns all banner records',
         tags: ['api', 'banners'],
+        auth: {
+            strategies: ['token'],
+            scope: ['user', 'build', 'pipeline']
```
sagar1312, line 15:
```suggestion
            scope: ['user', '!guest']
```

sagar1312 (changes requested):

plugins/banners/get.js
```diff
@@ -13,6 +13,10 @@ module.exports = () => ({
         description: 'Get a single banner',
         notes: 'Return a banner record',
         tags: ['api', 'banners'],
+        auth: {
+            strategies: ['token'],
+            scope: ['user']
```
sagar1312, line 18: Banners with `GLOBAL` scope are currently being fetched without the user being authenticated.
image


This change makes it backward incompatible. We are unable fetch 'GLOBAL' banners when the user is not authenticated.
image

Auth check may need to be done selectively based on the scope. `GLOBAL` can be accessed by anyone like before. But, `PIPELINE` should be accessed by a user with valid token.

sagar1312 (dismissed):

minghay (approved):

PR #3303, opened 2025-03-11
[Revert] commit #f50c0429

pritamstyz4ever:
## Context

Reverting change to use admin in hapi auth scope instead using user scope and additional admin check.
Currently, token generation using api key does not yield a token with scope admin, hence we need to relax the route to allow user scope and add the admin check logic

## Objective

This PR reverts the change to use admin scope in auth for update and remove job build cluster apis

## References

https://github.com/screwdriver-cd/screwdriver/pull/3295/commits/f50c042983f483e1d838c5b2242dd7d8ce3c7708

## License

### DIFF
--- a/plugins/jobs/buildCluster/remove.js
+++ b/plugins/jobs/buildCluster/remove.js
@@ -15,12 +15,12 @@ module.exports = () => ({
         tags: ['api', 'jobs'],
         auth: {
             strategies: ['token'],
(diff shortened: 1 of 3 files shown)


minghay:
The scopes require a revisit and clean up.  For now, this revert makes the most sense.

minghay (approved):

Issue #3304, opened 2025-03-14
Relax pipeline update to allow admins from different SCM contexts


VonnyJap:
**What happened**:

Screwdriver restricts pipeline updates to admins within the same SCM context as the user’s login.

**What you expected to happen**:

Admins from different SCM contexts should be able to update the pipeline. This would allow users to retain pipeline history when migrating between SCMs.

**How to reproduce it**:

1. Navigate to any pipeline.
2. Attempt to update the checkout URL to a different SCM.

<img width="1697" alt="Image" src="https://github.com/user-attachments/assets/0d5e75e7-6b80-448e-8f5b-e7ec1b0063a7" />

PR #3305, opened 2025-03-14
feat(3304): Relax permission check to allow admins from other SCMs to update pipeline

VonnyJap:
## Context

Screwdriver restricts pipeline updates to admins within the same SCM context as the user’s login.

## Objective

This PR relaxes the permission check, allowing admins from different SCM contexts to update the pipeline.

## References

https://github.com/screwdriver-cd/screwdriver/issues/3304

## License

### DIFF
--- a/plugins/pipelines/update.js
+++ b/plugins/pipelines/update.js
@@ -26,9 +26,10 @@ function getPermissionsForOldPipeline({ scmContexts, pipeline, user }) {
         if (pipeline.admins[user.username]) {
             permission.admin = true;
         }
+
         return Promise.resolve(permission);
     }
-    
+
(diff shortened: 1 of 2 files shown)


tanoda:
Generally, the user with the same username in the different SCMs is just one with the same username but a different user.
We should add the configuration of which SCM contexts have the same username restriction.

VonnyJap:
@tanoda - a possibility of having duplicate usernames from different SCMs is possible as you noted. With this we will implement a check as such that the admin check relies on Screwdriver user ID from the user table. But to maintain backward compatibility, we need to still fallback to check username when the user ID is not populated as admin in the pipeline table.

sagar1312 (approved):

Issue #3287, opened 2025-02-21
Metadata cannot be preferentially inherited from virtual jobs


yakanechi:
**What happened**:
related #3235

In the following workflow, `target` cannot get the metadata of a in priority.
Therefore, `target` may take over `b`'s metadata instead of `a`'s metadata.

```
shared:
  image: node:22

jobs: 
  a:
    requires: [ ~commit ]
    steps:
      - meta set x "a"

  b:
    requires: [ ~commit ]
    steps:
      - meta set x "b"

  virtual:
    requires: [ ~a ]
    steps:
      - meta set x "a"
    annotations:
      screwdriver.cd/virtualJob: true

  target:
    requires: [ ~virtual ]
    blockedBy: [ ~b ]
    steps:
      - check_meta: test "a" = `meta get "x"`
```

**What you expected to happen**:
The metadata of `a` can be preferentially inherited.

**How to reproduce it**:

yakanechi:
Basically, the following PRs will fix the problem.
https://github.com/screwdriver-cd/screwdriver/pull/3288

However, if a Virtual Job is triggered "AND" triggered by multiple jobs, and if those jobs complete at about the same time, the bug may occur.

For example, if `a` and `b` are completed almost simultaneously in the following workflow
1. After the completion of `a`, the trigger process of virtual is performed and the meta of `a` is transferred to `virtual`.
2. At this time, if the status of `b` is SUCCESS, virtual is executed without waiting for the trigger process from `b`.
3.`target` is triggered immediately from `virtual`
4. After that, `b`'s trigger process is performed and `virtual` inherits b's meta, but `target` cannot inherit b's meta because `target`'s build has already been started.
```
shared:
    image: node:22
    steps:
        - echo: echo "test"

jobs:
    hub:
        requires: [ ~commit, ~pr ]
        annotations:
            screwdriver.cd/virtualJob: true

    a:
        requires: [ ~hub ]
        steps:
            - set: meta set foo.bar "success"

    b:
        requires: [ ~hub ]
        steps:
            - set: meta set foo.baz "success"

    c:
        requires: [ ~hub ]
        steps:
            - set: meta set foo.bar "error"

    d:
        requires: [ ~hub ]
        steps:
            - set: meta set foo.baz "error"

    virtual:
        requires: [ a, b ]
        annotations:
            screwdriver.cd/virtualJob: true

    target:
        requires: [ ~virtual ]
        blockedBy: [ ~c, ~d ]
[...]

PR #3307, opened 2025-03-17
fix(3287): Set parentBuildId just before execution

y-oksaku:
## Context

Sometime, `parentBuildId` is not correct when parent builds are finished in same time.
For this reason, the build miss some metadata from the parent builds in launcher.

Likewise, virtual job cannot get some metadata from the parent builds.
https://github.com/screwdriver-cd/screwdriver/issues/3287#issuecomment-2674046847

## Objective

- Update `parentBuildId` just before build execution
- Merge `metadata` from all of the parent builds into the virtual job just before build execution

## References

Issue - https://github.com/screwdriver-cd/screwdriver/issues/3287#issuecomment-2674046847

## License

### DIFF
--- a/plugins/builds/triggers/orBase.js
+++ b/plugins/builds/triggers/orBase.js
@@ -61,7 +61,10 @@ class OrBase {
                 nextBuild.status = Status.SUCCESS;
                 nextBuild.statusMessage = BUILD_STATUS_MESSAGES.SKIP_VIRTUAL_JOB.statusMessage;
                 nextBuild.statusMessageType = BUILD_STATUS_MESSAGES.SKIP_VIRTUAL_JOB.statusMessageType;
-                nextBuild.meta = merge({}, nextBuild.meta, this.currentBuild.meta);
+
+                // Overwrite metadata by current build's
+                nextBuild.parentBuildId = [this.currentBuild.id];
+                nextBuild.meta = merge({}, this.currentBuild.meta);
 
                 return nextBuild.update();
             }
@@ -83,7 +86,7 @@ class OrBase {
             event,
             baseBranch: event.baseBranch || null,
             parentBuilds,
-            parentBuildId: this.currentBuild.id,
+            parentBuildId: [this.currentBuild.id],
             start: hasWindows || !isNextJobVirtual,
             causeMessage
         });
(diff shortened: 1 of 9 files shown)


minghay (commented):
This PR changes `parentBuildId` to be an array contrary to the model specification: https://github.com/screwdriver-cd/models/blob/master/lib/buildFactory.js#L171

Is there a reason for changing to an array?

y-oksaku:
Originally, `parentBuildId` was also an array.  I modified it to be consistent.

https://github.com/screwdriver-cd/screwdriver/blob/22967afca4701ea57064fcb5656b7c0eb65cde14/plugins/builds/triggers/helpers.js#L529-L534

https://github.com/screwdriver-cd/screwdriver/blob/22967afca4701ea57064fcb5656b7c0eb65cde14/plugins/builds/triggers/helpers.js#L560

However, this change is not required, so I will revert it in this PR if there are any issues.

yakanechi (approved):

minghay:
> Originally, `parentBuildId` was also an array. I modified it to be consistent.
> 
> https://github.com/screwdriver-cd/screwdriver/blob/22967afca4701ea57064fcb5656b7c0eb65cde14/plugins/builds/triggers/helpers.js#L529-L534
> 
> https://github.com/screwdriver-cd/screwdriver/blob/22967afca4701ea57064fcb5656b7c0eb65cde14/plugins/builds/triggers/helpers.js#L560
> 
> However, this change is not required, so I will revert it in this PR if there are any issues.

I think there shouldn't be any adverse effects with the `parentBuildId` being forced to an array with this PR change.  I do think that we should adhere to the model specification better.

minghay (dismissed):

y-oksaku:
Thank you for your comment. I reverted the arguments of `buildFactory.create`.

plugins/builds/triggers/orBase.js
```diff
@@ -56,12 +56,16 @@ class OrBase {
                 return nextBuild;
             }
 
+            nextBuild.parentBuildId = [this.currentBuild.id];
```
minghay, line 59:
```suggestion
            nextBuild.parentBuildId = this.currentBuild.id;
```
Should this be a singular item?

y-oksaku: No. Definition of `parentBuildId` is array.

https://github.com/screwdriver-cd/data-schema/blob/714afa62b464326b28a9fe3bc82cffba14e54b41/models/build.js#L36

minghay (approved):

minghay (approved):

PR #3315, opened 2025-04-02
Proposal for revamping the workflow graph

minghay:
## Context
The pipeline workflow graph is in serious need of a ground up rebuild.  Before the actual work happens, here's a high level proposal for how to rework the graph implementation.

## Objective
Lays out the high level plan for how to replace the existing workflow graph implementation.  The plan calls for using more third-party libraries to replace all of the custom code that exists for the graph rendering.

## License

### DIFF
--- a/design/workflow-graph.md
+++ b/design/workflow-graph.md
@@ -1,0 +1,42 @@
+# Workflow Graph Revamp
+
+## Context
+The current workflow graph does not leverage any graphing libraries and everything is hand-rolled.  The current workflow graph implementation is now very difficult to work with, prone to various errors, and lacks flexibility.  A new solution needs to be implemented to split the workflow graph into distinct areas of concern: 
+- core graph data that is compatible with various graphing libraries for leveraging graph algorithms
+- graph layout composition that can leverage different graph layout libraries and algorithms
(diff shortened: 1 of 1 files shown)


design/workflow-graph.md
```diff
@@ -0,0 +1,42 @@
...
+Rendering the graph and supporting various user interactions should be offloaded to a third-party library.  Currently, D3 is used and does have support for these features; however, we may be able to streamline the rendering layer by using another library for the near-term.
+
+With the above points in mind, the current proposal for updating the workflow graph is as follows:
+- graph layout library options
+  - ELK
+  - dagre
+- graph rendering
+  - cytoscape
+    - extension for the elk layout
+    - extension for the dagre layout -- the plugin is not actively developed and a few years out of date.  Using dagre for the layout algorithm will be best served by porting the old plugin or coming up with a new shim layer
```
kumada626, line 30: There is a little concern about the fact that the library has not been actively developed for several years. In using the library, we would like to be concerned about whether it is properly maintained.

kumada626:
I have a small comment, but I don't see any major issues with the proposal.

PR #3317, opened 2025-04-04
feat: log request payload when available

VonnyJap:
## Context

It will be useful for debugging to see the payload sent to API. 

## Objective

This PR serves to add payload prior to logging the response to help troubleshooting. 

## References

N/A

## License

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -14,7 +14,6 @@ process.on('unhandledRejection', (reason, p) => {
  * @param  {Hapi.h}     h   Hapi Response Toolkit
  */
 function handlePreResponseLogs(request, h) {
-
     const { response } = request;
     const { release } = request.server.app;
 
--- a/lib/server.js
+++ b/lib/server.js
@@ -9,47 +9,6 @@ process.on('unhandledRejection', (reason, p) => {
 });
 
 /**
- * If we're throwing errors, let's have them say a little more than just 500
- * @method prettyPrintErrors
- * @param  {Hapi.Request}    request Hapi Request object
- * @param  {Hapi.h}     h   Hapi Response Toolkit
- */
-function prettyPrintErrors(request, h) {
-    const { response } = request;
-    const { release } = request.server.app;
-
-    if (release && release.cookieName && request.state && !request.state[release.cookieName]) {
-        h.state(release.cookieName, release.cookieValue);
-    }
-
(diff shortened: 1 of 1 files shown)


lib/server.js
```diff
@@ -164,8 +123,51 @@ module.exports = async config => {
...
+
+            // Pretty print errors
+            if (response.isBoom) {
+                const err = response;
+                const errName = err.output.payload.error;
+                const errMessage = err.message;
+                const { statusCode } = err.output.payload;
+                const stack = err.stack || errMessage;
+
+                if (statusCode === 500) {
```
minghay, line 143: Are there any other 5xx errors that will pass through here?

lib/server.js
```diff
@@ -164,8 +123,51 @@ module.exports = async config => {
...
+                };
+
+                if (err.data) {
+                    res.data = err.data;
+                }
+
+                return h.response(res).code(statusCode);
+            }
+            // Log request payload
+            if (request.payload) {
```
minghay, line 160: Do we necessarily want to log the payload for all requests or limit the payload logging only for non-200 status codes?

jithine: we should be doing this only for debug logging, webhook payloads can get pretty big

lib/server.js
```diff
@@ -164,8 +123,51 @@ module.exports = async config => {
             });
         }
 
-        // Write prettier errors
-        server.ext('onPreResponse', prettyPrintErrors);
+        // logging behavior before response log is printed
```
jithine, line 126: why is this fn moved inline ?

lib/server.js
VonnyJap: original intent is to log only errors cause by server, which is in this case 500, it has been as such for quite sometimes, are you thinking to relax to log anything 5xx?

lib/server.js
VonnyJap: moved back the logic to a helper function

lib/server.js
VonnyJap: new commit had a change to only log when the response time is more than 5s

minghay (approved):

lib/server.js
minghay: If we're sticking with only 500 errors being returned by the server there's no particular need to relax it.  Just something to keep in mind if we ever do return other 5xx codes though.

Issue #3032, opened 2024-02-16
JavaScript version in eslint configuration is old


itleigns:
**What happened**:
JavaScript version in eslint configuration is old, so using modern JavaScript features results in errors in eslint.
**What you expected to happen**:
Modern JavaScript syntax is usable.
**How to reproduce it**:

PR #3321, opened 2025-04-11
fix: Add feature flag for control log payload

kumada626:
## Context

When payloads are logged, it can result in a large volume of logs, which might not always be desirable.
So I implement a feature flag to control the logging of payloads.

## Objective

Add feature flag for control log payload.

## License

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -49,7 +49,14 @@ function handlePreResponseLogs(request, h) {
 
     // Log request payload when it takes longer than 5 seconds to respond
     // This is to prevent logging payloads for every request
-    if (log && log.payload.enabled && request.info && request.info.received && Date.now() - request.info.received > 5000 && request.payload) {
+    if (
+        log &&
+        log.payload.enabled &&
+        request.info &&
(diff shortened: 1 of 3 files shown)


lib/server.js
```diff
@@ -49,7 +49,14 @@ function handlePreResponseLogs(request, h) {
 
     // Log request payload when it takes longer than 5 seconds to respond
     // This is to prevent logging payloads for every request
-    if (request.info && request.info.received && Date.now() - request.info.received > 5000 && request.payload) {
+    if (
+        log &&
+        log.payload.enabled &&
```
yakanechi, line 54: Also, if `log.payload &&` is present, `log.payload.enabled` ensures that undefined errors are avoided

yakanechi (approved):

y-oksaku (approved):

minghay (dismissed):

y-oksaku:
Sadly,  we cannot use optional chaining since our eslint is too old.
https://github.com/screwdriver-cd/screwdriver/issues/3032

PR #3345, opened 2025-05-21
feat: Allow SD admins to update pipelines

minghay:
## Context
There are a few cases where SD admins may be required to update pipeline settings (e.g., the repository has no admins).  In these instances there is no way for the pipeline to be recovered for any active users of the pipeline.

## Objective
Adds an additional step to determine if the user making the API request is an SD admin.  If the user is an SD admin, then the checks on repository admin privileges can be ignored.

## License

### DIFF
--- a/plugins/pipelines/update.js
+++ b/plugins/pipelines/update.js
@@ -54,8 +54,8 @@ module.exports = () => ({
         handler: async (request, h) => {
             const { checkoutUrl, rootDir, settings, badges } = request.payload;
             const { id } = request.params;
-            const { pipelineFactory, userFactory, secretFactory } = request.server.app;
-            const { scmContext, username } = request.auth.credentials;
+            const { bannerFactory, pipelineFactory, userFactory, secretFactory } = request.server.app;
+            const { scmContext, username, scmUserId } = request.auth.credentials;
             const scmContexts = pipelineFactory.scm.getScmContexts();
             const { isValidToken } = request.server.plugins.pipelines;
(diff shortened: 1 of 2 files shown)


plugins/pipelines/update.js
```diff
@@ -120,7 +135,13 @@ module.exports = () => ({
...
 
                 // get the user permissions for the repo
-                await getUserPermissions({ user, scmUri });
+                try {
+                    newRepositoryUserPermissions = await getUserPermissions({ user, scmUri });
+                } catch (err) {
+                    if (!isSdAdmin) {
+                        throw err;
+                    }
+                }
```
pritamstyz4ever, line 144: When we get user permissions from github, sd admin(github user) will not have permissions on the repo, does that mean we still override this behavior and update the pipeline?

minghay: Yeah, that's the idea.  SD admin is a permission that supersedes any SCM repository permissions that the logged in user has.  In this particular change, SD admins will not need any SCM repository permissions to make the updates to the pipeline (e.g., changing the checkout URL).  This should allow SD admins to perform the needed recovery/reconfiguration work necessary in the event that there aren't any valid admins on the SCMs.

minghay: The webhooks are going to be an issue.  I guess the best option would be to spin up a different API endpoint for SD that would allow SD admins to change the checkout URL of the pipeline.  That way it would be possible to recover a pipeline, but it would require the recovered pipeline to have an SCM admin go in and resync the webhooks.

Issue #3363, opened 2025-07-08
Add/update an endpoint to get an admin for a pipeline from the specified SCM context


sagar1312:
**What happened**:
The current endpoint `/pipelines/{id}/admin` returns an admin for a pipeline where the admin user is from the same SCM context as that of the pipeline.

**What you expected to happen**:
Need an endpoint to return an admin for a pipeline where the admin user SCM context is different pipeline SCM context.

**How to reproduce it**:

PR #3370, opened 2025-07-22
feat(3363): Update the existing endpoint to get admin for a pipeline from the specified SCM context

sagar1312:
## Context

The current endpoint `/pipelines/{id}/admin` returns an admin for a pipeline where the admin user is from the same SCM context as that of the pipeline.

## Objective

Update the endpoint to find admin for a pipeline where the admin user SCM context is different pipeline SCM context.
SCM context is passed via the query param `scmContext`

## References

https://github.com/screwdriver-cd/screwdriver/issues/3363

## License

### DIFF
--- a/plugins/pipelines/admins/get.js
+++ b/plugins/pipelines/admins/get.js
@@ -16,13 +16,19 @@ module.exports = () => ({
         tags: ['api', 'pipelines'],
         auth: {
             strategies: ['token'],
-            scope: ['user', 'pipeline', '!guest']
+            scope: ['user', 'admin', 'pipeline', '!guest']
         },
 
         handler: async (request, h) => {
-            const factory = request.server.app.pipelineFactory;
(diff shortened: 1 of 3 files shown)


plugins/pipelines/admins/get.js
```diff
@@ -20,14 +21,28 @@ module.exports = () => ({
 
         handler: async (request, h) => {
             const factory = request.server.app.pipelineFactory;
+            const { scmContext, includeUserToken } = request.query;
```
jithine, line 24: Does this mean that anyone with a  user or Pipeline APItoken can access amn admin users  token ? Sounds like a security hole ?

sagar1312: Added check to only allow user with `admin` scope to request for user token

jithine: It should only be `scope: ['admin', '!guest']`

jithine (approved):

Issue #3379, opened 2025-07-31
Non-virtual job builds may be treated as virtual when the event is created


sagar1312:
**What happened**:
Changes made in the PR https://github.com/screwdriver-cd/screwdriver/pull/3252 relied on the `status` of the builds that were created during event creation to identify virtual builds.

There could be scenarios when the non-virtual job builds may not be queued for execution and remain in `CREATED` status. This may lead to incorrectly identifying them as virtual builds (status would be updated to `SUCCESS` and downstream jobs would be triggered)

**What you expected to happen**:
Switch to more reliable approach to identify virtual job builds.

**How to reproduce it**:

PR #3380, opened 2025-07-31
fix(3379): Non-virtual job builds may be treated as virtual when the event is created

sagar1312:
## Context

Changes made in the PR https://github.com/screwdriver-cd/screwdriver/pull/3252 relied on the `status` of the builds that were created during event creation to identify virtual builds.

There could be scenarios when the non-virtual job builds may not be queued for execution and remain in `CREATED` status. This may lead to incorrectly identifying them as virtual builds (status would be updated to `SUCCESS` and downstream jobs would be triggered)

## Objective

Use the event workflow graph to identify is a build is for a virtual job.

## References

https://github.com/screwdriver-cd/screwdriver/issues/3379

## License

### DIFF
--- a/plugins/events/helper/createEvent.js
+++ b/plugins/events/helper/createEvent.js
@@ -8,6 +8,23 @@ const { Status, BUILD_STATUS_MESSAGES } = require('../../builds/triggers/helpers
  */
 
 /**
+ *
+ * @param workflowGraph
+ * @returns {Array<Number>}
+ */
+function getVirtualJobIds(workflowGraph) {
+    const virtualJobIds = [];
+
+    workflowGraph.nodes.forEach(node => {
+        if (node.virtual) {
+            virtualJobIds.push(node.id);
+        }
+    });
+
+    return virtualJobIds;
+}
+
+/**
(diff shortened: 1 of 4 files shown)


plugins/events/helper/createEvent.js
```diff
@@ -7,6 +7,23 @@ const { Status, BUILD_STATUS_MESSAGES } = require('../../builds/triggers/helpers
...
+    const virtualJobIds = [];
+
+    workflowGraph.nodes.forEach(node => {
+        if (node.virtual) {
+            virtualJobIds.push(node.id);
+        }
+    });
+
+    return virtualJobIds;
+}
```
pritamstyz4ever, line 25: using map and filter is always better than forEach.

pritamstyz4ever (approved):

jithine (approved):

kumada626:
@sagar1312 In this change, PR cases are not being handled properly. Probably because of this, virtual jobs not being executed in PR builds.
https://beta.cd.screwdriver.cd/pipelines/25006/pulls

PR #3383, opened 2025-08-05
fix: Make webhook payload size limits configurable

kumada626:
## Context

After #3371, webhook payload size limit was hardcoded.
Once again, we have made it possible to set webhook size limits.

## Objective

- Add webhook size limit settings environment in the config file.
- If a size limit is set in the config file, use that value.

## License


config/custom-environment-variables.yaml
```diff
@@ -361,32 +361,33 @@ scms:
...
+    #  "github.example.com": {
+    #     "username": "someuser",
+    #     "ignoreCommitsBy": ["someuser", "anotheruser"],
+    #     "restrictPR": "branch",
+    #     "chainPR": true
+    #   }
+    # }
+    __name: WEBHOOK_SETTINGS
+    __format: json
+  maxBytes: WEBHOOK_MAX_BYTES
```
pritamstyz4ever, line 388: this configuration is fine, but it will be common for all scm context

plugins/webhooks/index.js
```diff
@@ -77,7 +78,7 @@ const webhooksPlugin = {
                         for await (const chunk of request.payload) {
```
pritamstyz4ever, line 78: probably it makes sense to get rid of custom logic for payload size and let hapi js do it in the payload configuration.
so I propose removing this and set
 payload: {
                    maxBytes: parseInt(pluginOptions.maxBytes, 10)

pritamstyz4ever (approved):

PR #3385, opened 2025-08-06
fix: update stage build status properly

y-oksaku:
## Context

The stage build status is not updated when

- Stop the event.
- All jobs in the stage are virtual.

Furthermore, the status transition as `created -> queued -> running -> success -> queued -> running -> ... -> success` since it is not blocked transitioning from success to queued.

## Objective

- Update stage build status properly
- Refactor the stage build status transition flow

## License


plugins/builds/helper/updateBuild.js
```diff
@@ -241,6 +270,50 @@ function deriveEventStatusFromBuildStatuses(builds) {
...
+ *
+ * @param {Object} param
+ * @param {StageBuild} param.stageBuild
+ * @param {String} param.newStatus
+ * @param {Job} param.job
+ * @returns {StageBuild} new stage build's status
+ */
+async function updateStageBuildStatus({ stageBuild, newStatus, job }) {
+    if (stageBuild.status === newStatus) {
+        return stageBuild.status;
```
yk634, line 287: I don’t think this causes any issues in the current usage, but it feels a bit inconsistent that the return value of the function can be either a `string` (status) or a `stageBuild` object.

It might be cleaner to always return the `stageBuild` itself and drop the `.status` return to keep the return type consistent.

y-oksaku: Thank you. I must have overlooked it.

minghay (approved):

PR #3406, opened 2025-09-26
fix(3400)[2]:use raw query in badge api

yuichi10:
## Context

Pipeline id and create time index cause a loading problem in our environment.
https://github.com/screwdriver-cd/data-schema/pull/607

We want to use raw query to improve badge api instead of using an index.

## Objective

- using raw query 
  - https://github.com/screwdriver-cd/models/pull/656

## License


foxtrot0304 (approved):
lgtm

PR #3454, opened 2026-01-22
fix(3455): Virtual jobs processes blockedby params

rudra5:
## Context

Virtual jobs were not checking if any `blockedby` jobs were currently running

## Objective

Check for `blockedby` jobs when running Virtual jobs

## License

### DIFF
--- a/plugins/builds/triggers/orBase.js
+++ b/plugins/builds/triggers/orBase.js
@@ -49,7 +49,7 @@ class OrBase {
         });
 
         const hasWindows = hasFreezeWindows(nextJob);
-        const isBlocked = hasBlockedBy(nextJob);
+        const hasBlocked = hasBlockedBy(nextJob);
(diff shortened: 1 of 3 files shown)


plugins/builds/triggers/orBase.js
```diff
@@ -49,6 +49,7 @@ class OrBase {
         });
 
         const hasWindows = hasFreezeWindows(nextJob);
+        const isBlocked = hasBlockedBy(nextJob);
```
sagar1312: Consider renaming
```suggestion
        const jobHasBlockedBy = hasBlockedBy(nextJob);
```
Also `hasWindows` to `jobHasFreezeWindows`

sagar1312 (changes requested):
Also, we need to submit the virtual build to queue when an event is started from a virtual job to honor `blockedBy`
https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/events/helper/createEvent.js#L83

Please add tests.

Issue #2727, opened 2022-07-05
Unable to download large build logs from API


ibu1224:
**What happened**:
In my environment, when I try to download about 500,000 lines of build logs, I get a 504 Gateway timeout between the API and the UI.

In the following Issue. It appears that the build log download has been moved to the API.
https://github.com/screwdriver-cd/screwdriver/issues/2068

With the current approach, if the number of waiting seconds between each component is exceeded, the download will fail.

We have to think of some way.

**What you expected to happen**:
Able to download large size build logs.

The following comments were made within the YJ members.
- Select the required build log from the UI and make it available for download.
- Use Multi-part download in the UI.

**How to reproduce it**:
Download large sized build logs from the UI

s-yoshika:
@minghay We’re about to start working on this issue on our end, so could you please reopen it?

PR #3467, opened 2026-02-24
fix(2727): Stream log download process

s-yoshika:
## Context
In the current log download implementation, the entire log file is loaded from Store all at once before being returned. As a result, downloading large build logs can sometimes cause errors due to insufficient memory.

## Objective
This PR addresses the issue by using streams to retrieve logs from Store, which improves memory usage.

## References
https://github.com/screwdriver-cd/screwdriver/issues/2727

## License

### DIFF
--- a/test/plugins/builds.test.js
+++ b/test/plugins/builds.test.js
@@ -7753,6 +7753,18 @@ describe('build plugin test', () => {
             });
         });
 
+        it('returns 400 when the timestampFormat is unknown', () => {
+            nock('https://store.screwdriver.cd')
+                .get(`/v1/builds/${id}/${step}/log.0`)
+                .replyWithFile(200, `${__dirname}/data/step.log.ndjson`);
+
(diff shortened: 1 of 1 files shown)


plugins/builds/steps/logs.js
```diff
@@ -276,84 +346,54 @@ module.exports = config => ({
...
+                            .type('text/plain')
+                            .header('content-disposition', `attachment; filename="${stepName}-log.txt"`);
+                    });
                 })
                 .catch(err => {
                     throw err;
                 });
         },
-        response: {
-            schema: schema.api.loglines.output
```
s-yoshika: We have removed the response schema validation.
When returning a response as a stream, the response body to be validated has not yet been generated, so it cannot be validated.

Instead, for requests where type is not download, validation is performed inside the handler.
https://github.com/sonic-screwdriver-cd/screwdriver/blob/ebcb21632e618d0615eb4d309490704fc8d76fc8/plugins/builds/steps/logs.js#L362

foxtrot0304 (approved):
lgtm

Issue #3475, opened 2026-03-12
Support enabling/disabling of pipeline


sagar1312:
**What happened**:
Screwdriver supports enabling and disabling individual jobs to prevent builds from being created or executed for those jobs. This is useful for temporarily pausing specific parts of a pipeline without removing them.                                                                                                   
                                                                                                                                                                                                                                                                                                                          
However, for large pipelines with many jobs, managing this at the job level becomes time-consuming and inefficient — each job must be disabled individually. There is currently no way to disable an entire pipeline in one action.

Additionally, pipelines currently only support three states: ACTIVE, INACTIVE (used for child pipelines managed by a config pipeline), and DELETING. There is no first-class concept of a user-initiated disabled state at the pipeline level.

**What you expected to happen**:

**New pipeline state: DISABLED**                                                                                                                                                                                                                                                                                              
	  - The system shall support a new DISABLED state for pipelines, in addition to the existing ACTIVE, INACTIVE, and DELETING states.
	  - State transitions via the update API shall be restricted to ACTIVE → DISABLED and DISABLED → ACTIVE only. Any other transition must be rejected.                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                          
**Update pipeline API (PUT /pipelines/{id})**
	  - The pipeline update API shall accept two new optional fields: state (valid values: ACTIVE, DISABLED) and stateChangeMessage (free-text reason for the change).
	  - When state is changed, the system shall automatically record the username of the person who made the change (stateChanger) and the timestamp of the change (stateChangeTime). These fields must not be settable by the caller.
	  - If state is not provided in the request, the pipeline's existing state shall be retained unchanged.
	  - Pipeline admins shall be allowed to update state and stateChangeMessage in addition to all existing updatable fields (checkout URL, settings, badges, etc.).
	  - Screwdriver cluster admins who are not pipeline admins shall be allowed to update only state and stateChangeMessage. Requests from such users that include any other fields must be rejected with an appropriate error.
	  - Child pipelines shall allow updates to state and stateChangeMessage only. All other fields must be rejected. Previously, all updates to child pipelines were blocked entirely.

**Block event creation for disabled pipelines**
	  - The system shall reject event creation requests for pipelines in the DISABLED state, the same way it currently rejects them for INACTIVE and DELETING pipelines. This prevents builds from being triggered for any job in the pipeline while it is disabled.

**Child pipeline sync**
[...]

PR #3476, opened 2026-03-13
feat(3475): support enabling/disabling of pipeline

sagar1312:
## Context

Screwdriver supports enabling and disabling individual jobs to prevent builds from being created or executed for those jobs. This is useful for temporarily pausing specific parts of a pipeline without removing them.

However, for large pipelines with many jobs, managing this at the job level becomes time-consuming and inefficient — each job must be disabled individually. There is currently no way to disable an entire pipeline in one action.

Additionally, pipelines currently only support three states: ACTIVE, INACTIVE (used for child pipelines managed by a config pipeline), and DELETING. There is no first-class concept of a user-initiated disabled state at the pipeline level.

## Objective

- Add DISABLED pipeline state with ACTIVE<->DISABLED transition validation
- Update PUT /pipelines/{id} to accept state and stateChangeMessage fields
- Pipeline admins can update state alongside all existing fields
- Screwdriver cluster admins (non-pipeline-admins) can update state fields only
- Child pipelines now allow state-only updates instead of blocking all updates
- Block event creation for DISABLED pipelines in POST /events
- Preserve DISABLED state on child pipeline sync (no longer reset to ACTIVE)
- Bump screwdriver-data-schema to ^26.1.0 and screwdriver-models to ^33.3.0

## References

https://github.com/screwdriver-cd/screwdriver/issues/3475
https://github.com/screwdriver-cd/data-schema/pull/612
https://github.com/screwdriver-cd/models/pull/664

## License


package.json
```diff
@@ -122,7 +123,7 @@
     "screwdriver-executor-queue": "^6.0.0",
     "screwdriver-executor-router": "^5.0.0",
     "screwdriver-logger": "^3.0.0",
-    "screwdriver-models": "^33.2.0",
+    "screwdriver-models": "^33.3.0",
```
minghay:
```suggestion
    "screwdriver-models": "^34.1.0",
```

minghay (approved):

PR #3498, opened 2026-05-29
fix: dedupe webhook deliveries within a short TTL

sagar1312:
## Summary
- New singleton `plugins/webhooks/dedupStore.js` tracks each incoming `x-github-delivery` for a configurable TTL (default 5 minutes) and silently drops duplicates.
- Webhook handler returns `204 No Content` with an empty body on duplicate so the SCM stops retrying and an attacker probing IDs cannot distinguish a seen ID from an unseen one.
- Reuses the Redis client owned by `plugins/lock.js` — operators configure one Redis target via `redisLock.options` and both features use it. Falls back to a per-process in-memory `Map` when Redis is unavailable.
- Fail-open: any Redis error logs a warning and lets the delivery through, so replay-protection infrastructure cannot block legitimate webhooks. Includes an `error` listener on the shared Redis client to prevent unhandled error events from crashing the process.
- `parseBool` is lifted to `plugins/helper.js` so `lock.js` and `dedupStore.js` share a single implementation.
- Bumps `screwdriver-scm-github` to `^14.6.2`.

## Why 5 minutes
Long enough to defeat realistic fast-replay attempts (an attacker who captured a delivery and replays it within minutes), short enough that legitimate manual redelivery from the SCM UI — which preserves the original `x-github-delivery` ID — is unaffected in practice. The TTL is configurable via \`webhooks.replayProtection.ttlSeconds\`.

## Test plan
- New \`test/plugins/webhooks.dedupStore.test.js\` covers: disabled state, Redis-backed claim (fresh + duplicate + fail-open + default TTL), in-memory fallback (claim + TTL eviction with fake timers + per-key independence + fallback when lock has no Redis).
- New test in \`test/plugins/webhooks.test.js\` asserts the handler returns 204 with empty body and skips downstream processing on duplicate.
- Existing webhook plugin tests continue to pass.
- \`npm run lint\` clean; full plugin suite green except one pre-existing unrelated jobs.test.js failure.

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -278,6 +278,20 @@ scms: {}
 #         # SCM clone type (https or ssh)
 #         cloneType: https
 webhooks:
+  # Replay protection for incoming webhook deliveries. Dedupes identical
+  # x-github-delivery values within a short TTL so a captured webhook cannot be
+  # replayed by an attacker. The Redis backend reuses the connection configured
+  # under redisLock.options.redisConnection (a single Redis instance generally
+  # serves both features). When no Redis host is configured, falls back to a
+  # per-process in-memory store. Fail-open: infrastructure errors never block
+  # legitimate webhook deliveries.
+  replayProtection:
+    # Master switch. Set to false to disable replay protection entirely.
+    enabled: true
+    # Window in which a duplicate x-github-delivery is rejected. Defaults to 5
+    # minutes — long enough to defeat realistic fast-replay attempts, short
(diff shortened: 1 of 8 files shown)


config/default.yaml
minghay, line 447: Do you want to set this to true now?

sagar1312: No. This is a shared config and depends on if redis has been setup.
We can leave it as `false`. It is (would be) set to `true` in the SD cluster by admins when they have redis setup.

minghay (approved):

Issue #3508, opened 2026-07-30
[Feature Request] Enhance API Token Permissions and Expiration


y-oksaku:
**What happened**:

The current API Token implementation does not allow configuring token permissions. Every API Token is granted the same level of privileges as the user, resulting in a JWT token with full user permissions.

In addition, API Tokens do not support expiration, so only non-expiring tokens can be issued.

**What you expected to happen**:

API Tokens should support configurable permissions and expiration settings.

This issue is intended to propose extending the API Token functionality to allow finer-grained permission control and configurable expiration.

y-oksaku:
# Proposal

## 1. Add columns to the `tokens` table

Add the following columns to the `tokens` table.

### `expiresAt`

The token expiration timestamp.

When creating a token, users should be able to select one of the following expiration periods:

* 1 day
* 30 days
* 90 days
* 1 year
* No expiration

The calculated expiration timestamp should be stored in this column.

Example: `2026-12-01T12:00:11Z`  
For tokens without an expiration date, the value should be `NULL`.

### `issuerId`

The user ID of the user who created the token.

Example: `123`

### `options`

The token's permission and resource restrictions.

The following permission levels should be supported:

| Permission | Description |
| --- | --- |
| `read` | Allows read-only access to resources such as builds, jobs, and pipelines. |
| `execute` | Includes `read` permissions and allows builds to be started or stopped. |
| `write` | Includes `execute` permissions and allows operations such as enabling or disabling jobs and deleting caches. |
| `all` | Allows all operations. |

For backward compatibility, existing API Tokens should be assigned the `all` permission.

For a User Token, `options` would have the following structure:

```json
{
  "permission": "read",
  "resources": {
    "pipelines": [12, 34],
    "privatePipeline": true,
    "organizations": ["org1", "org2"]
  }
}
```

The supported value for `permission` is one of:

```text
read | execute | write | all
```

For a Pipeline Token, `options` would have the following structure:

```json
{
  "permission": "execute",
  "resources": {
    "jobs": [111, 222]
  }
}
```

## 2. Add authentication and permission information to JWT claims

Add authentication source, permission, and resource information to issued JWTs.

The `permission` and `resources` claims should be populated from the corresponding values stored in the `tokens` table.

An `auth` claim should also be added so that the API can distinguish between JWTs issued from an API Token, OAuth authentication, and temporary authentication.

Example:

```json
{
  "username": "foobar",
  "scmContext": "github.com",
[...]

PR #3513, opened 2026-08-14
feat(3508): configure permissions for pipeline endpoints

raihata:
## Context

The authorization hook compares the permission required by an endpoint with the permission granted to the JWT. This PR adds permission settings to endpoints under `plugins/pipelines`.

## Objective

- Assign permissions to authenticated pipeline endpoints:
  - `read` for read-only operations
  - `execute` for starting child pipelines
  - `write` for deleting pipeline caches
  - `all` for other privileged operations
- Keep public badge and template-validation routes unchanged.
- Require both the Screwdriver `admin` scope and `all` permission when `GET /pipelines/{id}/admin` is called with `includeUserToken=true`.
- Preserve compatibility with legacy JWTs that do not contain a permission.
- Add tests to comprehensively verify the permission settings for each route.

## References
- [screwdriver-cd/screwdriver#3508](https://github.com/screwdriver-cd/screwdriver/issues/3508)

## License

### DIFF
--- a/plugins/pipelines/admins/get.js
+++ b/plugins/pipelines/admins/get.js
@@ -18,13 +18,18 @@ module.exports = () => ({
             strategies: ['token'],
             scope: ['user', 'admin', 'pipeline', '!guest']
         },
+        plugins: {
+            authorization: {
+                permission: 'read'
+            }
+        },
 
         handler: async (request, h) => {
             const { pipelineFactory } = request.server.app;
-            const { scope } = request.auth.credentials;
+            const { scope, permission = 'all' } = request.auth.credentials;
             const { scmContext, includeUserToken } = request.query;
 
-            if (includeUserToken && !scope.includes('admin')) {
+            if (includeUserToken && (!scope.includes('admin') || permission !== 'all')) {
                 throw boom.forbidden('Only Screwdriver admin is allowed to request user token');
             }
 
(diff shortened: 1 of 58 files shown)


kumada626 (approved):

plugins/pipelines/admins/get.js
```diff
@@ -18,13 +18,18 @@ module.exports = () => ({
...
+        plugins: {
+            authorization: {
+                permission: 'read'
+            }
+        },
 
         handler: async (request, h) => {
             const { pipelineFactory } = request.server.app;
-            const { scope } = request.auth.credentials;
+            const { scope, permission = 'all' } = request.auth.credentials;
```
yakanechi, line 29: Why is the default value set to “all” here?
Is it for backward compatibility?

yakanechi (approved):

yk634 (approved):

plugins/pipelines/admins/get.js
raihata: Yes, this is for backward compatibility.

Legacy JWTs do not have a permission claim, and `onPostAuth` treats a missing permission as `all`. Since this endpoint performs an additional permission check when `includeUserToken=true`, it needs the same fallback. Otherwise, legacy Screwdriver admin JWTs would unexpectedly receive a 403 response.

An explicit `read` permission is not affected by this fallback and is still rejected.

minghay (approved):

PR #3514, opened 2026-08-17
feat(3508): configure API token permissions for shared resource endpoints

raihata:
## Context

Shared resource routes must declare their required permission so that API token permissions are enforced consistently.

## Objective

Configure permissions for 37 routes across templates, commands, collections, users, and banners.

- User-token generation additionally requires `all` permission. 
- Public banner routes remain unchanged. 
- Tests are added for the configured permissions.

## References

- [https://github.com/screwdriver-cd/screwdriver/issues/3508](https://github.com/screwdriver-cd/screwdriver/issues/3508)

## License

### DIFF
--- a/plugins/commands/create.js
+++ b/plugins/commands/create.js
@@ -108,6 +108,11 @@ module.exports = () => ({
             strategies: ['token'],
             scope: ['build', '!guest']
         },
+        plugins: {
+            authorization: {
+                permission: 'all'
+            }
+        },
 
         payload: {
             parse: true,
--- a/plugins/commands/create.js
+++ b/plugins/commands/create.js
@@ -108,11 +108,6 @@ module.exports = () => ({
             strategies: ['token'],
             scope: ['build', '!guest']
         },
-        plugins: {
-            authorization: {
-                permission: 'all'
-            }
-        },
 
         payload: {
             parse: true,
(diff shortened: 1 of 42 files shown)


plugins/commands/create.js
```diff
@@ -108,6 +108,11 @@ module.exports = () => ({
             strategies: ['token'],
             scope: ['build', '!guest']
         },
+        plugins: {
```
yk634: This endpoint only accepts JWTs with the `build` scope. JWTs issued from API Tokens have either the `user` or `pipeline` scope, so they cannot access this endpoint. Build Token JWTs also have `auth.type: temporary` and therefore bypass the API Token permission check in `onPostAuth`.

It seems that the newly added `authorization.permission` setting is not enforced here. Do we need permission settings on endpoints that are only accessible with Build Tokens?

raihata: Thank you for pointing this out. You are right: this permission setting is ineffective because onPostAuth only checks permissions for API Token JWTs.

I reviewed the other routes in this PR and removed the permission settings from all six Build Token-only endpoints in commands and templates. 
- POST /commands
- PUT /commands/{namespace}/{name}/tags/{tagName}
- DELETE /commands/{namespace}/{name}/tags/{tagName}
- POST /templates
- PUT /templates/{templateName}/tags/{tagName}
- DELETE /templates/{templateName}/tags/{tagName}

I also updated the corresponding tests.

yk634 (approved):

minghay (approved):

PR #3516, opened 2026-08-17
feat(3508): configure API token permissions for runtime endpoints

raihata:
## Context

Authenticated runtime routes must declare their required permission so that limited API tokens cannot perform operations beyond their assigned role.

## Objective

Configure permissions for 42 routes across builds, jobs, stages, build clusters, secrets, tokens, coverage, and the admin check endpoint.

Additionally, require both Screwdriver administrator scope and `all` permission when generating an impersonated build token through `/auth/token/{buildId}`. Legacy JWTs without a permission value remain compatible.

Existing route scopes and resource-specific authorization remain unchanged. Tests are added for the configured permissions and build impersonation checks.

## References

- [https://github.com/screwdriver-cd/screwdriver/issues/3508](https://github.com/screwdriver-cd/screwdriver/issues/3508)

## License

### DIFF
--- a/plugins/jobs/update.js
+++ b/plugins/jobs/update.js
@@ -17,6 +17,11 @@ module.exports = () => ({
             strategies: ['token'],
             scope: ['user', '!guest', 'pipeline']
         },
+        plugins: {
+            authorization: {
+                permission: 'write'
+            }
+        },
 
         handler: async (request, h) => {
             const { jobFactory, pipelineFactory, userFactory, bannerFactory } = request.server.app;
--- a/plugins/auth/index.js
+++ b/plugins/auth/index.js
@@ -115,13 +115,13 @@ const authPlugin = {
                 throw boom.unauthorized();
             }
 
-            const credentials = request.auth.credentials;
+            const { credentials } = request.auth;
 
             if (!credentials.auth || credentials.auth.type !== 'api_token') {
                 return h.continue;
             }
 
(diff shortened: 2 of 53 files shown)


raihata:
The permission check in onPostAuth applies only to API Token JWTs. The following endpoints accept only Build or Temporal Token JWTs, so their authorization.permission settings were ineffective:
- PUT /builds/{id}/steps/{name}
- POST /builds/{id}/token
- GET /coverage/token

I removed these settings and updated the tests accordingly. The existing auth.scope restrictions remain unchanged.

plugins/jobs/update.js
```diff
@@ -17,6 +17,11 @@ module.exports = () => ({
             strategies: ['token'],
             scope: ['user', '!guest', 'pipeline']
         },
+        plugins: {
+            authorization: {
+                permission: 'write'
```
yakanechi, line 22: Is “write” the right permissions to use here?

raihata: This is because `PUT /jobs/{id}` is used to toggle a job between ENABLED and DISABLED states and to update the reason for the change.
Based on the proposal, `write` is the appropriate permission.
https://github.com/screwdriver-cd/screwdriver/issues/3508

yk634 (approved):

yakanechi (approved):

minghay (approved):

PR #3518, opened 2026-08-21
Fix: coverage git app binding

minghay:
## Context
Second half of the work needed to bind the Sonar tokens to the project.  https://github.com/screwdriver-cd/coverage-sonar/pull/80 contains the main work for managing and validating the sonar project to pipeline/job.

## Objective
Updates the API for getting Sonar project tokens to require just the projectKey parameter

## References
https://github.com/screwdriver-cd/coverage-sonar/pull/80

## License

### DIFF
--- a/plugins/coverage/README.md
+++ b/plugins/coverage/README.md
@@ -27,7 +27,12 @@ server.register({
 ### Routes
 
 #### Returns an access token to talk to coverage server
-`GET /coverage/token?scope=job&projectKey=job:123&projectName=d2lam/mytest&username=user-job-123`
+`GET /coverage/token?projectKey=job:123`
+
+`scope`, `projectName`, and `username` are not accepted as query parameters — all three are always resolved
+server-side from the build's own JWT-verified identity. `projectKey` is accepted only if it names a project
+the calling build is authorized for (its own pipeline, its own job, or its PR parent job); anything else is
+rejected with a `403`.
 
 #### Get an object with coverage info
 
--- a/plugins/coverage/README.md
+++ b/plugins/coverage/README.md
@@ -30,9 +30,13 @@ server.register({
 `GET /coverage/token?projectKey=job:123`
 
 `scope`, `projectName`, and `username` are not accepted as query parameters — all three are always resolved
-server-side from the build's own JWT-verified identity. `projectKey` is accepted only if it names a project
-the calling build is authorized for (its own pipeline, its own job, or its PR parent job); anything else is
-rejected with a `403`.
+server-side from the build's own JWT-verified identity. `projectKey` is rejected with a `403` unless it
+names a project the calling build is authorized for (its own pipeline, its own job, or its PR parent job)
+**and** matches the build's own resolved coverage scope.
+
+> **Deployment order:** requires `screwdriver-coverage-sonar` ≥6.0.0
+> ([coverage-sonar#80](https://github.com/screwdriver-cd/coverage-sonar/pull/80)). This API change must be
+> deployed *before* that dependency is bumped — see that PR's README for why.
 
 #### Get an object with coverage info
 
--- a/plugins/coverage/token.js
+++ b/plugins/coverage/token.js
@@ -22,7 +22,10 @@ module.exports = config => ({
             const { jobFactory, pipelineFactory } = request.server.app;
             const buildCredentials = request.auth.credentials;
             const { jobId, pipelineId } = buildCredentials;
-            const { scope, projectKey, projectName, username, selfSonarHost, selfSonarAdminToken } = request.query;
+            const { scope, projectKey, selfSonarHost, selfSonarAdminToken } = request.query;
+            // `projectName` and `username` are deliberately not read from the query. The coverage plugin
+            // derives both from the pipeline name resolved below, so a build cannot point its Sonar
+            // project's Git App binding at another organization's repository. See PSECBUGS-115276.
             const tokenConfig = {
                 buildCredentials,
                 scope
@@ -30,14 +33,6 @@ module.exports = config => ({
 
             if (projectKey) {
                 tokenConfig.projectKey = projectKey;
-            }
-
-            if (projectName) {
-                tokenConfig.projectName = projectName;
-            }
-
-            if (username) {
-                tokenConfig.username = username;
             }
 
             // Get scope and job name
@@ -56,8 +51,8 @@ module.exports = config => ({
             }
             let pipeline;
 
-            // Get pipeline name
-            if (pipelineId && (!projectName || projectName.includes('undefined'))) {
+            // Get pipeline name; always resolved here so it can never be supplied by the caller
+            if (pipelineId) {
                 pipeline = await pipelineFactory.get(pipelineId);
 
                 if (!pipeline) {
@@ -90,12 +85,6 @@ module.exports = config => ({
 
             const data = await config.coveragePlugin.getAccessToken(tokenConfig);
             const { projectUrl } = config.coveragePlugin.getProjectData(tokenConfig);
-
-            if (!pipeline && pipelineId) {
-                pipeline = await pipelineFactory.get(pipelineId);
-
-                logger.info(`looking up again, pipeline:${pipelineId}, and found pipeline: ${pipeline}`);
-            }
 
             if (pipeline && projectUrl) {
                 try {
--- a/plugins/coverage/token.js
+++ b/plugins/coverage/token.js
@@ -21,7 +21,7 @@ module.exports = config => ({
         handler: async (request, h) => {
             const { jobFactory, pipelineFactory } = request.server.app;
             const buildCredentials = request.auth.credentials;
-            const { jobId, pipelineId, prParentJobId } = buildCredentials;
+            const { jobId, pipelineId } = buildCredentials;
             const { projectKey, selfSonarHost, selfSonarAdminToken } = request.query;
             // `scope`, `projectName`, and `username` are deliberately not read from the query. The coverage
             // plugin derives all three from records resolved below via the build's own JWT-verified ids, so
@@ -83,19 +83,7 @@ module.exports = config => ({
             }
 
             const data = await config.coveragePlugin.getAccessToken(tokenConfig);
-            // projectKey is deliberately not forwarded here: getAccessToken above only actually uses a
-            // caller-supplied projectKey when it's both authorized and scope-consistent, silently ignoring
-            // it otherwise, so re-deriving from the same scope/jobName/pipelineName inputs is what keeps
-            // this badge URL consistent with the project the minted token is actually scoped to.
-            const { projectUrl } = config.coveragePlugin.getProjectData({
-                enterpriseEnabled: config.coveragePlugin.config.sonarEnterprise,
-                jobId,
-                jobName: tokenConfig.jobName,
-                pipelineId,
-                pipelineName: tokenConfig.pipelineName,
-                prParentJobId,
-                scope: tokenConfig.scope
-            });
+            const { projectUrl } = config.coveragePlugin.getProjectData(tokenConfig);
 
             if (pipeline && projectUrl) {
                 try {
--- a/plugins/coverage/token.js
+++ b/plugins/coverage/token.js
@@ -43,6 +43,13 @@ module.exports = config => ({
                 }
 
                 tokenConfig.jobName = job.name;
+                // Invariant this rejection below relies on: screwdriver-cd/launcher's coverage/info call
+                // (launch.go, GetCoverageInfo) reads this exact field/index -
+                // job.Permutations[0].Annotations.CoverageScope - and always sends it as `scope`, empty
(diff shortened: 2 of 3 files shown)


DekusDenial:
Thanks for splitting this out. Resolving `pipelineName` server-side and dropping the caller-supplied `projectName`/`username` is the right call — that's what stops a build pointing its Sonar project's Git App binding at another org's repository.

I ran both PR heads against the real route/plugin pair. One ordering issue to call out, plus a few behavior changes that go past the original ticket.

---

## Blocking

### 1. This PR needs to merge and deploy *before* coverage-sonar#80

The description points at #80 as "the main work," which reads as though #80 lands first. That's the order that breaks.

#80 stops trusting `projectName` and derives it from `pipelineName`, but the API today only sends `pipelineName` when `projectName` is absent. In the window where #80 is merged and this isn't, `pipelineName` is undefined and `configureGitApp` skips binding entirely — 200, valid token, PR decoration silently not configured for newly created enterprise projects.

Same request in each combination:

```
master + master   (today)                -> 200  gitAppBinding=my-org/my-repo
PR80  + master    (#80 merges first)     -> 200  gitAppBinding=NOT BOUND      <--
PR80  + PR3518    (both merged)          -> 200  gitAppBinding=my-org/my-repo
```

Good news: **this PR is safe standalone.** Against current `coverage-sonar`, it produces the same project and the same binding as today, including when the job annotation is absent:

```
master sonar + master route   : 200 project=pipeline:333 binding=my-org/my-repo
master sonar + PR3518 route   : 200 project=pipeline:333 binding=my-org/my-repo
master sonar + PR3518, no ann : 200 project=pipeline:333 binding=my-org/my-repo
```

So **this one first, then #80.** Could we note that in both PR bodies so whoever merges doesn't pick the other order?

---

## Worth discussing

### 2. Dropping `scope` from the query removes a consistency guarantee

Today the `scope` the build sends is the *already-resolved* `projectScope` that `_getInfo` computed, and `projectKey` comes from that same call — so the token's scope and the scanner's `SD_SONAR_PROJECT_KEY` cannot disagree, by construction. After this PR they're derived through two independent paths: `/coverage/info` resolves scope from the launcher's query, `/coverage/token` resolves it from the job annotation.

Paired with #80's new rule that discards a `projectKey` whose prefix doesn't match the derived scope, a disagreement silently redirects the build:

```
Build sends projectKey=job:123&scope=job, enterprise, no job annotation:

  master + master  -> 200  project=job:123
  PR80  + PR3518   -> 200  project=pipeline:333   <-- token for a different project
```

The token is minted for `pipeline:333` while the scanner uploads to `job:123`, so the upload 403s and coverage is silently lost — `commands.txt` runs the scanner under `|| true`, so the build still goes green.

The key question is whether the launcher's `scope` for `/coverage/info` can ever differ from `permutations[0].annotations['screwdriver.cd/coverageScope']`. That's outside both repos — do you know offhand? Matrix jobs look like a candidate, since the annotation is read from `permutations[0]` only. If they can't diverge this is moot, but I'd want to confirm rather than assume.

Note that `tokenConfig.scope` is now also `null` whenever the annotation is missing, where previously a caller-supplied scope was honored.

### 3. `if (jobId)` is now unconditional

Two consequences, both outside the ticket:

- One extra `jobFactory.get` per token request, in the case where `scope` was previously supplied and the lookup was skipped.
- `job missing + scope previously supplied` goes 200 → 404.

Both are defensible, just worth being deliberate about. Worth noting `jobName` isn't actually needed for the Git App binding — `_getRepositoryName` strips everything after the `:` — so if the goal is only to fix the binding, the job lookup can stay conditional. It *is* needed for deriving scope, so this follows from #2 rather than from the binding fix.

### 4. Re-deriving `projectUrl` is a separate bug fix

```js
const { projectUrl } = config.coveragePlugin.getProjectData({ enterpriseEnabled, jobId, jobName, pipelineId, pipelineName, prParentJobId, scope });
```

This is a genuine fix — passing `tokenConfig` meant `getProjectData` saw no top-level `jobId`/`pipelineId` and derived `pipeline:undefined`, writing a dead badge URL. But it's a different bug from the one this PR is about, and it will rewrite `pipeline.badges` rows on the next build for affected pipelines.

No objection to fixing it, just flagging that it's unrelated scope and has a data-write side effect — might be cleaner as its own PR so it can be reverted independently if a badge URL turns out wrong.

---

## Nit

The unconditional pipeline lookup means `if (!pipeline && pipelineId)` was dead, so removing it is correct. Worth noting for reviewers that this doesn't add a DB read on the main path — the old re-fetch already fetched the pipeline unconditionally, just later and without the 404. The one place it does add a fetch is the `selfSonarHost` path, which returned before the re-fetch.

DekusDenial:
Re-reviewed against the new commit. The two things I raised are both addressed, and I checked
  the `projectUrl` revert empirically rather than eyeballing it.

  ## Resolved — verified

  **The `projectUrl` revert is genuinely behavior-neutral.** `getProjectData(tokenConfig)` is now
  character-for-character master's call, and I confirmed the badge URL matches master across
  every shape:

  ```
  ann=<unset>  ent=true   master=.../dashboard?id=pipeline%3A333   #3518=.../dashboard?id=pipeline%3A333   SAME
  ann=<unset>  ent=false  master=undefined                        #3518=undefined                         SAME
  ann=pipeline ent=true   master=.../dashboard?id=pipeline%3A333   #3518=.../dashboard?id=pipeline%3A333   SAME
  ann=pipeline ent=false  master=.../dashboard?id=pipeline%3A333   #3518=.../dashboard?id=pipeline%3A333   SAME
  ann=job      ent=true   master=undefined                        #3518=undefined                         SAME
  ann=job      ent=false  master=undefined                        #3518=undefined                         SAME
  ```

  Worth noting for whoever picks up the follow-up: the `pipeline:undefined` bug doesn't actually
  surface for real builds, because `/coverage/info` always puts a `projectKey` in the auth URL and
  `getProjectData`'s `if (projectKey)` branch derives the id from *that* rather than from the
  absent `tokenConfig.pipelineId`. It only bites a caller that omits `projectKey`. Still a real
  bug, still worth its own PR, but lower urgency than it looked in round 1 — and splitting it out
  was the right call given it rewrites `pipeline.badges` rows as a side effect.

  Dropping `prParentJobId` from the destructure is correct now that the explicit config is gone —
  it has no other reader in this file, and `buildCredentials` still carries it through to
  `getAccessToken`.

  **Deployment order documented.** Combined with the `BREAKING CHANGE` footer on coverage-sonar#80
  (which takes it to `6.0.0`, outside our `^5.0.0` range), the ordering is now enforced by npm
  rather than by people remembering. That fully addresses my round-1 blocker.

  ## One loose end

  **Nothing actually bumps the dependency.** `package.json` still says
  `"screwdriver-coverage-sonar": "^5.0.0"`, and the README now says we require `≥6.0.0`. That's
  the right call for *this* PR — the whole point is that this deploys first — but it does mean
  there's a mandatory third step that currently exists only in prose across two READMEs. Could we
  get a tracking issue, or a `TODO` next to the dep, so the bump doesn't get lost? Right now a
  reader of `package.json` alone has no signal that the pin is deliberately stale.

  ## Discussion — the scope derivation

  Same underlying point as my comment on coverage-sonar#80; I won't duplicate the evidence, but
  it's worth stating from this side since this is the PR that makes the change.

  Master reads the annotation only under `if (jobId && !scope)`. Since `_getInfo` always appends
  `&scope=${projectScope}` to the auth URL with a truthy value — I checked all 36 input shapes —
  that branch never runs for a real build. This PR makes it authoritative on every request, and
  #80 turns any disagreement with the caller's `projectKey` into a 403.

  I understand and agree with why the caller's `scope` had to go: round-tripping it through the
  auth URL meant a build could assert its own scope. But swapping "trust the caller" for "trust
  the annotation" replaces a value that was consistent-by-construction with `SD_SONAR_PROJECT_KEY`
  (both came from one `getProjectData` call in `_getInfo`) with a second, independent derivation.
  `/coverage/info` still resolves scope its own way — a caller `projectKey` prefix wins outright,
  otherwise `_resolveCoverageScope` applies the enterprise default. When those two disagree, the
  build now gets a 403 where master gave it a token.

  I couldn't determine from either repo what `screwdriver-cd/launcher` passes to
  `/coverage/info` — that's the fact that decides whether this is reachable. If you know, it'd be
  worth writing down next to the annotation lookup here, since this route is now the only thing
  standing between that annotation and a rejected build.

  If we wanted the invariant to hold by construction rather than by assumption, the shape would be
  `/coverage/info` and `/coverage/token` sharing one scope derivation. I'm not proposing that
  here — `/coverage/info` is also called by the UI with user credentials, so it's a bigger change
  than this PR should carry — but it seems like the right eventual destination.

minghay:
On the scope derivation — I posted the full evidence on coverage-sonar#80 rather than duplicate it
here, since you didn't duplicate it either. Short version: I traced `screwdriver-cd/launcher`
directly. It never omits `scope` (the `CoverageURL` template has `scope=%s` fixed, `default:""`
when the annotation is absent, never dropped) and never sends a `projectKey` to `/coverage/info`
(the template has no such param, and `plugins/coverage/info.js` is the only caller of `getInfo` in
the API). Both routes read `job.Permutations[0].Annotations.CoverageScope` — same field, same
index — so the divergence rows in your enumeration all require an input shape this caller can't
produce. Matrix jobs don't add risk either: the launcher reads the same index 0, and matrix
permutations don't vary annotations in the first place.

The one real gap: a pipeline sync rewriting the annotation between build start and coverage
teardown does diverge, and `commands.txt`'s `|| true` swallows it silently. Narrow and
self-healing, but real — worth having in view.

I added a comment at `plugins/coverage/token.js:46-52` (pushed) recording this invariant with the
launcher file:line citations, per your "cheapest option" suggestion. Not committing to unifying
derivation across `/coverage/info` and `/coverage/token` right now — agreed it's the right eventual
shape, but `/coverage/info` also serves the UI under user creds, so it's a bigger change than this
fix should carry, and there are two follow-ups queued already.

On the dependency bump: opened as an actual PR rather than a tracking issue or a `package.json`
comment (JSON can't hold one anyway) — #3520, one line,
`"screwdriver-coverage-sonar": "^5.0.0"` → `"^6.0.0"`, no lockfile changes. It's opened early and
expected to fail CI right now since `6.0.0` isn't published yet — do not merge it until this PR is
deployed and coverage-sonar#80 has actually cut `6.0.0`.

DekusDenial:
Two new commits since round 2: the invariant comment (`cdfe494`) and a master merge
  (`6da0861`). Both good. **No changes requested on this PR.**

  ## The invariant comment — verified

  I checked each citation against `screwdriver-cd/launcher` rather than taking it on trust, and
  all of them hold:

  | Comment claims | Verified |
  |---|---|
  | `launch.go`, `GetCoverageInfo` reads `job.Permutations[0].Annotations.CoverageScope` | `launch.go:404-408` — `coverageScope := ""`, then `if
  len(job.Permutations) > 0 { ... }`, passed straight in ✔ |
  | always sends it as `scope`, empty string when absent, never omitted | `screwdriver.go:150` is tagged `default:""`; `scope=%s` is a fixed
  positional verb in the `CoverageURL` template ✔ |
  | `CoverageURL` has no `projectKey` param at all | `screwdriver.go:23` — confirmed ✔ |

  That resolves my round-2 question, and this is the right place for it: the comment sits on the
  exact line the 403 depends on and cites the source, so the next reader can re-verify in a minute
  instead of an afternoon. The "short of a pipeline sync rewriting this annotation between build
  start and this request" caveat is accurate, and naming it beats claiming the invariant is total.

  One detail that made it more convincing than it needed to be: I'd have expected `default:""` on
  one side and a missing annotation (`undefined`) on the other to diverge, but
  `_resolveCoverageScope` maps `''`, `undefined` and `'undefined'` to the same default, so the
  empty case agrees too.

  The master merge doesn't touch `plugins/coverage/` — `git diff cdfe494 6da0861 -- plugins/coverage/`
  is empty. 22 tests passing on the new head.

  ## For visibility only — the fix belongs in #80

  I've left a finding on coverage-sonar#80: annotation values that aren't exactly `pipeline`,
  `job`, or empty (a miscased `Pipeline`, say) now get a 403, because `getProjectData` treats
  "anything not `pipeline`" as job scope while #80's new check does exact string equality against
  the key prefix. It's reachable *through* the launcher precisely because this route forwards the
  annotation verbatim — which is correct behavior on your side.

  Explicitly **not** asking for anything here. Forwarding the raw annotation is right, and
  normalising it in this route would just push coverage-sonar's internal inconsistency across a
  repo boundary. This PR is fine as-is.

  ## One note, not a request

  `package.json` still pins `"screwdriver-coverage-sonar": "^5.0.0"` while the README now says
  `≥6.0.0`, so the security fix doesn't actually take effect until someone widens that range.
  That's correct for this PR — the whole point is that this deploys first — and I see it's already
  on your follow-up list. Flagging only so it doesn't get lost between two READMEs; I don't think
  it should hold up this merge, and I don't need a tracking issue to approve.

  LGTM.

DekusDenial (approved):

